Issue #1492, opened 2015-01-13
Python Error when using keyword editor


adissadis:
# Problem
Got python error code when tried to define new shapefile as hazard and exposure

# Detail
I tried to using keyword editor on new shapefile to define the shapefile as hazard layer. But when i click InaSAFE keyword editor, Python error window pop up and showed that i got an error while executing python code. 

When Harry, tried to using keyword editor to the hazard shapefile and exposure shapefile (Jakarta flood and building osm) he got an error also.

I think it's because our shapefile missing .keywords file so we can't run InaSAFE keyword editor to define the shapefiles as hazard or exposure.

# Screenshot
![inasafe_1](https://cloud.githubusercontent.com/assets/2520159/5717408/fb031452-9b32-11e4-9a80-a33b7be9e922.png)

# Platform
First device:
QGIS 2.4 and also 2.6 with InaSAFE 3 (2.2.dev-master) and with operating system Windows 7 64bit

Second device:
QGIS 2.6 with InaSAFE 3 (2.2.dev-master) and with operating system Windows 8.1 64

ismailsunni:
Well, this is a regression. It works in previous InaSAFE. I tried with version 2.2

ismailsunni:
After some debugging, this is quite strange (at least for me):

The reason is, in `plugin.py`, all modules are imported with `inasafe-dev` prefix. See image below for a better explnation:
![evaluate expression_027](https://cloud.githubusercontent.com/assets/1421861/5751876/59e15cf2-9c9d-11e4-9369-40a687bc21cb.png)

Both of them are `NoKeywordsFoundError`. But the first one is referenced by <class `inasafe-dev.safe.common.exceptions.NoKeywordsFoundError> and the second on is referenced by <class `safe.common.exceptions.NoKeywordsFoundError>. So, they are "different". And this makes the exception is not caught any more.

akbargumbira:
The same kind of issue http://stackoverflow.com/questions/1988475/exception-not-getting-caught-when-not-in-the-right-package-in-python

ismailsunni:
@akbargumbira I fixed it like this : https://github.com/AIFDR/inasafe/pull/1515/files#diff-66c71ea3faaa52c277917c951aa67921R41

What do you think?

akbargumbira:
ouch that is ugly :p

As safe context is now should contain all the codes related to plugin, I think it's better if we move the plugin.py to safe package. It was in safe_qgis package.

[...]

PR #1515, opened 2015-01-16
Fix 1492

ismailsunni:
Fix for #1492


ismailsunni:
Don't merge yet. It still has failed unit test.

akbargumbira:
is other changes necessary to fix this? seems like we just need to move plugin.py to solve

ismailsunni:
@akbargumbira 
I added unit test for keywordless also. But, the load_layer method can't load it. So, I update it.

I updated the Makefile, (only removing unused part).

And, I add a handler for CSV layer. See the parseurl part.

Issue #1489, opened 2015-01-12
[IFCW] Add Filter for Impact Function


ismailsunni:
Problem
-----------
See #1482 
Since we want to remove impact function browser, we need to add filter functionality in impact function centric wizard.

Proposed Solution
--------------------------
Add some filtering for the impact function. :
- by exposure type (no need by hazard type, since we have group them by hazard type)

I think that's enough. Because there is no so many impact function for now. Perhaps, we can add some filter function later. E.g. by exposure layer type, hazard layer type, author, etc.

Perhaps we can put it above the list of impact function:
![inasafe_026](https://cloud.githubusercontent.com/assets/1421861/5704199/c44fe764-9aa1-11e4-9b3b-d3603622d033.png)

Expected Result
---------------------
Impact Function Wizard has filter functionality for impact function.

cc @Charlotte-Morgan @timlinux @akbargumbira perhaps you have better ideas about this ticket.

Charlotte-Morgan:
hi @ismailsunni 
Thank you for raising this issue and for the proposed solution.
I am concerned that although the filtering will help the users, the list of IF is still a little (well actually very) confusing for users. 

*I would like to propose using the same terminology in the 'filtering' that users are already familiar with in the software; and to use a 'matrix' approach to the filtering.

the matrix would filter on hazard and exposure and allow the user to specify:

"I want to look at the impact of <hazard> on <exposure>."

The interface might look something like this and be presented as the first screen in the IFCW:
![image](https://cloud.githubusercontent.com/assets/5743135/5755727/8682b534-9cda-11e4-9325-10a3b0421d33.png)

invalid options would be grayed out.

the users would then click on a cell (the intersection of the hazard and exposure they are interested in) and from there be offered to define the data they have available or specify the impact function they wish to use.

eg clicking on the flood x people cell could open a new matrix in which "Pretty IF" names could be presented:
![image](https://cloud.githubusercontent.com/assets/5743135/5755805/54b3d834-9cdb-11e4-9d2a-5dd8aa73516d.png)

another alternative would be to nest the 2 matrices into 1 ...

* we also need some super clear metadata to support each filter element and IF name so users are very clear on what they are selecting and what it will do.

@fredychandra @mahardika @iyan31 @adissadis @akbargumbira @ismailsunni @timlinux ; please consider this pseudo RFC and comment. I am happy to host a meeting at AIFDR to discuss this and any other ideas AIFDR and HOT may have to ensure this is really user friendly.

PR #1520, opened 2015-01-19
Impact function matrices for the wizard

borysiasty:
Solves #1489

- The labels over matrices are more or less reasonable
- Cells are stretched to available space
- Doubleclicks on available cells move to the next step
- The test is updated
- The 'generic' hazard is hidden
- The vector-line datatype for hazards is hidden

@Charlotte-Morgan suggested to change hazard name from "tephra" to "volcanic ash". I believe this change should be done in metadata.


borysiasty:
Superseded by #1525

PR #1525, opened 2015-01-25
Impact function matrices for the wizard

### LINKED ISSUE #1472: [IFCW] - only displays results when data layers added from disk
# problem
When the IFCW adds layers from disk these are not added to the display.

# solution
is this what the user would expect?

### LINKED ISSUE #1489: [IFCW] Add Filter for Impact Function

borysiasty:
Supersedes #1520. Maybe this time Travis will pass. 

Closes #1489: 
 - The labels over matrices are more or less reasonable
 - Cells are stretched to available space
 - Doubleclicks on available cells move to the next step
 - The test is updated
 - The 'generic' hazard is hidden
 - The vector-line datatype for hazards is hidden
@Charlotte-Morgan suggested to change hazard name from "tephra" to "volcanic ash". I believe this change should be done in metadata.

Closes #1472:
 - Layers selected in the browser are now added to map canvas (if they didn't)
 - Fixed layer names in the summary tab

### DIFF
--- a/safe/gui/tools/test/test_wizard_dialog.py
+++ b/safe/gui/tools/test/test_wizard_dialog.py
@@ -202,11 +202,12 @@ class WizardDialogTest(unittest.TestCase):
         message = 'Invalid categories! It should be "%s" while it was %s' % (
             expected_categories, categories)
         self.assertEqual(set(categories), set(expected_categories), message)
-        # Check if the Next button state is on the right state
+        # The Next button should be on disabled state first unless the keywords
+        # are already assigned
         message = ('Invalid Next button state in step 1! Enabled while '
                    'there\'s nothing selected yet')
-        self.assertTrue(
-            not dialog.pbnNext.isEnabled(), message)
+        self.assertTrue(not dialog.pbnNext.isEnabled()
+                        or len(dialog.lstCategories.selectedItems()), message)
         # Select hazard one
         dialog.lstCategories.setCurrentRow(hazard_index)
         message = ('Invalid Next button state in step 1! Still disabled after '
@@ -237,10 +238,12 @@ class WizardDialogTest(unittest.TestCase):
                    '%s') % (expected_subcategories, subcategories)
         self.assertEqual(
             set(subcategories), set(expected_subcategories), message)
-        # The Next button should be on disabled state first
-        self.assertTrue(
-            not dialog.pbnNext.isEnabled(), 'Invalid Next button'
-            ' state in step 2! Enabled while there\'s nothing selected yet')
+        # The Next button should be on disabled state first unless the keywords
+        # are already assigned
+        self.assertTrue(not dialog.pbnNext.isEnabled()
+                        or len(dialog.lstSubcategories.selectedItems()),
+                        'Invalid Next button state in step 2! '
+                        'Enabled while there\'s nothing selected yet')
         # Set to tsunami subcategories
         dialog.lstSubcategories.setCurrentRow(tsunami_index)
         message = ('Invalid Next button state in step 2! Still disabled after '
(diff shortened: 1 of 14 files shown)


ismailsunni:
Hi @borysiasty 

Thanks for the PR. The table looks great.

But, I found a small problem. When we open IFCW, we will got something like this:
ifcw starting step

Then I chose one cell, then go to the next step. If I went back to the first step (by clicking Back Button), the cell that I selected is grey. Something like this:
ifcw starting step after go back

We can see that the flood vs population is grey.

I think it should be orange (like when we select cell), it will be more natural.

borysiasty:
Strange. As a KDE user, I don't remember the style details in GTK, please help me to understand it correctly: 

Your normal selection style is orange gradient in the cell + bold headers in addition. When you click the cell, it goes orange properly. And when you come back to this step (the next screenshot), the cell is actually still properly selected (I see the gradient and the bold headers), the only problem is with the missing color? 

Or is the cell disabled like other gray cells? What happen if you select that cell again?

Does this problem occurs every time and with any cell?

ismailsunni:
Yes, the only problem is the cell becomes grey.

If I select the cell, it becomes orange.
If I select any green cell (or any location in the table, except the grey cells), that cell becomes green.
If I click grey cells, that cells stays grey.

And yes, it occurs every time and with any cell.

timlinux:
A few more cosmetic requests:

* could you put a nice big checkmark in the selected cell?
* Can you capitalise the column and row headings

@all do we really want tephra in the list of hazards? They are only supported by categorical / categorised IFs...

timlinux:
@ismailsunni also will you go ahead and strip out the IF browser (old one)?

ismailsunni:
@timlinux 
- tephra, -1 from me. We'd better remove it.
- Strip IF Browser : ok

ismailsunni:
@timlinux 

PR for removing tephra : #1528

Charlotte-Morgan:
WOW @borysiasty  I'm super impressed!! This looks absolutely fantastic. Thank you for adopting the matrix.

Some cosmetic suggestions (in order of appearance) & some other ideas:
- [ ] in the matrices, make the writing bigger and the boxes smaller. +1 for capitalisation + 1 for icons and text on different lines.
- [ ] consistent font styles. eg can the "Available functions ... " text at the bottom of the window be the same font as the "This wizard ... " at the top of the window?
- [ ] on screen 2 of the wizard, is it possible to return some feedback from the users click in screen 1; ie if they click on the intersection of flood and population; can the next screen refer to Flood Hazard & Population Exposure and again later ask them to select their flood hazard data.  I think this would be great, i expect it's more than a 5 minute job, so would be nice but not a show stopper.
- [ ] improve metadata descriptions for IF #1530
- [ ] the wizard refers to the 'hazard data registration wizard' when users choose to select data from disk. Is this name (for the other wizard) what it is called elsewhere & in doco and training materials?
 @ismailsunni @timlinux @borysiasty - great stuff.

borysiasty:
> in the matrices, make the writing bigger and the boxes smaller.

@Charlotte-Morgan do you mean to not extend the grid to all available space? Previously Tim told me to do so ;)

Issue #1441, opened 2014-12-10
InaSAFE minimum needs result problem


adissadis:
# Problem 
InaSAFE minimum needs produce different result from total minimum needs compared to minimum needs based on aggregation. 

# Detail
We got email from Elida that she ask about the different result from minimum needs from total minimum needs compared with detailed minimum needs. 

**Total result minimum needs:** 
![image](https://cloud.githubusercontent.com/assets/2520159/5372702/c6c6a2da-807f-11e4-8890-ae9e49c4e132.png)

**Detailed minimum needs:**
![image](https://cloud.githubusercontent.com/assets/2520159/5372737/74c35b26-8080-11e4-873c-a917c88dc0d1.png)

From the 2 images above, there are huge different number for detailed minimum need, for example **Rice** in total minimum needs only need 140kg but when i looked at detailed minimum needs, it become 15.935kg and it also affect for other needs such as family kits, toilet, drinking water, and clean water.

# Platform
We're using 2 devices to make sure the problem are not the computer problem.

First device run using:
QGIS 2.4
InaSAFE 2.1
Windows 7

Second device run using:
QGIS 2.6
InaSAFE 2.2
Windows 7

akbargumbira:
From Elida's email,  this difference is because the first table calculates minimum needs based on evacuated people, but in the postprocessor, it calculates based on affected people.

ismailsunni:
This is quite easy to fix, actually.

So, it's true what Elida's said.

Therefore, the solution, we should choose, which one we want to use to calculate minimum needs. Is it `evacuated people` or `affected people`. This question is for @Charlotte-Morgan (but perhaps @timlinux or @akbargumbira can give some suggestions)

But, still, there will be slight difference between the total and the detailed one. Since we round up the total affected or evacuated people. And this is normal for me.

I don't really want to round up the detailed one (we will add rounded numbers, and it will give more difference with the total one).

ismailsunni:
Now, we use evacuated for minimum needs based on aggregation. See below:
![people_affected_by_flood_prone_areas_table-page-001](https://cloud.githubusercontent.com/assets/1421861/5914932/c693a28c-a630-11e4-9fb5-6496c9862471.jpg)
![people_affected_by_flood_prone_areas_table-page-002](https://cloud.githubusercontent.com/assets/1421861/5914959/1bc7f42e-a631-11e4-88f3-1c04f4674c88.jpg)

I will make sure the related unit tests are passing and commit the fix.

[...]

PR #1526, opened 2015-01-27
Fix 1441 minimum needs result

ismailsunni:
For #1441


Charlotte-Morgan:
hi @ismailsunni - this is looking good.
- [ ] I have looked at flood raster on population raster with aggegation (jakarta) and its fine. 
so in this case affected = evacuated = basis for minimum needs. I'd like to see rounding in the general & detailed reports. But that can wait for IF review. we probably need something that states that all affected people are assumed to be evacuated etc
- [ ] i have looked at flood polygon on population raster with aggregation (jakarta). In this case evacuated = 1% of affected and so the numbers in the reports are different. I like the extra information you include in the detailed minimum needs report (for people needing evacuation). Things get messy though with the detailed gender report and the detailed age report. Both are for affected and not the evacuated population; but then hygiene packs and additional rice are calculated on the affected rather than evacuated population. It would be helpful to add (affected people) to the title for the detailed gender and detailed age reports. Please add this to things to be included in the IF review
# cc
@adissadis

ismailsunni:
For point 1, basically we want to do it in IF review.
I will put in the notes that all affected people are assumed to be evacuated for now. So, for flood raster on population raster, it will give report like:
selection_037


For point 2:
For flood polygon and population raster.
I have added extra notes in the table header. See below:
selection_038

Is that ok?

Actually, I added the same table header for population-IF, for the detailed table. To make them uniform.

safe/impact_statistics/aggregator.py
```diff
@@ -949,27 +946,27 @@ def _aggregate_line_impact(self, safe_impact_layer):
             tmp_filename = unique_filename(
                 suffix='.shp', dir=output_directory)
             res = self.processing.runalg('qgis:exportaddgeometrycolumns',
-                projected_layer,
-                # 2, # Ellipsoidal
-                0,  # Layer CRS
-                tmp_filename)
+                                         projected_layer,
```
timlinux: it was correctly indented above (pull left!)

timlinux:
Great that is nice and clear thanks! LGTM

PR #1528, opened 2015-01-28
Remove tephra from supported hazard.

ismailsunni:
Remove tephra since it's not supported yet.


cchristelis:
Why do you just comment it out and not remove it? Is it going to be added back soon?

Charlotte-Morgan:
please leave this hazard & rename it volcanic ash.
Why - because if anyone turns up with hazard data with kg/m2 as a unit and wants to run a categorical IF. 
Yes - these data do exist ... and someone might want to do it

ismailsunni:
Closing this. Applied in #1549  Change tephra to volcanic ash.

PR #1546, opened 2015-01-30
Update translation.

ismailsunni:
All is translated.


timlinux:
Thanks @ismailsunni. As a general rule just commit translation PR's once travis passes (after we fix travis of course) since it is probably difficult for others to evaluate it.

PR #1559, opened 2015-02-01
Updates to the text for wizards. Added descriptions to the different lay...

timlinux:
...er types in metadata. Try to have a single point of truth for metadata instead of duplicating it in different places.


Charlotte-Morgan:
Some more text for wizards: This goes with screen 2 of the IFCW.

A continuous raster is one where cell values are real data values such as: depth of flood water in meters or the number of people per cell.

A classified raster is one where cell values represent classes or zones such as: high hazard zone, medium hazard zone, low hazard zones.

timlinux:
Thanks @Charlotte-Morgan I will update it

PR #1569, opened 2015-02-02
Pylint

cchristelis:


safe/common/dynamic_translations.py
```diff
@@ -48,7 +48,7 @@
 # for s in strings:
 #     names[s] = '_(%s)' % s
 
-import PyQt4.QtCore.QObject.tr as tr
+import PyQt4.QtCore.QObject.tr as tr  # pylint: disable=F0401
```
timlinux: OK but we should still rather be importing from our utils i18n module here...

safe/common/resource_parameter_widget.py
```diff
@@ -21,12 +21,23 @@ def __init__(self, parameter, parent=None):
...
+        # pylint: disable=E1002
         super(ResourceParameterWidget, self).__init__(parameter, parent)
+        # pylint: enable=E1002
         self.set_unit()
 
+    def get_parameter(self):
+        """Obtain the parameter object from the current widget state.
+
+        :returns: A BooleanParameter from the current state of widget
+
```
timlinux: :rtype: BooleanParameter

If you use rtype, pycharm will add extra syntax checking and syntax awareness to code completion for you.

timlinux:
LGTM thanks!

Ill merge it now

PR #1618, opened 2015-02-08
[Trial] Remove exception if there is duplicate IF.

ismailsunni:


ismailsunni:
Fyi, I think we *should* raise an exception if there is duplicate impact function name. Duplicate means, different impact function with the same name

Issue #1159, opened 2014-07-28
UTF8 characters in a layer path make InaSAFE crash


vdeparday:
If the full path has a folder name with accents ( /données/expositions/ ). InaSAFE crashes when trying to read the data with the following error:

Traceback (most recent call last):
  File "/Users/viv/.qgis2/python/plugins/inasafe/safe_qgis/plugin.py", line 526, in show_keywords_editor
    keyword_io.read_keywords(self.iface.activeLayer())
  File "/Users/viv/.qgis2/python/plugins/inasafe/safe_qgis/utilities/keyword_io.py", line 96, in read_keywords
    source = str(layer.source())
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 73: ordinal not in range(128)

Going in debug mode with pdb, this is what I have found at line 96: 

(Pdb) layer.source()
u'/Users/viv/Dropbox/Worldbank/LAC/Haiti/UR Haiti/USB Key/QGIS InaSAFE/donn\xe9es/al\xe9as/inondations/hti_hazardhydrology_frequentflood_westdepartment_nathat_polygon/hti_hazardhydrology_frequentflood_westdepartment_nathat_polygonPolygon.shp'

After removing all the accent from the path manually, it worked again.

Gustry:
@vdeparday I've made a pull request. Could you check if this solve your problem ?

akbargumbira:
This is fixed here https://github.com/AIFDR/inasafe/pull/1674

akbargumbira:
I merged the PR, closing this

mbernasocchi:
@ismailsunni can you check if this is still happening when a Layername (in layer properties) includes utf8 I get:

```Ein Fehler trat bei der Ausführung von Python-Code auf: 

ProgrammingError: You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings (like text_factory = str). It is highly recommended that you instead just switch your application to Unicode strings. 
Traceback (most recent call last):
  File "/home/marco/.qgis2/python/plugins/inasafe-dev/safe/gui/widgets/dock.py", line 546, in update_layer_name
    self.keyword_io.update_keywords(layer, {'title': name})
  File "/home/marco/.qgis2/python/plugins/inasafe-dev/safe/utilities/keyword_io.py", line 184, in update_keywords
    self.write_keywords(layer, existing_keywords)
[...]

PR #1631, opened 2015-02-10
add encoding to UTF8 about layers #1159

Gustry:
see #1159  - add encoding to UTF8 about layers


timlinux:
LGTM - please merge when ready - thanks!

ismailsunni:
hi @timlinux @Gustry 

I'm not sure about this. The last time, I tried similar solution, but it ends up with an error if we try to play with UTF8 character else where.

Perhaps, we can make a unit test related to UTF8 character first before merging, i.e. create a unit test with UTF8 character for the file name (or even the layer name, keyword, and so on)

cc @akbargumbira

akbargumbira:
Hi, yes, from looking at the codes (have not tried this PR yet), pretty sure that this is only a partial solution. This will not fix e.g the case when the data itself has a non-ASCII character..

timlinux:
Hi

Yes it is acknowledged that it is only a partial solution - if fixes an unhandled exception on @Gustry 's computer when using the new developer mode 'load standard layers' button. 

Regards

Tim

akbargumbira:
Can one of the admins verify this patch?

akbargumbira:
Can one of the admins verify this patch?

Issue #1625, opened 2015-02-10
We need to rename metadata.py to definitions.py


timlinux:
We want to use metadata.py for a true metadata structure and currently metadata.py is just a list of terms and definitions.

# CC

@lucernae
@ismailsunni
@mbernasocchi

Issue #1620, opened 2015-02-08
Ambiguous use of variable name 'key'


timlinux:
I found this in ``safe/impact_functions/core.py:835``

```
    for key, func in plugins_dict.iteritems():
        if not is_function_enabled(func):
            continue
        dict_retval['title'].add(get_function_title(func))
        dict_retval['id'].add(key)
        for requirement in requirements_collect(func):
            dict_req = parse_single_requirement(str(requirement))
            for key in dict_req.iterkeys():
                if key not in atts:
                    break
                if type(dict_req[key]) == type(str()):
                    dict_retval[key].add(dict_req[key])
                elif type(dict_req[key]) == type(list()):
                    dict_retval[key] |= set(dict_req[key])

```

The key in the inner loop overshadows that of the outer loop. It is not clear in the index references whichshould refer to the inner versus the outer. Could you please update this to use clear unambiguous variable names? Thanks!

Tim

timlinux:
Thanks @ismailsunni

PR #1633, opened 2015-02-10
Fix 1625 1620

ismailsunni:
For #1620 and #1625


ismailsunni:
hi @lucernae would you like to review my PR :)

safe/impact_functions/core.py
```diff
@@ -674,8 +658,9 @@ def get_admissible_plugins(keywords=None):  # , name=None):
 
 
 def parse_single_requirement(requirement):
-    '''Parse single requirement from impact function's doc to category,
-        subcategory, layertype, datatype, unit, and disabled.'''
+    """Parse single requirement from impact function's doc to category,
```
akbargumbira: It's better to have a one-line explanation of the function and explain more after that as a different paragraph

akbargumbira:
LGTM

safe/impact_functions/core.py
ismailsunni: Yes, of course.

I just change the single quote to double quote. We will remove this function soon we remove the docstring. It's 80% finished. so, I will just leave it now for now

Issue #1570, opened 2015-02-02
Postgis layer keywords are not always found due to changed details in the uri 


borysiasty:
Problem
-----------
A postgis layer added in various ways may have different uri. Especially it may contain or not the srid and type parameters, the srid may be changed, the sql filter may be applied. It all affects the uri hash and makes the layer keywordless.

Proposed Solution
------------------------
Strip down the uri for creating the hash and only use the host + schema + table.

PR #1641, opened 2015-02-12
Issue #1570: fix layer uri normalization for postgres and delimitedtext

lucernae:
Regarding Issue #1570:

Can someone review this? Perhaps @borysiasty or @timlinux  could?
The problem actually much bigger than that. The delimitedtext also has the same problem.
The layer source has different format for each providerType, so other layer types probably has the same problem. Right now, the PR fixes postgis and delimitedtext issues.


lucernae:
Oops, will be careful next time :p :+1:

PR #1643, opened 2015-02-13
fix #1423 save .xml and .keywords when using 'saved as'

### LINKED ISSUE #1423: We need metadata and other resources handled on 'save as'
When saving a file using 'save as', the layer metadata and any other related files should also be copied over.

See also #1244

Gustry:
fix #1423 
When we use the "saved as" function on a layer, .xml and .keywords are also saved

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -567,15 +567,15 @@ class Dock(QtGui.QDockWidget, FORM_CLASS):
 
     @pyqtSlot(QgsMapLayer, str)
     def save_auxiliary_files(self, layer, destination):
-        """Save auxiliary files when using the saved as function.
+        """Save auxiliary files when using the 'save as' function.
 
         If some auxiliary files (.xml or .keywords) exist, this function will copy them
-        when the saved as function is used on the layer.
+        when the 'save as' function is used on the layer.
 
         :param layer: The layer which has been saved as.
         :type layer: QgsMapLayer
 
-        :param destination: The new filename of the layer
+        :param destination: The new filename of the layer.
         :type str
 
         """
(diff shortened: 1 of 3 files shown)


safe/gui/widgets/dock.py
```diff
@@ -560,6 +565,43 @@ def validate(self):
             message = self.ready_message()
             return True, message
 
+    @pyqtSlot(QgsMapLayer, str)
+    def save_auxiliary_files(self, layer, destination):
+        """Save auxiliary files when using the saved as function.
```
timlinux: 'save as'

safe/gui/widgets/dock.py
```diff
@@ -560,6 +565,43 @@ def validate(self):
...
+
+        If some auxiliary files (.xml or .keywords) exist, this function will copy them
+        when the saved as function is used on the layer.
```
timlinux: 'save as'

safe/gui/widgets/dock.py
```diff
@@ -560,6 +565,43 @@ def validate(self):
...
+
+        :param layer: The layer which has been saved as.
+        :type layer: QgsMapLayer
+
+        :param destination: The new filename of the layer
```
timlinux: Full stop at end of sentence please.

safe/gui/widgets/dock.py
```diff
@@ -560,6 +565,43 @@ def validate(self):
...
+        source_basename = os.path.splitext(layer.source())[0]
+        source_keywords = "%s.keywords" % source_basename
+        source_xml = "%s.xml" % source_basename
+
+        destination_basename = os.path.splitext(destination)[0]
+        destination_keywords = "%s.keywords" % destination_basename
+        destination_xml = "%s.xml" % destination_basename
+
+        #Keywords
+        if os.path.isfile(destination_keywords):
```
timlinux: probably you need to add an exception handler for file permissions errors here and similar clauses below.

timlinux:
LGTM if my small comments are taken into account.

Thanks

PR #1647, opened 2015-02-16
Update parameters and remove parameter test.

ismailsunni:
- Remove test in safe_extras/parameters
- Uniform the way we import in  safe_extras/parameters


safe_extras/parameters/string_parameter.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, guid=None):
         :param guid: Optional unique identifier for this parameter. If none
             is specified one will be generated using python hash. This guid
             will be used when storing parameters in the registry.
-        :type guid: str
+        :type guid: str, None
         """
         super(StringParameter, self).__init__(guid)
-        self.expected_type = [str, unicode]
```
cchristelis: Why do you should we not allow unicode here? I mean the if unicode is passed into it, then the type test would break. That is a condition that I had and it made sense to me to allow for unicode.

safe_extras/parameters/qt_widgets/test/test_boolean_parameter_widget.py
```diff
@@ -1,53 +0,0 @@
-# coding=utf-8
```
cchristelis: Hi Ismail, I haven't had a detailed look at the surrounding code, but what is the rationale for removing these tests and what is the effect on test coverage?

ismailsunni: The rationale is we take safe_extras as outside code, we don't need to test them as they are not our code. And it doesn't effect to our test coverage since we only cover safe packager.

But, for parameters, it's actually not inasafe's code, it's from different repo (https://github.com/AIFDR/parameters). Although the developers are in the same team.

safe_extras/parameters/string_parameter.py
ismailsunni: Hi @cchristelis thanks for pointing this out. I think there was someone adding unicode directly to safe_extras/parameters. That's why it will have some different with the original repo.

I will update the original repo (https://github.com/AIFDR/parameters), and fix this.

Thanks

PR #1652, opened 2015-02-17
Fix 1650 : add new methods to display informations through QgsMessageBar and QMessageBox

Gustry:
Fix 1650 : add new methods to display informations through QgsMessageBar and QMessageBox


timlinux:
LGTM

PR #1655, opened 2015-02-17
Added api docs scripts from docs repo

timlinux:
Please do not merge yet


ismailsunni:
Need to update the create_impact_function_docs.py to match with current inasafe code first before merging.

Issue #1648, opened 2015-02-16
We need to allow the user to save impact layer to their own designated directory


timlinux:
# Problem

Impact layers are save to the user's temporary directory (that varies according to Operating System). Under unix-like OS's these products get destroyed on system reboot. In all cases the file name and path is non-intuitive to the user.

# Proposed solution

For this issue I am proposing a new option in the InaSAFE options dialog that allows the user to designate their preferred output directory for impact layers so that they may be organised and persistent. Note I am *not* proposing that all derivative products (e.g. clip layers) are saved there, only the final impact layer.

The expected behaviour will be:

* when an impact layer is saved it is named according to this convention: ``EXP + On + Haz + DDMMMMYYYY + HHhMM.SS.EXT`` so a preactical example may be: ``FloodOnBuildings_12March2015_10h22.04.shp``
* if the user has designated an impact layer output directory, the layer and its attendant metadata files should be saved there, otherwise the existing behaviour should continue (though using the new impact layer naming scheme).

# Expected outcomes

* Users should be able to collate and browse the output layers produced by InaSAFE in an intuitive manner
* Users should be able to persist Impact Function outputs across system reboots

# CC

@Charlotte-Morgan 

# See also

I know there was a similar ticket filed but I can't find it at the moment, please add any references to related tickets to this section.

Gustry:
You will get a file like this : 
``Haz + On + Exp + DDMMMMYYYY + HHhMM.SS-UNIQUE.EXT``
instead of above : 
``EXP + On + Haz + DDMMMMYYYY + HHhMM.SS.EXT``

Do we need to replace accentuated characters ?

Gustry:
When the @akbargumbira 's work about unicode will be merged, I will fix if the user directory contains an accent.

PR #1663, opened 2015-02-23
Allow the user to save results to their own directory [OLD branch]

Gustry:
Allow the user to save results to their own directory.

Please, don't care about unicode stuff, it won't be merged directly. I've put the code here for review about #1648


Gustry:
We should maybe avoid accent in filename, isn't it ?
February in french contains an accent.
When the work about unicode will start, I will add a function to replace accented characters.

safe/engine/test/test_engine.py
```diff
@@ -414,6 +418,45 @@ def test_polygon_hazard_with_holes_and_raster_exposure(self):
...
+        # These imports must be inside the test.
+        from PyQt4.QtCore import QCoreApplication, QSettings
+        from qgis.core import QgsApplication
+
+        # noinspection PyCallByClass,PyArgumentList
+        QCoreApplication.setOrganizationName('QGIS')
+        # noinspection PyCallByClass,PyArgumentList
+        QCoreApplication.setOrganizationDomain('qgis.org')
+        # noinspection PyCallByClass
+        QCoreApplication.setApplicationName('QGIS2InaSAFETesting')
```
akbargumbira: Hi @Gustry, we already have this in get_qgis_app, so you should not specify this for QSetting anymore and just call get_qgis_app in the beginning of the module. This is for an example https://github.com/AIFDR/inasafe/blob/develop/safe/gui/tools/batch/test/test_batch_dialog.py#L27

safe/messaging/item/message_element.py
```diff
@@ -69,6 +70,7 @@ def _is_qstring(message):
     def _is_stringable(message):
         return (isinstance(message, basestring)
                 or isinstance(message, int)
+                or isinstance(message, str)
```
akbargumbira: str is implemented from basestring class. No need to have this as it will be catched by isinstance(message, basestring) above

akbargumbira:
@Gustry in this PR, you have a PR related to unicode issue. Can you differentiate this PR with that?

Gustry:
@akbargumbira Yes, I'm sorry I have some code about unicode issue inside.
However, I won't merge this PR directly. I will "cherry-pick" commits about this PR to another new branch based on "develop". I need these lines of code about unicode to get QGIS working on my laptop. So don't care about the unicode in this PR, it won't be merged. I have make this PR just to review the code about #1663 .

safe/messaging/item/message_element.py
Gustry: This line won't be merged. It's a hack for my qgis.
Thanks for the tip !

safe/engine/test/test_engine.py
timlinux: @akbargumbira when you run it as a single test you need this because in this test module, APP is not created.

timlinux: @Gustry YEs best practice would be for files / paths created by InaSAFE to not include accents (unless we are propagating a name defined by the user).

Gustry:
The good pull request will be https://github.com/AIFDR/inasafe/pull/1676
Sorry for that. Next time, I will cherry-pick before pushing to right branch.

Gustry:
https://github.com/AIFDR/inasafe/pull/1676 has been merged. I'm closing this one.

PR #1710, opened 2015-03-04
Fix PEP8 warning on master

lucernae:
Fix most of PEP8 warning and rewrote lambda function as def


lucernae:
Sorry, wrong target branch

PR #1718, opened 2015-03-04
Fix #1207 : add 'building points' to the osm downloader

### LINKED ISSUE #1207: Allow an option in the OSM downloader to output building layer as points.
# Problem

Currently, OSM downloader extracts data for building only as polygons (closed way in OSM definition).  But, there are cases where the buildings in OSM are points (nodes).  The current IF can already use buildings as points as an exposure layer.  It would be good to add an option in the OSM downloader to choose the geometry type for building points or polygons.

# Solution
In the OSM downloader, add option to extract the building as points/polygons.  When point was selected, add another step to convert the building polygons as centroid and combine this layer to the output point layer.

# Expected Outcome
OSM downloader can download building data as points.

Gustry:
Add 'building points' to the osm downloader.
It will download buildings but as a point layer.
The "All" option will fetch : 
  - buildings
  - building points
  - roads

### DIFF
--- a/safe/gui/tools/osm_downloader_dialog.py
+++ b/safe/gui/tools/osm_downloader_dialog.py
@@ -74,6 +74,7 @@ class OsmDownloaderDialog(QDialog, FORM_CLASS):
 
         self.iface = iface
         self.buildings_url = "http://osm.linfiniti.com/buildings-shp"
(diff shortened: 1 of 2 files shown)


timlinux:
LGTM

PR #1733, opened 2015-03-09
fix #1732 : OSM downloader : skip a feature type if one is empty

### LINKED ISSUE #1732: The OSM downloader stops if one feature doesn't have any data
# Problem
If a user requests "all features" (buildings then roads), the downloader will stop if the extent doesn't contain any buildings even if there are some roads.
 
# Solution
The downloader should skip one feature and continue to the next feature if the is no data.

Gustry:
fix #1732 : OSM downloader : skip a feature type if one is empty

### DIFF
--- a/safe/common/exceptions.py
+++ b/safe/common/exceptions.py
@@ -259,6 +259,11 @@ class ImportDialogError(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


timlinux:
LGTM thanks!

PR #1736, opened 2015-03-10
fix #1735 : improvements about the OSM downloader's progress dialog

### LINKED ISSUE #1735: OSM Downloader progress bar dialog should be more informative
# Problem

Currently the OSM downloader does not tell you very much information:

inasafe openstreetmap downloader_180

And the default size is quite small (I had to expand it for the above screenie).

# Proposed solution

- [x] Update the progress numbers to decimal megabytes if total download is to be more than 1mb, decimal kb if dowload is less than 1 mb
- [x] Add some text after the bytes done / bytes total indicating units are in kb or mb e.g. ``10.1kb of 300kb``
- [x] Indicate which file number is downloading e.g. ``Fetching building points (file 1 of 3)```
- [x] Make the progress dialog larger to start with so that the above information can be seen without resizing.

Gustry:
fix #1735 : improvements about the OSM downloader's progress dialog
osm-downloader3

### DIFF
--- a/safe/common/test/test_utilities.py
+++ b/safe/common/test/test_utilities.py
@@ -34,6 +34,7 @@ from safe.common.utilities import (
     temp_dir,
     log_file_path,
     romanise,
+    humanize_file_size,
     add_to_list)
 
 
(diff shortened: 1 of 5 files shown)


timlinux:
Yeah that looks awesome - thanks @Gustry . LGTM

PR #1740, opened 2015-03-10
fix #1739 exception about canceled download

### LINKED ISSUE #1739: Download canceled by user raise a wrong exception
# Problem
When a user cancels a download, a ``DownloadError`` is raised, or according to [the code and comments](https://github.com/AIFDR/inasafe/blob/develop/safe/gui/tools/osm_downloader_dialog.py#L332), it should be a ``CanceledImportDialogError``

Gustry:
An ``CanceledImportDialogError`` is raised instead of ``DownloadError`` when a download is canceled.


timlinux:
LGTM

PR #1768, opened 2015-03-25
fix unicode in file path

Gustry:
fix unicode in file path when saving to a specific directory

### DIFF
--- a/safe/storage/raster.py
+++ b/safe/storage/raster.py
@@ -288,7 +288,7 @@ class Raster(Layer):
         #              precision even though Float64 is specified
         #              - see issue #17
         driver = gdal.GetDriverByName(file_format)
-        fid = driver.Create(filename, M, N, 1, gdal.GDT_Float64)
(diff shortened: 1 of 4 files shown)


safe/storage/raster.py
```diff
@@ -288,7 +288,7 @@ def write_to_file(self, filename):
         #              precision even though Float64 is specified
         #              - see issue #17
         driver = gdal.GetDriverByName(file_format)
-        fid = driver.Create(filename, M, N, 1, gdal.GDT_Float64)
+        fid = driver.Create(filename.encode('utf-8'), M, N, 1, gdal.GDT_Float64)
```
akbargumbira: just a small comment, can you change it to use get_unicode from safe.utilities.unicode? it's gonna be easier to see which part of the codes that we patch by getting its usage..

Gustry: Oh yes, sorry ! Thanks for this function.

Issue #1179, opened 2014-08-12
RFC: for people in buildings


Charlotte-Morgan:
# Problem

InaSAFE impact functions assessing the impact of hazards on people rely on raster population data. There are limited sources of raster population data (Landscan 2008, AsiaPop / WorldPop) and many InaSAFE users have reported that the analysis results based on these population data are inaccurate. 

This RFC explores one idea for improving the accuracy of affected population results that InaSAFE produces for impacts on people.
# Duration

November - December 2014
# Current State

This proposal is in **draft - please contribute**.
# Proposers
- Charlotte Morgan (@Charlotte-Morgan)
- Christian Christelis (christian@kartoza.com @cchristelis)
# Detail

InaSAFE impact functions assessing the impact of hazards on people rely on raster population data. There are limited sources of raster population data (Landscan 2008, AsiaPop / WorldPop) and many InaSAFE users have reported that the analysis results based on these population data are inaccurate.  Indonesian users are keen to be able to use Indonesian census data to determine affected populations. At present Indonesian census data (as in many parts of the world) are captured and distributed as polygon data. The challenge is how to turn these census data into a meaningful dataset that can be used in InaSAFE. & to improve the assessment of affected populations.

There are at least four options:
1. Rasterise the census data without any expert input regarding population distribution within the census collection area.
2. Supply high resolution census population data to population modelling experts and get them to generate raster data suitable for use in InaSAFE.
3. Make InaSAFE run on vector population data.
4. In areas where buildings have been mapped, statistically assign the population to buildings.

This RFC seeks to explore option 4 above in more detail and includes two issues:
1. How to generate a new combined exposure data set from existing population and buildings data. This is the focus of this ticket but should not forget how the data will be used in IF in the future.
1. The need for a new IF to assess impact on people when the exposure data consists of people in buildings. New IF should also consider the capacity to include both building and population vulnerability attributes if and when these are available some time in the future. This component should be expanded as a separate ticket.
# Proposed solution

That we limit consideration for this new exposure data set to the combination of OSM buildings and census population data at the highest capture resolution (smallest block sizes).
That we assume the census records 'usual place of residence' and as such is where people live / sleep.

OSM buildings will need a certain number of attributes to be populated (filled in) in order to support this proposal. OSM buildings will need to be classified by type (those used for residential purposes) and include attributes for size (footprint and number of floors used for residential purposes).

That the population could be statistically distributed between the buildings based on building floor area.
eg 
[...]

PR #1769, opened 2015-03-27
Feature/inasafe implement #1179 people in buildings

cchristelis:
New people in buildings tool.


safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+from safe.utilities.resources import get_ui_class
+
+FORM_CLASS = get_ui_class('people_in_buildings_base.ui')
+
+People_Calculated = "People_Dec"
+People_Rounded = "People_Rnd"
+
+
+class PeopleInBuildingsDialog(QtGui.QDialog, FORM_CLASS):
+    """Options dialog for the InaSAFE plugin."""
```
timlinux: cut and paste

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def __init__(self, iface, parent=None):
+        """Constructor for the dialog.
+
+        :param iface: A Quantum GIS QGisAppInterface instance.
```
timlinux: Quantum GIS -> QGIS

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        :type iface: QGisAppInterface
+
+        :param parent: Parent widget of this dialog
```
timlinux: Please end comment lines with full stops for consistency throught docstrings.

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        """Load the layer options and column options into the combobox.
+        """
+        layer_names = self._get_layer_names()
+        self._update_combobox(self.buildingLayerComboBox, layer_names)
+        self._update_combobox(self.censusLayerComboBox, layer_names)
+        self.handle_building_layer(self.buildingLayerComboBox.currentText())
+        self.handle_census_layer(self.censusLayerComboBox.currentText())
+
+    def _select_layer_by_name(self, layer_name):
+        """Get a layer by name
```
timlinux: First line of docstring should be a full sentance of < 80 chars followed by a newline.

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        index = layer_names.index(layer_name)
+        return layers[index]
+
+    def _get_layers(self):
+        """Get all layers.
+
+        :returns: All layers currently loaded.
+        :rtype: list
+        """
+        layers = self.iface.legendInterface().layers()
```
timlinux: I think this will only return visible layers. If you truly want all layers get them from the QgsMapLayerRegistry

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        return layers
+
+    def _get_layer_names(self):
+        """Get the names of all layers.
+
+        :returns: A list of all layer names
```
timlinux: .

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        layers = self._get_layers()
+        layer_names = [l.name() for l in layers]
+        return layer_names
+
+    @staticmethod
+    def _get_field_names(layer):
+        """Get all attribute names of the layer.
+
+        :param layer: The layer to be investigated.
+        :type layer: VectorLayer
```
timlinux: QgsVectorLayer

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+
+        :returns: Returns a list of all attribute names of the layer.
+        :rtype: list
+        """
+        field_names = [field.name() for field in layer.pendingFields()]
+        return field_names
+
+    def handle_building_layer(self, layer_name):
```
timlinux: Is this a slot? Normal naming convention would be ``building_layer_changed(self, layer_name)``

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        :type layer_name: basestring
+        """
+        layer = self._select_layer_by_name(layer_name)
+        if layer is None:
+            return
+        field_names = self._get_field_names(layer)
+        self._update_combobox(self.usageColumnComboBox, field_names)
+        self._update_combobox(self.levelsColumnComboBox, field_names)
+
+    def handle_census_layer(self, layer_name):
```
timlinux: cencus_layer_changed

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+    def handle_building_layer(self, layer_name):
+        """Handler for change buildings layer event.
+
+        :param layer_name: The name of the layer that was selected.
+        :type layer_name: basestring
+        """
+        layer = self._select_layer_by_name(layer_name)
+        if layer is None:
+            return
+        field_names = self._get_field_names(layer)
```
timlinux: We could probably refactor the above lines into a helper : ``get_field_names`` since the same logic is used below

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        self._update_combobox(self.populationCountComboBox, field_names)
+
+    @staticmethod
+    def _update_combobox(combobox, options):
+        """Shorthand for clearing and loading options (this happens a lot)
```
timlinux: .

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        """
+        combobox.clear()
+        combobox.addItems(options)
+
+    @staticmethod
+    def _add_attribute_to_layer(layer, attribute_name, data_type):
+        """Add a new attribute to an existing layer.
+
+        :param layer: The layer to be altered.
+        :type layer: VectorLayer
```
timlinux: QgsVectorLayer

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+
+        :param data_type: The type of the new attribute.
+        :type data_type: QVariant.Double, QVariant.Int
+
+        :returns: The success state of the operation.
+        :rtype: bool
+        """
+        provider = layer.dataProvider()
+        # Check if attribute is already there, return "-1" if not
+        ind = provider.fieldNameIndex(attribute_name)
```
timlinux: field_index

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        if ind != -1:
+            return False
+        field = QgsField(attribute_name, data_type)
+        result = provider.addAttributes([field])
+        return result
+
+    def add_population_attribute(self, layer):
+        """Add people estimate columns/attributes to the output layer.
+
+        :param layer: The layer that the population attribute should be added.
```
timlinux: Layer that the population attribute should be added to.

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+
+        :return: A lookup of the field values by field name.
+        :rtype: dict
+        """
+        attributes = feature.attributes()
+        attribute_dict = dict(zip(field_names, attributes))
+        return attribute_dict
+
+    @staticmethod
+    def _overlapping_area(feature, layer):
```
timlinux: I thought we had said we would use building centroids for determining if a building is flooded etc. so that we can prevent double counting using overlaps?

cchristelis: It is using the centroid. Although the method name might be a bit confusing. I'll change that.

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        geometry_feature = feature.geometry()
+        geometry_layer = layer.geometry()
+        centroid = geometry_feature.centroid()
+        if not centroid.within(geometry_layer):
+            return 0
+        return geometry_feature.area()
+
+    @staticmethod
+    def _get_levels(attributes, levels_attribute):
+        """Try to extract the levels from the attributes dict as an int.
```
timlinux: CRLF

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        ..Note: If the levels of this building has not been set we will
```
timlinux: .. note::    (this is rst / sphinx markup and will get rendered as a nice bubble)

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        assume it to be 1.
+
+        :param attributes: The attributes dict from the feature
+        :type attributes: dict
+
+        :param levels_attribute: The name of the column that contains layer
+        information
```
timlinux: indent and full stop.

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+
+        :param building_use: The name of the attribute containing the
+            building use.
+        :type building_use: basestring
+
+        :returns: The residential proprtion
+        :rtype: int, float
+
+        ...Note: The residential proportion is based on the building type.
+        Currently the types 'Residential' and 'House' are considered to be
```
timlinux: .. note:: and subsequent lines need indenting for rst markup compliance

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        if use in ['Residential', 'residential', 'house', 'House']:
+            return 1
+        # if use == '' or use is None or isinstance(use, QPyNullVariant):
+        #     # Default to residential, if it is unclassified
+        #     return 1
+        return 0
+
+    @staticmethod
+    def _feature_fully_in_extent(layer, feature):
+        """Determine wether the feature is fully within the layer's extent.
```
timlinux: whether

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+
+        :param layer: The layer to be considered.
+        :type layer: VectorLayer
```
timlinux: QgsVectorLayer

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        extent = layer.extent()
+        geometry_extent = QgsGeometry.fromRect(extent)
+        geometry_feature = feature.geometry()
+        intersection = geometry_feature.intersection(geometry_extent)
+        if intersection.area() < geometry_feature.area():
+            return False
+        return True
+
+    def estimate_people_in_buildings(self):
+        building_layer_name = self.buildingLayerComboBox.currentText()
```
timlinux: docstring - would be good to have a fairly detailed description here of what the underlying logic is.

safe/gui/tools/people_in_buildings_dialog.py
```diff
@@ -0,0 +1,380 @@
...
+        self.add_population_attribute(buildings_layer)
+        field_names_population = self._get_field_names(population_layer)
+        field_names = self._get_field_names(buildings_layer)
+        field_names += [People_Calculated, People_Rounded]
+
+        #TODO: Make a better progress estimator
+        self.progressBar.setValue(0)
+        progress = 1
+        progress_increment = 100.0/len(
+            [p for p in population_layer.getFeatures()])
```
timlinux: ``population.featureCount()`` is probably much cheaper

safe/gui/tools/test/test_people_in_buildings_dialog.py
```diff
@@ -0,0 +1,213 @@
...
+        self.assertIn(buildings_name, buildings_options, message)
+        message = (
+            'The population layer %s has not been loaded into the people '
+            'in buildings dialog' % population_name)
+        self.assertIn(population_name, population_options, message)
+        message = 'The layers in the layer dropdowns should be equal'
+        self.assertItemsEqual(buildings_options, population_options, message)
+
+    def test_0002_building_layer_and_attributes(self):
+        """Ensure the building attributes are being assigned to the
```
timlinux: the ....?

timlinux:
Thanks @cchristelis - I have only pedantic suggestions so far. Will be nice if you can address them to add some extra polish to our code base. For the UI I will make separate comments.

cchristelis:
Thanks @timlinux for taking the time to read through this. Especially the comments like using population.featureCount(), rather than looping through the features are quite valuable!

safe/gui/tools/people_in_buildings_dialog.py
cchristelis: I assume you mean visible layers are layers, which are 'checked' in QGIS? I tried it returns all layers including the unchecked ones.
I'm changing it to QgsMapLayreRestitry anyway though...

safe/gui/tools/people_in_buildings_dialog.py
cchristelis: I thought about it, but don't think that it is worth it, in this case. We will still have to have the if statement and the return. So we are creating a helper function etc with some complicated condition to distinguish between no fields and no layer and return that and handle it. So as a whole it doesn't seem worth it.

safe/gui/tools/people_in_buildings_dialog.py
cchristelis: Thanks!

Charlotte-Morgan:
Hi @cchristelis - thank you so much for progressing this to the stage its at & the opportunity to test it.
I will separate comments into analysis results and user interface.
Analysis Results
- [ ] FYI i used the sample data we provided and also OSM building polygons downloaded with the OSM downloader as i think this is how we 'expect' the buildings to be obtained for this tool.
- [ ] the analysis proceeds as per the RFC - ie if a building has its centroid within the extent of the census boundary then it is included. In all cases, boundary buildings were handled well. However, some buildings completely within the census area were not included in the analysis results.My expectation is that all input buildings with centroids within the census layer will be exported regardless of type and whether people are assigned.
image
In this image: orange (original buildings) blue (PIB output). Some orange buildings should have also been included in the PIB output.
- [ ] Checksums. The total value of the rounded and unrounded columns differs (20575 vs 20563) and the total population of the area was 20573. So 10 people were lost somewhere?? (A small error - I know, unless it's your family :))
-----------------------------------------------------------------------------------------------------------------------------------
User Interface
- [ ] is it possible to allow the user to step out of the PIB tool to interact with the interface (ie open the attribute table to look for suitable fields) as per the "Set analysis area" tool?
- [ ] Is it possible to set the default datasets and fields for the buildings data especially if we are assuming that the input has been obtained from OSM download. ie look for default buildings data and a TYPE field for usage and LEVELS field for levels.
- [ ] do we have consistency in terminology in other bits of the software ie do we use layer and column elsewhere or do we use dataset and attribute (i don't know the answer)
- [ ] Can you add a 'People In Buildings' line to go with the output new layer checkbox?
- [ ] Should the user be able to navigate to where the new layer is saved?
- [ ] other comments on text strings are in the files directly - if i can work it out

Thanks again -  you have build what we asked for, a lot of these comments are superficial. I'm guessing some are easy and some are not worth doing.

Charlotte-Morgan:
Wah - can not comment on the relevant files so here some tweaking for UI text:
Disclaimer:
Please note this tool requires that all buildings have been mapped within the census areas and that the census data are available for the entire study area.

How to use the tool:
1) Load a buildings layer and a census layer in to QGIS.
2) The buildings layer will need two specific attributes to make full use of this tool:
An attribute for building type or building use. The default attribute for OSM building polygons downloaded with the OSM_downloader is 'type'. Only buildings with the type of "Residential" or "House" will be assigned people.Where this attribute is not populated, a default of 'non-residential' is assumed. No people will be assigned in this case.
An attribute for the number of residential floors or levels. The default attribute for OSM building polygons downloaded with the OSM_downloader is 'level'. Where this attribute is not available or not populated, a default value of 1 is used.
3) The census layer will need to have an attribute for population count.
4) The output layer can be written to the current buildings layer or to a new layer.
5) The output layer will have two additional attributes: People_Rnd and People_Dec. The People_Rnd is the value that will be used as a reference for the number of people estimated to reside in the residential building. The People_Dec is a value that will only be used by the Impact Function.

cchristelis:
Hi @Charlotte-Morgan,

thanks for having such a detailed look at this. I thought I had already replied to this when I started looking at it last Friday, but it seems like I didn't actually post my comment :p. 

Internal Workings:

- Yes, OSM is the expected, but not exclusive, source of buildings data.

- Disappearing buildings. So the buildings are not actually being clipped. The buildings with no people assigned are not being assigned any value at all. This is effectively a Null. It seems like with the styling applied, features with a Null value are not shown. Would we expect to clip the resulting PIB layer?

- Check sums (People_Dec): I ran the same analysis and got 20573 people as the sum of the decimal fields. It strikes me as odd, that if you got a whole number discrepancy. Would it be possible to send me the input and resulting layers please. I would like to see if I can pin-point the source of this discrepancy. 

- Check sums (People_Rnd): The sum of this field will almost always give a discrepancy and should be treated as human, and for viewing displaying purposes only. The contrived scenario of 4 people living in 3 equal houses shows this. The People_Dec will assign 1.333... people to each buildings, while People_Rnd 1. Hence the discrepancy here is 3 vs the actual 4.

UI

- User stepping out: This has been done.

- Setting defaults: This will enhance the user experience, I will action accordingly.

- Naming consistency: I had a bit of a chat with @timlinux, about consistency in presenting help texts, and I think it will be well worth the while to assure these are the same.

- 'People In Buildings' line: I'm not quite sure what this means.

- Allow users to navigate to new layer output: I think this is a good idea and worth implementing.

- Text improvement: I will change accordingly.

Thanks again for all the input!

Charlotte-Morgan:
Hi @cchristelis - I know how you feel when you forget to post comments :) I have done it too.
- [ ] Regarding the PIB output layer & disappearing buildings -  I have mixed feelings.
If the goal is to make a new exposure data set that can be used in subsequent IF analyses; then all the buildings considered (ie have their centroid within the census area) should be written to the output, assigned a value & be visible in the display. In this sense the NULL option might be best used for buildings that fall outside. Is it reasonable to give non residential buildings that are inside the census area a People_Rnd value of 0 & a People_Dec value of 0.0000 as these buildings will / must be considered in subsequent impact functions if we are to ask, how many buildings of what type and how many people are affected in a single IF analysis.
Whether the data are clipped ie the NULL outside buildings are not written to the output is an option. I prefer to keep them (I think) rather than clipping. But clipping is also a reasonable thing to do.

- [ ] checksums - I am already using the same data that you are so I wont sent it again.

- [ ] the UI People in Buildings line. Can you add "People in Buildings" as text under Population Count Column as a new row aligned with the  "output in New Layer" check box? Line is the pic below.

image

cchristelis:
Target branches cannot be updated in PR's see #2124.

PR #1771, opened 2015-03-30
fix #1770 check if the map canvas is using 4326 and enable OTF if necessary

### LINKED ISSUE #1770: OSM Downloader should always use extent in EPSG:4326
# Problem

Currently if I use e.g. EPSG:3857 and try to use the downloader, it displays coordinates in m based on the spherical mercator projection and the download fails.

inasafe openstreetmap downloader_023

# Proposed solution

- [x] For the OSM Downloader ensure that the CRS for extent coordinates is always EPSG:4326
- [x] Ensure that for extent selector, existing behaviour is not broken

Gustry:
fix #1770 check if the map canvas is using 4326 and enable OTF if necessary


timlinux:
LGTM

PR #1864, opened 2015-04-22
Fix #1854: Modified aggregation to use impact layer extent 

### LINKED ISSUE #1854: Fix zero post processor in Earthquake Polygon + Population
When we ran Earthquake Polygon Hazard and Population Exposure using the test data, the report succesfully computed but the postprocessor returns zero results.

lucernae:
Modify aggregation to use impact layer extent if aggregation layer is not provided.

This fix is used to handle cases where the hazard vector layer contains polygon which is outside the optimal extent but must be included in the postprocessor calculations.

### DIFF
--- a/safe/impact_statistics/aggregator.py
+++ b/safe/impact_statistics/aggregator.py
@@ -124,7 +124,7 @@ class Aggregator(QtCore.QObject):
             'inasafe/use_native_zonal_stats', False, type=bool))
         self.use_native_zonal_stats = flag
 
-        self.extent = extent
+        self._extent = extent
         self._keyword_io = KeywordIO()
         self._defaults = get_defaults()
         self.error_message = None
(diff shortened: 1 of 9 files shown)


akbargumbira:
can we ask @mbernasocchi if this is a wise decision?

safe/impact_statistics/aggregator.py
```diff
@@ -171,6 +171,23 @@ def __init__(self, extent, aggregation_layer):
...
+        return self._extent
+
+    @extent.setter
+    def extent(self, value):
+        self._extent = value
+        # update layer extent to match impact layer if in aoi_mode
+        if self.aoi_mode:
+            try:
+                self.layer = self._extents_to_layer()
+                self.safe_layer = safe_read_layer(str(self.layer.source()))
```
akbargumbira: Remove casting it to str. It will raise unicode error if the layer source's got some non ASCII

lucernae: Then, I will take note that we should also change the same statement in _prepare_layer

lucernae:
Perhaps I should wait for further discussion for this PR

mbernasocchi:
All Looks good to me

PR #1872, opened 2015-04-27
Fix layer source used converted as unicode

lucernae:
Layer source string was previously converted using str() cast.
By @akbargumbira 's [comments](https://github.com/AIFDR/inasafe/pull/1864#discussion_r28900048), we should change the cast to unicode


akbargumbira:
No need to cast to unicode. It returns unicode already.

lucernae:
Ah yes, I think I misunderstood your comment.

lucernae:
Close this because it already fixed by #1874 :)

Issue #1851, opened 2015-04-21
Philippine Minimum Needs profile break the options dialog.


cchristelis:
When selecting the Philippine Minimum Needs profile and selecting "Earthquake Polygon" and "Population" exposure. Opening the Options dialog and closing it again via "OK" causes a python error.

Jannes123:
Looks like there are still some old-style parameters hanging around or merged in.  Will fix those first I think that will solve most of the problems.

Jannes123:
@lucernae it seems like the get_parameter function returns (0.0)  when called in function_options_dialog.py

lucernae:
@Jannes123 I can't replicate the bug. But I do notices that I can't change Family Kits value. Is that also a bug? :v

lucernae:
@Jannes123  I'm able to replicate it now. It seems dangerous because we can't modify some of the parameters because it already out of bounds.

Jannes123:
@cchristelis seems like the precision of the nrs remaim fixed at 2 for resource parameters.  So when you put a nr like 0,04 in its fine but 0.0003 will create that bug.

cchristelis:
Oh, so that is why we keep getting 0.00... I wonder whether we should increase the precision.

Jannes123:
@cchristelis  #1868  PR contains that update

Jannes123:
#1881 fixes this issue

PR #1878, opened 2015-04-28
Fix minimum needs dialog handles out of precision value

lucernae:
This PR relates to issue #1851 .
Branching from @Jannes123 's PR #1868  to include another precision check and translation string candidate.


lucernae:
I will close this, since it is already included in other's PR

Issue #1862, opened 2015-04-22
InaSAFE Impact Merge Tools Error


adissadis:
# Problem
I'm trying to use impact merge tool to combine population and building in Jakarta, but when i click OK it show this dialog:
![inasafe_1](https://cloud.githubusercontent.com/assets/2520159/7271867/b2d6b202-e910-11e4-9b5d-d056f3db702f.png)
It said that it's being used by another process but I'm don't know what process that open this file. I didn't print the impact result and open in pdf.

# Platform
Windows 7 64 bit QGIS 2.8.1 InaSAFE 3.1.0 dev-master

lucernae:
@adissadis I'm not be able to reproduce it. Could you tell me which layer did you use in the merge? Are you using QGIS 64 bit?

adissadis:
@lucernae sorry for my late respond. I'm still encountered this error, even i update the plugin according to recent file in this repository, i'm still got this error.

Here my step by step:
1. I run impact analysis using building_osm_jakarta (from InaSAFE training data) and flood_hkv_jakarta (from InaSAFE training data) using district_osm_jakarta as aggregation layer.
2. I run impact analysis also using popmap10_all and flood_hkv_jakarta and also using district_osm_jakarta as aggregation layer.
3. Both of impact analysis result i merge it using InaSAFE impact merge tool with first impact layer is Population and second impact layer is building affected and using district_osm_jakarta as aggregation layer

That's my step before the error appear.

Here's the newest screenshot using recent InaSAFE from repo and using QGIS 2.8.1 in Windows 7 64 bit:
![inasafe_1](https://cloud.githubusercontent.com/assets/2520159/7558414/ef444238-f7cd-11e4-94fb-585f1af63694.png)


Also, i dont know whether this information from log message need or not:

2015-05-11T10:53:21	1	warning:C:/Users/Sony/.qgis2/python/plugins\inasafe\safe\gui\tools\impact_merge_dialog.py:944: DeprecationWarning: QgsAtlasComposition.setComposerMap() is deprecated
			  atlas.setComposerMap(composer_map)
			
			traceback:  File "C:/Users/Sony/.qgis2/python/plugins\inasafe\safe\plugin.py", line 587, in show_impact_merge
			    dialog.exec_()  # modal
			  File "C:/Users/Sony/.qgis2/python/plugins\inasafe\safe\gui\tools\impact_merge_dialog.py", line 289, in accept
			    self.merge()
[...]

PR #1913, opened 2015-05-04
Fix impact merge race condition

lucernae:
Related to issue #1862 .

I can't reproduce the error, but I add some code to avoid race conditions.


timlinux:
Test this please, jenkins

PR #1923, opened 2015-05-07
add IDP camp export and improve GUI

Gustry:
Add two options to download potential IDP and boundaries.
osm_downloader

Shapefiles need to be improved on osm-reporter.
All options are checked by default, except boundary.

The combobox about the country is optional. That's why I put it in a sentence which is italic.

### DIFF
--- a/resources/osm/admin_level_per_country.json
+++ b/resources/osm/admin_level_per_country.json
@@ -1,4 +1,28 @@
 {
+  "Afghanistan": {
+	"1": "N/A",
+	"2": "National Border",
+	"3": "N/A",
+	"4": "Province (ولايت wilayat)",
+	"5": "N/A",
+	"6": "Districts (Wuleswali)",
+	"7": "N/A",
+	"8": "N/A",
+	"9": "N/A",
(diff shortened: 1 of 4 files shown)


timlinux:
Looking good!

camp --> camps
Will the admin level text wrap nicely if the combo box is wide?

timlinux:
Etienne we need to have some tests for these features - please look at the other OSM downloader tests and emulate for IDP camps and admin areas. Also I have some thoughts on the styling:

I was thinking we could have  a single QML style file for all 8 levels, and then have rules so that each level gets it's own style. We can then ship the same QML file regardless of what level the user has chosen. What do you think?

Gustry:
Some tests has been added. Please check them if it's enough.
There is not a lot of test to do about IDP.

About the QML, it's on the osm-reporter side. So only the overpass query will be different between each boundary export ?

timlinux:
@Gustry please go ahead and merge this once tests are passing. Yes I think only the overpass query should differ.

timlinux:
LGTM

Issue #1931, opened 2015-05-12
Minimum Needs Configuration - Provenance statement


KathrynOwen:
# Problem
A user can modify a profile without updating the provenance statement

# Proposed Solution
- [ ] User needs to be prompted to check that the Provenance statement is still current when saving an updated or modified minimum needs profile.
- [ ] This is a potentially a good place for a user warning message

Jannes123:
Hi @KathrynOwen what is the text message you would like to see after the Warning?
![profile_changed](https://cloud.githubusercontent.com/assets/11268911/7630034/5fe2a186-fa32-11e4-842b-d9a87997621e.png)

KathrynOwen:
Hi @Jannes123 Perhaps 'Warning: Changes have been made to the above profile. Is the following provenance statement still correct?'

@Charlotte-Morgan what do you think?

Charlotte-Morgan:
LGTM or if you want to keep it shorter.
Warning: profile changed. Please update the provenance statement.

PR #1943, opened 2015-05-15
Fix 1931

Jannes123:
This PR contains the fix for issue #1931

@cchristelis please review

### DIFF
--- a/safe/gui/tools/minimum_needs/needs_manager_dialog.py
+++ b/safe/gui/tools/minimum_needs/needs_manager_dialog.py
@@ -21,6 +21,7 @@ from os.path import expanduser, basename
 # noinspection PyUnresolvedReferences
 # pylint: disable=unused-import
 from qgis.core import QGis  # force sip2 api
+from PyQt4 import QtCore
 
 # noinspection PyPackageRequirements
 from PyQt4 import QtGui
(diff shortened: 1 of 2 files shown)


safe/gui/tools/minimum_needs/needs_manager_dialog.py
```diff
@@ -49,6 +50,8 @@
 from safe.utilities.resources import resources_path, get_ui_class
 from safe.messaging import styles
 from safe.gui.tools.minimum_needs.needs_profile import NeedsProfile
+import logging
+LOGGER = logging.getLogger('InaSAFE')
```
cchristelis: Maybe you should leave this out of production code.

safe/gui/tools/minimum_needs/needs_manager_dialog.py
```diff
@@ -286,6 +297,11 @@ def mark_current_profile_as_pending(self):
         item = self.profile_combo.model().item(index)
         item.setForeground(QtGui.QColor('red'))
 
+    def profile_changed_notification(self):
+        """Un-hides the profile changed warning.
+        """
+        self.profile_changed.show()
```
cchristelis: What happens when the user changes profile?

cchristelis:
Hi @Jannes123, have a look at my comments. I would expect that the user might save and save as, in which cases the note should be removed. Also please add same tests to follow the new logic flow.

PR #1940, opened 2015-05-13
this closes #1926 #1930 1933

Jannes123:
Minimum needs bug fixes:
* Needs Manager resize
* Dropdown list populates with correct minimum needs profile
* More space for Readable Sentence

Issue #1933, opened 2015-05-12
Minimum Needs Manager - issue with default profile upon opening


KathrynOwen:
# Problem
Profile details do not match the profile listed in the Profile box when the Minimum Needs Manager is opened for a second time in a session. This could result in edits being applied to the wrong profile.

# Details
First time the Minimum Needs Manager is opened after opening QGIS, the default Profile and matching details are displayed - e.g. 'BNPB_en'

![image](https://cloud.githubusercontent.com/assets/10971278/7580957/9ede46ba-f8c0-11e4-811b-ca27f31e23d8.png)

Then if the profile is changed (e.g. to 'Kats_test') and the minimum needs tool is closed, the next time it is opened, the default details are shown (i.e. BNPB_en') but the Profile name in the profile box is the one that was selected last (i.e. 'Kats_test) - so they do not match.

![capture](https://cloud.githubusercontent.com/assets/10971278/7581002/ff8e0e82-f8c0-11e4-83c8-4126eb8b39c0.JPG)

PR #1950, opened 2015-05-20
Fix for minimum needs User Interface errors

### LINKED ISSUE #1944: [BUG] Cannot resize Global Minimum Needs Configuration Window
# Problem
I cannot resize Global Minimum Needs Configuration's Window vertically. In the first time the window appear was really big that i need to resize it to find OK button that located in bottom window, but I only can resize the window horizontally and when i want to resize the window vertically, i can't resize it. Here's the screenshot for initial window appear:
inasafe_1

# Platform
I'm running InaSAFE 3.2.0.dev-master in QGIS 2.8.2 with Windows 7 64bit as OS

lucernae:
fix #1944: allows a scrollbar so the user doesn't have to resize window when it exceeds the OS windows

I only add a vertical scrollbar for this commit.

### DIFF
--- a/safe/gui/tools/minimum_needs/needs_manager_dialog.py
+++ b/safe/gui/tools/minimum_needs/needs_manager_dialog.py
@@ -209,6 +209,7 @@ class NeedsManagerDialog(QDialog, FORM_CLASS):
         See https://github.com/AIFDR/inasafe/issues/1387
         """
         if self.stacked_widget.currentWidget() == self.resource_edit_page:
+            self.edit_item = None
             self.switch_context(self.profile_edit_page)
         else:
             super(NeedsManagerDialog, self).reject()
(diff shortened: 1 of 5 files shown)


Jannes123:
This is very strange because on my develop branch the window can resize and  scrollbar is also there.
I'm using ubuntu 14.04 and QGIS2.8

lucernae:
@Jannes123 Wow? is that so? perhaps your develop is ahead of AIFDR develop and haven't been merged yet, so AIFDR develop doesn't have this commit.

In any case, I'm trying to group up other bugs related with minimum needs UI reported by @adissadis

Jannes123:
Please look at PR #1940 deals with 3 bugs

and  PR#1943 which still needs some work

Yes there are changes to my develop which is still work in progress but
those bugfixes were submitted in seperate PRs

lucernae:
That fix in #1940 is great, we should also include it on develop branch, also for PR #1943.

lucernae:
@Jannes123 I will let you handle issue #1933 to merge it to develop since it hasn't been merged

safe/gui/tools/minimum_needs/needs_manager_dialog.py
```diff
@@ -8,6 +8,8 @@
      (at your option) any later version.
 
 """
+from safe.utilities.i18n import tr
```
akbargumbira: Put this import below

PR #1960, opened 2015-05-26
change mmi range and augment test

akbargumbira:


safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
```diff
@@ -115,7 +115,7 @@ def __init__(self):
                 1: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 1.0,
                 7: 1.0, 8: 1.0, 9: 1.0, 10: 1.0
             }),
-            ('mmi_range', range(2, 10)),
+            ('mmi_range', range(4, 11)), # from MMI 4 to 10
```
akbargumbira: is there any consideration why you specify the mmi range only from 4? we also need to update the report (see http://realtime.inasafe.org/en/20150523143408-en.pdf)

dynaryu: I think MMI < 4 is not likely to make any fatality. Also note that mmi < 6 is ignored in the current estimation of displaced number.

akbargumbira: yes, but we also need to calculate exposed people from the range [2,10]. Are you suggesting that we should also show the exposed people from the MMI 4 in the report?

dynaryu: If the information is required, I have no objection. I can change it accordingly then.
Another thing is that I wonder whether exposed number can be pulled within +/-0.25 MMI which is more consistent with other records such as EXPO-CAT. For instance 4.75-5.25 is rounded to 5.0.

dynaryu: Then, MMI range will be np.arange(4,10.25,0.5)

akbargumbira: We can specify it with the step parameter. It is now set to 0.5 which will pull +/-0.5 for each MMI e.g for calculating the exposed, displaced, and fatality for MMI 2, it will pull all the MMI in interval (1.5, 2.5].

dynaryu: I guess it's hard coded now. And wonder if others are happy with the change from 0.5 to 0.25.

dynaryu: Maybe we can use 0.25 for internal computation, and still display result for each MMI.

akbargumbira: yeah, it was in the Option tab and we removed all the nitty-gritty variables from that. For changing the step, perhaps @timlinux or @Charlotte-Morgan can comment on that

safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
```diff
@@ -111,11 +111,8 @@ def __init__(self):
         self.hardcoded_parameters = OrderedDict([
             ('x', 0.62275231), ('y', 8.03314466),  # Model coefficients
             # Rates of people displaced for each MMI level
-            ('displacement_rate', {
-                1: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 1.0,
-                7: 1.0, 8: 1.0, 9: 1.0, 10: 1.0
-            }),
-            ('mmi_range', range(2, 10)),
+            ('displacement_threshold', 6.0), # mmi threshold for displacement
+            ('mmi_range', range(2, 11)), # from MMI 4 to 10
```
timlinux: Comment needs updating: ``from mmi 2 to 10``

timlinux:
Just a note from my side:

* The < 50 = 0 logic was added at the request of Ole / Jono / Trevor (I can't remember specifically which) because for the realtime maps we were getting some cases where low numbers of fatalities were being reported and they felt it gave an incorrect impression about the granularity with which the model results can be accepted.
* The changes look good, but there are some coding conventions that need to be adhered to before we can apply the changes. I will give you a PR against your own branch which you can then review (to understand how we are looking for docstrings to be formulated) and apply to your tree. That will automatically update this PR.

My apologies for the slow response to this PR.

Regards

Tim

CC @Charlotte-Morgan

timlinux:
I made a pull request to your repo: https://github.com/dynaryu/inasafe/pull/1 which fixes some style issues.

PR #1979, opened 2015-05-30
fix #1958: move provenance from IF option to minimum needs tab

### LINKED ISSUE #1958: Move provenance option into minimum needs section
# Problem
The provenance is best to put in the minimum needs option. Move this to minimum needs section with the value from the minimum needs file itself. User can update this as perhaps they change a bit the minimum needs before running the analysis. The provenance at first should be fetched from the minimum needs file itself.

lucernae:


timlinux:
LGTM

PR #1981, opened 2015-05-30
fix volcano polygon population metadata to not accept hazard point layer

lucernae:
If it is allowed to accept point hazard, the function can't handle it and raises exception.

@akbargumbira @ismailsunni if you are ok with this, just merge it.


akbargumbira:
Yes it shouldn't. Please merge it

ismailsunni:
Wuut... I also made the change, but I haven't push it. Ok, I will merge it.

PR #2009, opened 2015-06-04
Improve the boundary export in the OSM downloader

Gustry:
- Change the GUI about the political boundary.
- add the bbox to each country in the json file.
- detect if the center of the map canvas is in one of these countries.
  - if yes, select this country in the combobox.

osm_downloader


timlinux:
LGTM

PR #2016, opened 2015-06-08
Make flood vector-building impact function faster

### LINKED ISSUE #1889: We need to make Flood Polygon on Buildings faster
# Problem
Now that we removed the old style IF for this, we need to make the QGIS IF one faster. Running FLood Polygon on Buildings with Jakarta extent takes intolerable time for user to wait.

# CC
Perhaps @cchristelis or @timlinux can address this?

wonder-sk:
This is done by using spatial index to identify which flood polygons
will need exact intersection test (much faster than doing intersection
with one huge polygon)

A small extra optimization is to write features in batches as that is
more efficient than writing features one by one.

Measured speed improvement on Jakarta test data (2013 flood polygons
and OSM buildings) was ~40x (from 30min down to 40sec)

Fixes #1889


timlinux:
Great stuff! LGTM

Issue #2018, opened 2015-06-08
IF options help text and description


lucernae:
We need help text and descriptions for several IF options and post processor:
1. Earthquake
   - Low Threshold
   - Medium Threshold
   - High Threshold
2. ITB Earthquake
   - Doesn't have any
3. Pager Earthquake
   - Doesn't have any
4. Generic Classified Polygon Building
   - Hazard Zone Attribute
5. Generic Classified Polygon Population
   - Hazard Zone Attribute
6. Generic Classified Raster Building
   - Low Hazard Class
   - Medium Hazard Class
   - High Hazard Class
7. Generic Classified Raster Population
   - Low Hazard Class
   - Medium Hazard Class
   - High Hazard Class
8. Generic Continuous Hazard Population
   - Categorical Thresholds
9. Flood Polygon Population
   - Evacuation Percentage
10. Flood Polygon Roads
    - Won't have any
11. Flood Raster OSM Building
    - Thresholds
12. Flood Raster Population
    - Thresholds
13. Flood Raster Roads
    - Min Threshold
    - Max Threshold
14. Flood Vector Building 
    - Won't have any
15. Tsunami Population Evacuation
    - Thresholds
16. Volcano Point Building
    - Distance
    - Volcano Name Attribute
17. Volcano Point Population
    - Distance
    - Volcano Name Attribute
18. Volcano Polygon Building
    - volcano name attribute
19. Volcano Polygon Population
    - volcano name attribute

For the post processor part:
1. Gender
2. Age
3. Minimum Needs
# Notes

@timlinux @akbargumbira @ismailsunni :
I will update this from time to time, please comment on incorrect grammar/vocabulary.

ismailsunni:
@lucernae can you give some indication which part have you done and not done?

ismailsunni:
No longer valid for InaSAFE v4.

PR #2025, opened 2015-06-09
Develop if options revisions

lucernae:
Includes:
- Add descriptions for IF options
- Integrate IF options error message to display
- Add constraint and grouping for Age post processor via GroupParameter


lucernae:
test this please, Jenkins.

timlinux:
I'm not sure if you were planning to take care of it in this patch, but the road type field should still be moved off to coming from the keywords, not the IF options

screen shot 2015-06-09 at 23 53 13

timlinux:
Also @lucernae the design intent of paramters is to prevent user from making choices that are invalid. To achieve that we cant have e.g.:

screen shot 2015-06-10 at 00 04 02

Which allows user to add more classes than specified,  allows duplicate values etc. I think a collection of spinbox widgets would be more appropriate here.

lucernae:
@timlinux about the Road type options, yes it will also be removed (that's why I didn't put any description). But since the IF still got the value from the parameters, I still include it for now. The next step would be removing that from IF options and somehow acquire it from the keywords.

and about the UI, at first I only infer it from the type of parameter that the IF expects, since it expects a list. But we can change it to spinbox like ClassifiedRasterHazardPopulationFunction

lucernae:
test this please, Jenkins.

lucernae:
This PR deals with issue for: #2018 to add some descriptions to IF Options
Next stage is to remove some IF options and get it from the keywords.
Right now, this stage PR is sufficient.

PR #2031, opened 2015-06-10
Classified hazard vs polygon building with QGIS API

wonder-sk:
This PR makes the ClassifiedPolygonHazardBuildingFunction work by using QGIS API instead of using InaSAFE's own routines for geometry operations.

It has been discovered that InaSAFE's geometry operations do not handle polygons with holes correctly, the new functionality comes with unit tests that give slightly different results (differences were manually checked to confirm that the new results are correct).

This implementation may be faster than the original with many smaller hazard polygons, and slower than the original with few very complex polygons. The speed could be improved probably by using prepared geometries or by cutting complex polygons into smaller pieces.

### DIFF
--- a/safe/common/utilities.py
+++ b/safe/common/utilities.py
@@ -18,6 +18,8 @@ import colorsys
 from collections import OrderedDict
 # pylint: enable=unused-import
 
+from PyQt4.QtCore import QPyNullVariant
+
 from safe.common.exceptions import VerificationError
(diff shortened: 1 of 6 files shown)


wonder-sk:
@timlinux PR ready to be reviewed

timlinux:
Great stuff thanks @wonder-sk - I will go ahead and merge it.

PR #2039, opened 2015-06-11
Recent Wizard updates

borysiasty:


ismailsunni:
LGTM
Only one thing, see here: https://github.com/AIFDR/inasafe/issues/1956#issuecomment-111342070

ismailsunni:
hi @borysiasty 
For `Use invisible table to separate keys and values (keys to left column, values to right column)` it would be nice if we show the `name` of the keyword. We try to keep the `key` not visible from the user.

borysiasty:
@ismailsunni:
> One comment, for Use invisible table to separate keys and values (keys to left column, values to
> right column) it would be nice if we show the name of the keyword. We try to keep the key not visible
> from the user.

I use KeywordIO.read_keywords() to read the keywords from selected file, so probably it would be the simplest to have an keyword_name_for_value() API. I imagine it could just grab all dicts from the definitions.py and return the first one with matching key and existing name. Unless you have a better idea.

ismailsunni:
> I use KeywordIO.read_keywords() to read the keywords from selected file, so probably it would be the simplest to have an keyword_name_for_value() API. I imagine it could just grab all dicts from the definitions.py and return the first one with matching key and existing name. Unless you have a better idea.

Hmm, I think we will leave it for now. We will have an API for keyword in the near future.
Thanks 

LGTM

PR #2043, opened 2015-06-15
add inasafe version to the osm downloader

Gustry:


timlinux:
LGTM once tests pass

Gustry:
Sorry, this test always failed, because of my FR language.
I fixed the test, is it fine calling a function inside the test to get the right value ?

Issue #1216, opened 2014-09-23
RFC: Landcover (polygon hazard on polygon exposure)


akbargumbira:
# Problem

There is no land cover impact function in InaSAFE. Understanding the impacts of a hazard on landcover is important because we need to know the effects of a disaster on peoples livelihood assets (especially the crops) inside the land cover and food security.

![landcover - new page](https://cloud.githubusercontent.com/assets/178003/4368452/3b6a2178-42ed-11e4-899d-2db09d2d038f.png)

# Duration

TBD
# Status

This RFC is in **draft**.

# Proposers

@mbernasocchi 
@timlinux

# Detail

Land cover has some specific metrics we are interested in:

1. **Type of Land Cover**. This is a measure of what type of landcover affected by hazards
2. **Area of Land Cover**. This is a measure how many extents/areas affected by hazards

Specific for crops, we are also interested in : 

4. **Productivity**: This is a measure of crop yield e.g. 1000 kg rice / ha / annum.
5. **Economic value**: This is a measure of crop value e.g. Rp 100000 / rice / ha / annum.
6. **Area**. This is a measure of land cover/use total area e.g. 1000 ha or 100 m2
7. **type of plants**. This is a information what kind of crops existing or productive e.g rice, corn, etc
8. **harvested area**. This is a measure of productive crops area e.g 100 Ha or 100 Km2

We can treat these metrics using generalised assumptions in the same way gender ratios are done but there are differences in approach. In the first instance gender break down is based on post-processing of the impact with aggregation layer, whereas in this use case we would be post processing the impact layer. We have a problem though that it does not make sense to apply global defaults (equivalent to e.g. national gender ratio) as prices of crops and goods may vary greatly across the country. Is this assertion correct?

## Where do we find appropriate land use data?

* BIG provides land use data for Indonesia - this data can be downloaded free trough : inageoportal
  
http://portal.ina-sdi.or.id/faces/oracle/webcenter/portalapp/pagehierarchy/Page49.jspx?_afrLoop=30041381169853277&_afrWindowMode=0&_afrWindowId=null#%2Foracle%2Fwebcenter%2Fportalapp%2Fpagehierarchy%2FPage35.jspx%40%3F_adf.ctrl-state%3Dnq2gr7910_4
   
The data scale is 1 : 25.000 
These data attribute consist of  name of type of land cover, Fcode, and area (in m2)
The land cover spatial data production is under Center of Integration Thematic and Mapping-BIG. 

[...]

PR #2044, opened 2015-06-16
Land cover impact

wonder-sk:
Implemented as polygon hazard - polygon land cover exposure impact function - as suggested in #1216

Calculates affected and total area per each land cover type in the analysis area.

Only affected area are written to the resulting impact layer. It is possible to write also not affected areas, at the expense of extra processing time necessary for that (the code for that is currently commented out).

The impact function currently does not consider different classes of hazard.

Styling of resulting layer uses red border and no fill (so the fill color of land cover types from original layer can be seen).

Tested with artificial data and land cover of Kabupaten Bojonegoro as linked in #1216. (NOTE: the Bojonegoro shapefile has some duplicate polygon features, so total area reported by the IF is greater than the analysis area).

I am not very familiar with the whole metadata / keywords / parameters sub-system, so that area will need extra focus during the review. There are few TODOs where I would appreciate reviewer's comments.

Bojonegoro land cover + artificial hazard:
image

Artificial data:
image


ismailsunni:
Hi @wonder-sk 
This looks great.
Last thing to do is adding percentage for each land cover type as they requested.

Thank you...

wonder-sk:
@ismailsunni thanks for the review. added the percentages now.

I have realized there is an issue with how InaSAFE determines the analysis area as the intersection of hazard and exposure layers. For area analysis like this one, it makes more sense to use full extent of exposure layer (optionally intersected with user's analysis extent). Otherwise the total areas for individual land covers are misleading. Consider the case with artificial data - the total areas should read: Water = 199.5, Population = 399.1, Forest = 199.5, Meadow = 99.8. The total affected area is then reported to be ~86%, while the true affected area was supposed to be ~33%.

Any ideas how to deal with that? Add a piece of metadata to the impact function stating "ignore hazard extent when calculating analysis extent?

wonder-sk:
Also, what to do with the failing wizard-related unit tests? Should be wizard updated to include land cover analysis or just fix the tests so they don't complain?

I am also a bit lost what's wrong with the last failing test related to metadata - any clue?
http://jenkins.inasafe.org/job/pull-request/619/testReport/junit/workspace.safe.impact_functions.test.test_impact_function_manager/TestImpactFunctionManager/test_hazards_for_layer/

ismailsunni:
Hi @wonder-sk 
For the failing tests, it's alright. We can just update the  expected value in the test. The wizard automatically include land cover.

For the missleading, I think we should ask @timlinux for his opinion.

wonder-sk:
The failing test in Travis is caused by the fact that ubuntugis-unstable is still at QGIS 2.8.1 which contains a bug causing distances and areas to be computed as zero. This has been already fixed in QGIS 2.8.2 (released on May 10).

@timlinux is it OK to merge that to experimental branch anyway?

timlinux:
Hi

+1 to merge to experimental thanks

wonder-sk:
Merged to experimental branch.

PR #2060, opened 2015-06-18
Wizard updates

borysiasty:


ismailsunni:
LGTM...

Only some minor changes.

1. Put the icon in the center. Or at least not so much position change between steps.
center_icon
center_icon_hazard
It would be better if both steps above, we put the icon like the step below:
inasafe_032

Or if you prefer, put all icons in the side. But, I prefer in the center.

2. Remove icon in aggregation step (or change to more suitable icon if we have one for aggregation layer).
aggregation_no_icon

borysiasty:
@ismailsunni - thanks, Travis is checking the fixes. Let all icons be on the right, otherwise I'd need to restrict the combobox text length to the list widget (to have the same "center"). The centered icons was a stupid bug in Designer - I rearranged order of the widgets in the xml file and now it should resize correctly.

Issue #1998, opened 2015-06-03
Align Rasters Tool


wonder-sk:
# Align Rasters Tool

The idea is to add support for alignment of rasters to QGIS. By the alignment of rasters it is meant that several raster layers are resampled to a common coordinate reference system, having the same image size, cell size and cells are perfectly aligned among the individual raster layers. Aligned rasters are required for impact functions in InaSAFE when either hazard and exposure are raster layers.

The tool will consist of two parts:
- API in QGIS analysis library - for use in Processing framework, InaSAFE or other plugins
- module for Processing framework - to allow users to use the tool from graphical interface
## Integration with InaSAFE

InaSAFE code will be updated to use the raster alignment tool if it is available, creating temporary rasters. If an older version of QGIS is used, it will use current code path within InaSAFE: clipping of rasters to the requested extent if the rasters are aligned, terminating with an error otherwise.
## Implementation Details

The tool will be mainly built around GDAL library and its warping algorithms. The implementation will use C API calls for that (i.e. not running `gdalwarp` command line utility) to allow greatest flexibility. All the heavy lifting will be done by GDAL, with the Align Rasters tool serving as a wrapper that processes multiple files at once with the same resampling configuration. There are several reasons for choosing GDAL:
- the warping code is proven with time by broad community
- covers lots of situations (raster data types, raster bands, nodata values etc)
- has optimized code paths for high performance warping
- lots of configuration options
- already a dependency of QGIS

The GDAL library allows a wide range of configuration options to be used for resampling:
- clipping to region of interest
- reprojection between different CRS
- resampling methods
- masking
- handling of nodata values
- pre- and post- warping routines

The choice of resampling method will mainly depend on the type of input raster layers:
- discrete data (e.g. classes) are best resampled with nearest neighbor method (or mode)
- continuous data are typically resampled with interpolation (bilinear, bicubic or others)

GDAL also supports aggregation during resampling - it is possible to use average or mode (most often encountered value) of all contributing pixels. GDAL does not support weighted aggregates - the proportion of the source pixels covering the output pixel is ignored. From GDAL 2.0 there are further aggregation options: max, min, median, first/third quartile.

Handling of raster data where values represent absolute counts within area of the cell need to be treated specially: when the change of cells' areas during resampling it is necessary to rescale also the values. This will be done using pre- and post- warping routines, where the values are first transformed to values independent from cell size (e.g. count per square kilometer), then resampled and finally converted back to values representing counts per cell.
[...]

PR #2070, opened 2015-06-26
Align raster layers using QGIS functionality

wonder-sk:
The PR implements a new way of clipping and resampling of rasters within InaSAFE - see RFC #1998 for more details.

In order to test the new functionality one needs QGIS raster alignment tool - currently only available in my branch:
https://github.com/wonder-sk/QGIS/tree/raster-alignment-tool

If the QGIS raster alignment tool is not present, InaSAFE will use the original code path.

### DIFF
--- a/safe/utilities/clipper.py
+++ b/safe/utilities/clipper.py
@@ -35,12 +35,13 @@ from qgis.core import (
     QgsRasterLayer)
 from PyQt4.QtCore import QProcess
 
-from safe.common.utilities import temp_dir, which, verify
+from safe.common.utilities import temp_dir, which, verify, unique_filename
 from safe.utilities.keyword_io import KeywordIO
 from safe.common.exceptions import (
     InvalidParameterError,
     NoFeaturesInExtentError,
     CallGDALError,
+    AlignRastersError,
     InvalidProjectionError,
     InvalidClipGeometryError)
 from safe.utilities.utilities import read_file_keywords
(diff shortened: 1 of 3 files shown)


wonder-sk:
The existing InaSAFE clipper does not seem to correctly clip at the border of the raster. Compare the clipped population layer when running flooding IF using jakarta 2007 flood, with full analysis extent...

Clipped population with existing InaSAFE clipper:
image

Clipped population with new QgsAlignRaster class:
image

As can be seen on the pictures, the existing clipper does not output the rightmost original pixels. Because of this, there are some differences in the output of impact functions - e.g. with the above mentioned flooding IF the total amount of people flooded > 1 meters is:
- 979K with existing InaSAFE clipper
- 1001K with new QgsAlignRaster

akbargumbira:
Hmm, before doing the clipping we do some manipulation to the requested extent so that it doesn't clip some part of the pixels and fully include it as a pixel. Actually strange that the right most pixels seem to be clipped even with InaSAFE clipper.

wonder-sk:
The origin, cell size and raster size is the same between the existing and new resampler/clipper, so the extent is adjusted correctly. Probably something is going wrong in gdalwarp.

akbargumbira:
Right, I just looked at your code and you do also use the manipulated extent. So yeah, it must be in gdalwarp. LGTM for the logic, but probably Tim will buzz you with the coding standards :)

safe/utilities/clipper.py
```diff
@@ -676,3 +677,107 @@ def adjust_clip_extent(clip_extent, cell_size, layer_extent):
...
+           QgsAlignRaster.Item(exposure_layer.source(), exposure_output)]
+    # TODO: what keywords to consider for rescaling?
+    if exposure_keywords.get('exposure_unit') == 'count':
+        lst[1].rescaleValues = True
+    align.setRasters(lst)
+
+    # Find out which layer has finer grid and use it as the reference.
+    # This will setup destination CRS, cell size and grid origin
+    # TODO: need to respect 'allow_resampling' keyword?
+    if exposure_keywords.get('allow_resampling', 'true').lower() == 'true':
```
akbargumbira: probably would be better to have a more generic method of this (i.e having one parameter to which layer it should be resampled regarding the grid size) and passing the preference for that would be the job of analysis class (and moving the keywords 'allow_resampling' to that class). With that, this function would be also useful for testing native IF's as we need to clip the layer first. To do that now, I made a clipping function for testing purpose which a bit dodgy

akbargumbira: See this for what I meant https://github.com/AIFDR/inasafe/blob/develop/safe/test/utilities.py#L1181

timlinux:
@wonder-sk Actually I always find it strange that QGIS renders half pixel widths like in both your screenies on the left and right edges. It doesn't make any sense to me why it does it since I thought it was standard in raster data definition to work only in whole pixel units.

With regards to small differences in cells being returned, I think that is ok, especially if the new clipper produces something with better fidelity to the intended clip area. We will need to put logic in our tests to have different expected results though based on whether ``QgsAlignRaster`` was used or not....which is probably going to be a little tedious but I can't see another way past it.

wonder-sk:
@timlinux There is nothing wrong with QGIS rendering here: the layers actually have 638 x 649 pixels, so what looks like single pixels is actually composed from many small pixels... no pixels were cut in half or harmed in any other way.

timlinux:
@wonder-sk thanks for not harming the pixels :-)

Can you please update this PR to merge against experimental branch.

wonder-sk:
@akbargumbira Made qgis_align_rasters() more generic in 15115e1

wonder-sk:
Merged to experimental branch.

PR #2072, opened 2015-06-28
Keyword Version

ismailsunni:


safe/definitions.py
```diff
@@ -20,6 +20,9 @@
 # Please group them and sort them alphabetical
 from safe.utilities.i18n import tr
 
+inasafe_keyword_version_key = 'keyword_version'
+inasafe_keyword_version = 1.0
```
timlinux: Maybe its better to use '3.2' so we can match versions to the releases?

safe/gui/tools/test/test_wizard_dialog.py
```diff
@@ -176,7 +176,8 @@ def test_keywords_creation_wizard(self):
                 'flood_vector_hazard_classes',
             'value_map': {'wet': ['YES'], 'dry': ['NO']},
             'source': 'some source',
-            'title': 'some title'
+            'title': 'some title',
+            'keyword_version': 1.0
```
timlinux: This should be DRY - lets define it once and import wherever needed.

timlinux:
Hi @ismailsunni - LGTM except that I think we should call the version 3.2. My idea was that we would have a similar system to QGIS where each successive version provides a forward transform from the old version to the new. Where there is no difference between successive versions the transform will be a NOOP. @mbernasocchi will need to add the implementation for this in the metadata class.

ismailsunni:
Hi @timlinux 
oh, I see.
I have updated the PR according your comments.

Moreover, it will be a little (or much) bit complex to migrate from KeywordIO to metadata class. As we know, many classes use KeywordIO, and use it to get the raw keyword from the layer. But, the metadata class will give full definition for each keyword. Just for your note @mbernasocchi :) 

I will merge this.

Issue #1991, opened 2015-06-02
Update the IF's to get some parameters from Layer's keywords


akbargumbira:
See the task 4 in here https://docs.google.com/spreadsheets/d/1o8xRZ_Sa3gkt0H8sLrXwJg4OyBKKqtMjQG8h7scRkW8/edit#gid=973901724

ismailsunni:
Done in #2127

PR #2080, opened 2015-07-02
Fix 1991

ismailsunni:
PR for fixing issue #1991 
It's working on progress. Don't merge it yet.


ismailsunni:
test this please, jenkins

ismailsunni:
test this please, jenkins

ismailsunni:
I also remove unused parameters in this PR.

timlinux:
Hi @ismailsunni

Thanks for working on this. My main comment is that I would prefer a more semantic approach e.g.

```
self.hazard_keyword('foo')
self.exposure_keyword('foo')
```

It cuts out a lot of parameter passing around, the method names are simple and clear and the code is easy to read. Will it be a big job for you to restructure it a little to match that?

Regards

Tim

ismailsunni:
Hi @timlinux 

Actually, that's my first approach :)
I don't think it's a big job. I will change it to that.

ismailsunni:
test this please, jenkins

safe/impact_functions/base.py
```diff
@@ -70,6 +70,12 @@ def __init__(self):
         self._exposure = None
         # Layer used for aggregating results by area / district
         self._aggregation = None
+        # Keyword for hazard layer
```
timlinux: Can we use the plural form for these (``*_keywords``)?

safe/impact_functions/base.py
```diff
@@ -311,6 +322,63 @@ def aggregation(self, layer):
         # add more robust checks here
         self._aggregation = layer
 
+        # Automatically set the exposure from the hazard layer.
+        self.aggregation_keyword = self.aggregation.keywords
+
+    @property
+    def hazard_keyword(self):
```
timlinux: So this should be plural

safe/impact_functions/base.py
```diff
@@ -311,6 +322,63 @@ def aggregation(self, layer):
...
+        """Property for the hazard keyword to be used for the analysis.
+
+        :returns: A dictionary representing keyword.
+        :rtype: dict
+        """
+        return self._hazard_keyword
+
+    @hazard_keyword.setter
+    def hazard_keyword(self, keyword):
```
timlinux: and both mentions of keyword here

safe/impact_functions/base.py
```diff
@@ -311,6 +322,63 @@ def aggregation(self, layer):
...
+    def aggregation_keyword(self):
+        """Property for the aggregation keyword to be used for the analysis.
+
+        :returns: A dictionary representing keyword.
+        :rtype: dict
+        """
+        return self._aggregation_keyword
+
+    @aggregation_keyword.setter
+    def aggregation_keyword(self, keyword):
```
timlinux: and all between my last comment and this should be plural....

timlinux:
Hi @ismailsunni I still don't think it is quite right. So the properties you made are for the hazard / exposure etc keyword collections, but I was thinking rather we would have a pattern like this:

To get a single keyword:

```python
# use the keyword method
type_field = if.exposure_keyword('type')
```

To get all of the keywords:

```python
# use the keywords property
type_field = if.exposure_keywords
```

ismailsunni:
test this please, jenkins

akbargumbira:
This is  a bit counter-intuitive now. When users tries to do:
```
if = an_IF()
if.hazard = QgsVectorLayer(a_path)
```
and they access it:
```
if.hazard
````
they will get a SAFE layer or QgsWrapper (surprise, user!)

I think I have told @timlinux that we better make a class SafeLayer that generally wrap the layer. It's implemented something like this:
````
class SafeLayer():
     - property: layer (QgsMapLayer or safe.storage.Layer)
     - property: keywords (dict) (this will get populated when user tries to set layer property)
     - and other details for setter methods
````
With that implementation, we can:
- Remove QgsWrapper
- Remove redundant implementations in safe.storage.Layer that are provided by SafeLayer

In the IF base, all the layers (hazard, aggregation, exposure) are in the form of SafeLayer. So, user should do:
````
if = IF()
if.hazard.layer = QgsVectorLayer() or Raster()
````
The nice things about this are:
- In the IF concrete class, all are in the same form (SafeLayer). If you see now in the run method, accessing layer is different depending on the layer type
- Simplicity. With this PR, user can do: if.hazard.aggregation_keywords or if.hazard.aggregation.keywords (too much)
- Accessing keywords just do:
```
if.hazard.keywords # Get the dict
if.hazard.keywords('layer_type') # Get the value of this keywords

ismailsunni:
I agree with @akbargumbira, especially the surprise element.

> Simplicity. With this PR, user can do: if.hazard.aggregation_keywords or if.hazard.aggregation.keywords (too much)

What does this mean? Is it possible for hazard to have aggregation keywords?

akbargumbira:
Sorry I made mistake, I corrected it in the same comment

timlinux:
Note I wasn't particularly interested in ``IF.keywords`` (I agree we should not duplicate if from SafeLayer class. I was more interested in having a nice clear path for getting a keyword e.g.

```IF.hazard.keyword('foo')```

or similar....

**Comment by ismailsunni** [comment]
So, what about this PR? Should I merge it? Or should I create SafeLayer class like @akbargumbira said?

**Comment by timlinux** [comment]
@so I am +1 for implementing ``SafeLayer`` - transcript from our chat (13 July 2015):

```
		10:12
		so, moving the keywords management to the SafeLayer, not in IF base class, and we will get the keywords as IF.hazard.keywords or IF.hazard.keyword('foo')
[10:12] Tim Sutton: IF.hazard  (returns a SafeLayer) 		IF.hazard.keywords()  (calls SafeLayer.keywords() - which returns a dict) 		IF.hazard.keyword('foo;) (calls SafeLayer.keyword('foo') which returns a value
		@IsmailSunni are we on the same page for that?
		@IsmailSunni and for SafeLayer it 'wraps' both read_layer and QgsMapLayer classes
		eventually we can get rid of read_layer and our IF code will be unchanged
[10:13] GitHub: Build for AIFDR/inasafe (a4d632e159) is pending → https://travis-ci.org/AIFDR/inasafe/builds/70376977
[10:13] Tim Sutton: because we still need SafeLayer to provide the accessor / mutator for keywords (which QgsMapLayer does not have)
		and more eventually we could page QgsMapLayer to also have a keywords concept (in QGIS source) and deal with ISO metadata
		and move the whole lot out of InaSAFE into QGIS
[10:14] GitHub: Build for AIFDR/inasafe (a4d632e159) is pending → https://travis-ci.org/AIFDR/inasafe/builds/70376977
[10:14] Tim Sutton: and then finally drop SafeLayer in favour of only using QgsMapLayer
		@IsmailSunni does that meet with your vision (ok maybe the last step is a new idea for you)?
```

timlinux:
@ismailsunni can you get this ready to merge?

ismailsunni:
@timlinux I should close this PR. I have made another PR for it in #2127

ismailsunni:
This also fixes #2089

PR #2116, opened 2015-07-10
allow passing a PACKAGE arg to test_suite_quick

mbernasocchi:
This minimal change allows passing an argument to test_suite_quick so that we can use 
```make test_suite_quick PACKAGE=common```
to run tests only for that package


timlinux:
LGTM thanks @mbernasocchi

PR #2124, opened 2015-07-13
Feature/inasafe 1179 people in buildings

cchristelis:
This PR replaces #1769. (Target branches cannot be changed.)


timlinux:
Hi @christian - I'd like to review the UI for this with you - for now go ahead and merge and then we do the UI tweak as a follow up task ok?

timlinux:
@Gustry take a look if there is any logic that we can use in the new architecture.

Charlotte-Morgan:
@timlinux - if you find a way to add this exposure in version 4.0 architecture; I would be happy for this functionality to be included but disabled; ie available in debug mode. I think this may be a soloution to keeping code not ready for release current and not losing it?
Same with flood evacuation @Gustry - just an idea

timlinux:
@Charlotte-Morgan Yes I think that is a good approach to prevent code being left behind in an old, stale branch. Given the number of things on our plate right now, it might be better to target this for 4.1 though.

Charlotte-Morgan:
@timlinux - sure it was just an option for giving code a home and making it accessible. A different approach to the experimental repo maybe

PR #2127, opened 2015-07-14
Safe layer migration

ismailsunni:


ismailsunni:
test this please, jenkins

safe/common/utilities.py
```diff
@@ -732,6 +732,20 @@ def color_ramp(number_of_colour):
...
+
+    :param attribute_name: The name of the attribute that we want to get its
+        value.
+    :type attribute_name: str
+
+    :param feature: A row of data representing an vector feature.
+    :type feature: dict
+    :return:
+    """
+    return feature.get(attribute_name, None)
```
timlinux: Is there any value in having a wrapper around a 1-line call?

safe/engine/core.py
```diff
@@ -88,17 +88,17 @@ def check_data_integrity(layer_objects):
 
         # Ensure that geo_transform and dimensions is consistent across
         # all *raster* layers
-        if layer.is_raster:
+        if layer.layer.is_raster:
```
timlinux: is this right? Seems a double call to layer...?

safe/engine/core.py
```diff
@@ -88,17 +88,17 @@ def check_data_integrity(layer_objects):
...
             if geo_transform is None:
-                geo_transform = layer.get_geotransform()
+                geo_transform = layer.layer.get_geotransform()
```
timlinux: Same as above - double mention of layer?

safe/impact_functions/core.py
```diff
@@ -143,3 +143,21 @@ def has_no_data(layer_data):
...
+    :param value: The value mapped to a key in value_map.
+    :type value: int, str, float
+
+    :param value_map: A value mapping.
+    :type value_map: dict
+
+    :returns: A key for the value.
+    :rtype: str
+    """
+    for key, values in value_map.iteritems():
```
timlinux: What happens if more than one key stores the same value. In the docstring at least we should mention that the first match is returned...

safe/impact_functions/earthquake/earthquake_building/impact_function.py
```diff
@@ -167,7 +170,14 @@ def run(self):
                 building_value = building_value_density * area
                 contents_value = contents_value_density * area
 
-            usage = get_osm_building_usage(attribute_names, attributes[i])
+            if (structure_class_field in attribute_names and
+                    structure_class_field):
```
timlinux: ```if (structure_class_field  and structure_class_field in attribute_names and):```

would short circuit any attribute checks

safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
```diff
@@ -144,13 +144,10 @@ def run(self):
...
         displacement_rate = self.hardcoded_parameters['displacement_rate']
 
-        # Extract input layers
-        intensity = self.hazard
-        population = self.exposure
-
         # Extract data grids
-        hazard = intensity.get_data()   # Ground Shaking
-        exposure = population.get_data(scaling=True)  # Population Density
+        hazard = self.hazard.layer.get_data()   # Ground Shaking
```
timlinux: IMHO ```get_``` prefixes are superfluous (I appreciate it is not part of your patch, so just a general comment for future...)

safe/storage/safe_layer.py
```diff
@@ -0,0 +1,128 @@
...
+        :param name: A layer's name
+        :type name: Basestring or None
+        """
+        self._layer = layer
+        if isinstance(layer, Layer):
+            self._keywords = layer.keywords
+        elif isinstance(layer, QgsMapLayer):
+            keyword_io = KeywordIO()
+            self._keywords = keyword_io.read_keywords(layer)
+        else:
```
timlinux: Maybe we should raise an exception here? Or do we have a use case for allowing initialisation with no layer?

safe/storage/safe_layer.py
```diff
@@ -0,0 +1,128 @@
...
+        return self._layer
+
+    @layer.setter
+    def layer(self, layer):
+        """Setter for layer property.
+
+        :param layer: The actual layer.
+        :type layer: QgsMapLayer, Layer
+        """
+        self._layer = layer
```
timlinux: I guess this answers my question above - though we should grab the keywords when we set the layer shouldnt we - using same logic as in ctor...

safe/storage/safe_layer.py
```diff
@@ -0,0 +1,128 @@
...
+    @keywords.setter
+    def keywords(self, keywords):
+        """Setter for keywords property.
+
+        :param keywords: The actual layer.
+        :type keywords: QgsMapLayer, Layer
+        """
+        self._keywords = keywords
+
+    def keyword(self, key):
```
timlinux: OK great this was what I was after in our previous discussions. Thanks!

timlinux:
Thanks very much for this @ismailsunni - LGTM when comments above are taken into account. For reference we discussed on hipchat to rename class members to safe_layer so that we have:

```safe_layer.layer.foo()```

File: safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py (hunk 6 above)
Comment (lucernae) [inline safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py]: Some part of the java world :) . hahaha...
This is just for discussion, but what about ```get_``` prefixes in global method, not as an object getter/setter? Is it also pythonic to leave the ```get_``` out?

**Comment by lucernae** [comment]
Thank's @ismailsunni and @timlinux !
Another concern of mine, what about 12 base class properties that are retrieved from the keywords?
e.g. : ```hazard_value_attribute``` or ```hazard_class_mapping```? Are these going to be implemented and used in concrete IF classes?

[Old lucid chart diagram for reference ](https://www.lucidchart.com/documents/edit/02ca4602-780e-4f87-8974-f1b43d2fc67a/0)

**Comment by ismailsunni** [comment]
@lucernae We will, but I think not for now.  There are some other issues that have higher priority (e.g. volcano IF). I am sure it won't be difficult, since we only need to move from the concrete IF classes to the 12 base classes. And add some usage in some IFs that haven't used it.

File: safe/common/utilities.py (hunk 1 above)
Comment (ismailsunni) [inline safe/common/utilities.py]: Perhaps I was going to add some error checking, but I didn't do it. I will remove it.

File: safe/impact_functions/core.py (hunk 4 above)
Comment (ismailsunni) [inline safe/impact_functions/core.py]: Hmm, I thought we need to change the data structure for the value mapping. It should be real value mapped to single class. E.g.
```python
value_map = {
'KRB1' : 'low', 
'Kawasan Rawan Bencana 1' : 'low', 
'KRB2' : 'medium', 
'Kawasan Rawan Bencana 2' : 'medium', 
'KRB3' : 'high', 
'Kawasan Rawan Bencana 3' : 'high}
```

But for now, we will use the current data structure, and add some notes in this function like your suggestion.

PR #2145, opened 2015-07-22
Volcano sprint/2105

### LINKED ISSUE #2105: IF Issue : Volcanic Ash - Aggregation report
# Problem
There is some problem when using Volcanic Ash IF, especially when the result will be aggregated by admin boundary.

The result of aggregation appears on the report when it use to analysis population impacted.
selection_062

But, the result of aggregation not appears on the report when it use to analysis buildings impacted (polygon and point).
selection_061

cc :
@Charlotte-Morgan 
@timlinux 
@ismailsunni

### LINKED ISSUE #2144: Building type postprocessor does not tally totals correctly
# Problem

The building type postprocessor does not sum totals correctly. This results in total affected being 0, where they should have a value. There are no tests to catch this.

# Expected Outcome

1) Fix the adding
2) Add a test

cchristelis:
-> Added building type postprocessor to Volcanic Ash impact function (Generic Impact function) Fixes #2105
-> Building total function corrected. Fixes #2144

### DIFF
--- a/safe/postprocessors/test/test_building_type_postprocessor.py
+++ b/safe/postprocessors/test/test_building_type_postprocessor.py
@@ -76,6 +76,31 @@ class TestBuildingTypePostprocessor(unittest.TestCase):
             '2',
             message)
 
+    def test_total_affected_calculated_correctly(self):
+        """Test to see that the totalling of buildings is done correctly."""
+        # ratios_total < 1 should pass
+        params = {
+            'impact_total': 0,
+            'key_attribute': 'type',
+            u'Building type': True,
+            'target_field': 'safe_ag__4',
+            'impact_attrs': [
+                {'TYPE': 'Government', 'safe_ag__4': 'Zone 1'},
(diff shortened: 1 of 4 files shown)


safe/postprocessors/test/test_building_type_postprocessor.py
```diff
@@ -76,6 +76,31 @@ def test_process_string_values(self):
             '2',
             message)
 
+    def test_total_affected_calculated_correctly(self):
+        """Test to see that the totalling of buildings is done correctly."""
+        # ratios_total < 1 should pass
+        params = {
+            'impact_total': 0,
+            'key_attribute': 'type',
+            u'Building type': True,
```
ismailsunni: Why do we need 'u' (unicode) for Building type?

ismailsunni:
LGTM

safe/postprocessors/test/test_building_type_postprocessor.py
cchristelis: This is test data, which is a copy of real data.

PR #2162, opened 2015-07-28
Master

timlinux:


timlinux:
Thanks @Samweli for your patch!

Samweli:
Ok Tim,

I'm still working on classified polygon landcover impact function,
adding population percentage calculation as you suggested.

I will send pull request when i have finished it.



Samweli

All the Best.

timlinux:
Hi @Samweli 

Just a note, in case it wasnt clear, you should make a clone of that impact function (IF) and call the new IF something like polygon_flood_polygon_people.

You should actually chat a bit with @lucernae about this because ultimately the polygon people based IF should become an abstract base class with the following subclasses under it:

* polygon flood on polygon pop (using geometry intersection similar to landcover)
* polygon volcano on polygon pop (using geometry intersection similar to landcover)


I dont think we have use cases for the following:

* raster volcano on polygon pop
* tsunami on polygon pop (unless we want to have different thresholds - @Charlotte-Morgan ?)
* polygon earthquake on polygon pop (we dont really encoutner polygon eq data)


Ones I am on the fence about
* raster earthquake on polygon pop (perhaps we could class the eq data into mmi levels and express number of ppl per mmi level but we would not be able to do damage curve style approach of current eq algs)
* raster flood on polygon pop (as above, class flood cells into h/m/l and then count people in each class via intersection)


@samnawi for now to just have a simple implementation based on a clone of the landcover IF would be good, we can follow up with ABCs etc after.

# cc

@ismailsunni @lucernae @Charlotte-Morgan

PR #2169, opened 2015-08-10
Added new Impact Function for Vector hazard and Vector exposure

Samweli:
Created new impact function calculating area affected and number of people affected in the area.
I have used the floods.shp and land.shp in inasafe/safe/test folder as test data.

question_report
reportonly


safe/impact_functions/generic/classified_polygon_people/metadata_definitions.py
```diff
@@ -0,0 +1,104 @@
...
+                    'continuous_hazard_units': [],
+                    'vector_hazard_classifications': [
+                        generic_vector_hazard_classes],
+                    'raster_hazard_classifications': [],
+                    'additional_keywords': []
+                },
+                'exposure': {
+                    'layer_mode': layer_mode_classified,
+                    'layer_geometries': [layer_geometry_polygon],
+                    'exposure_types': [exposure_area],
```
timlinux: We should be referencing the people column in the attribute table here shouldn't we?

safe/impact_functions/generic/classified_polygon_people/parameter_definitions.py
```diff
@@ -0,0 +1,61 @@
...
+    :rtype list[StringParameter]
+    """
+    field = StringParameter()
+    field.name = 'Area Id'
+    field.is_required = True
+    field.value = 'id'
+    return field
+
+
+def area_population():
```
timlinux: This should be taken from keywords in the exposure layer - please chat to @ismailsunni about the procedure for doing that.

safe/impact_functions/generic/classified_polygon_people/parameter_definitions.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+__author__ = 'Samweli Twesa Mwakisambwe "Samweli" <smwltwesa6@gmail.com>'
+__date__ = '8/5/15'
+
+
+from safe_extras.parameters.string_parameter import StringParameter
+
+
+def area_id():
```
timlinux: This should also be expressed in keywords - chat to @ismailsunni

safe/impact_functions/generic/classified_polygon_people/parameter_definitions.py
```diff
@@ -0,0 +1,61 @@
...
+    :rtype list[StringParameter]
+    """
+    field = StringParameter()
+    field.name = 'Population'
+    field.is_required = True
+    field.value = 'population'
+    return field
+
+
+def area_type_field():
```
timlinux: I think we don't need this for pop polys

safe/impact_functions/generic/classified_polygon_people/test/test_classified_polygon_people.py
```diff
@@ -0,0 +1,99 @@
...
+        function.run()
+        impact = function.impact
+
+        impact = safe_to_qgis_layer(impact)
+
+        # self.assertEqual(impact.dataProvider().featureCount(), 3)
+        features = {}
+        for f in impact.getFeatures():
+            features[f['id']] = f.geometry().area() * 1e8
+        expected_features = {
```
timlinux: I guess we need to characterise this with real outputs

timlinux:
Hi

Thanks @Samweli for this patch!

I made some inline notes and will add a couple of extra here:

* the exposure layer properties should be keyword driven, not parameter driven
* lets create a separate layer instead of ```land.shp``` (which will have landuse related keywords) for the population polygons - ```cencus.shp``` might be  a nice choice
* for test datasets, lets avoid using spaces (``Flood Map.shp``)  and try to follow naming conventions from other data layers to embrace the principle of least surprise.
* it would be good to come up with a symbology scheme for the impact layer such that most affected areas (perhaps as a ratio of people : area) are highlighted red, lowest are in orange, unaffected are green. (Might need more colour breaks than that).
* For columns in the the report, it would be good to humanise numbers e.g. areas - humans cannot easily interpret long decimal place numbers. A single digit of precision would probably be good (and a note below the summary indicating that numbers have been rounded).

It will be great to have both @lucernae and @ismailsunni review this.

Regards

Tim

safe/impact_functions/generic/classified_polygon_people/parameter_definitions.py
ismailsunni: The metadata system is only exist in develop branch, we can do it when we merge to develop.

ismailsunni: Or perhaps we should keep the experimental branch up to date with develop branch?

lucernae:
Hi @Samweli ,
I am testing this from your branch. I will add several notes again besides @timlinux 's comments
- Is this only for single event? If so, perhaps you have to change the keywords in the test layer so it has single event keywords in it :D. 
- The IF options needs to be moved to keywords. Particularly if it is trying to select column (I infer three of them are trying to select columns: id, type, population). The IF wizard will also need to include these process (column select). @ismailsunni knows better about wizard related. After that, the IF can receive this information from layer keywords, not from IF options. IF options typically needed for users to input the function related properties, such as constants, or values used inside the function.

timlinux:
Note there is still one failing test:

```
======================================================================
FAIL: Test the IFCW mode.
----------------------------------------------------------------------
  File "/home/travis/build/AIFDR/inasafe/safe/gui/tools/test/test_wizard_dialog.py", line 1482, in test_input_function_centric_wizard
    message)
AssertionError: Invalid functions count in the IF matrix 2! For  raster and polygon there should be 2 while there were: 0
```

timlinux:
Closing this since it is merged to develop already.

Samweli:
yes

PR #2172, opened 2015-08-13
Update travis.yml for qgis 2.8.3

ismailsunni:


timlinux:
Lgtm thanks

ismailsunni:
@timlinux actually, it gives error. Travis uses ubuntu 12.04. But, when I try to install QGIS 2.8.3 in 12.04, I got error after adding the QGIS LTR repository.

PR #2175, opened 2015-08-14
revert zero fatality assumption in the ITB fatality model

dynaryu:


lucernae:
Close this since we don't need it anymore

CC:
@dynaryu @ismailsunni

PR #2176, opened 2015-08-14
Add .scrutinizer.yml and update travis.yml

ismailsunni:


.scrutinizer.yml
```diff
@@ -0,0 +1,103 @@
...
+        typecheck_too_many_function_args: true
+        typecheck_unexpected_keyword_arg: true
+        variables_global_at_module_level: true
+        variables_global_variable_not_assigned: true
+
+tools:
+    pylint:
+        python_version: '2'
+        config_file: pylintrc
+    external_code_coverage:
```
lucernae: CMIIW @ismailsunni :

I don't see .coveragerc files in InaSAFE, so I assume we don't have any external coverage service (like coveralls.io). So, we can either uses coverage service, or perhaps delete external_code_coverage node altogether since it causes errors.

PR #2178, opened 2015-08-14
changed the test code to reflect the change in the impact function

dynaryu:
I have a question regarding population rounding function in the core.py. What's the logic behind the function?


ismailsunni:
hi @dynaryu have you finished with this PR? If so, I will merge it so that the travis will be green again. For the question, we should ask @cchristelis or @Charlotte-Morgan. IIRC, they have some discussion about it.

dynaryu:
@ismailsunni Yes. it's finished. You can go ahead with the latest commit.
Thanks.

PR #2184, opened 2015-08-17
Fix bug, which breaks on vector floods with float flood values.

### LINKED ISSUE #2142: Minimum needs profile manager fails when adding a new item
```
  File "/Users/timlinux/.qgis2/python/plugins/inasafe/safe/gui/tools/minimum_needs/needs_manager_dialog.py", line 574, in save_resource
    self.add_resource(resource)
  File "/Users/timlinux/.qgis2/python/plugins/inasafe/safe/gui/tools/minimum_needs/needs_manager_dialog.py", line 243, in add_resource
    item.setText(updated_sentence)
RuntimeError: wrapped C/C++ object of type QListWidgetItem has been deleted

```

Options I used are as per notes below:

Press the new button to create a new profile. A dialog will appear asking for the name of your profile. Call it ‘training’ and press ‘save’.

The first thing to do is to enter the ‘provenance’ for the profile. Typically you should enter a short statement explaining which legislation or guideline was used to define this needs profile. In our case we will just use a sentence like ‘Needs as defined for InaSAFE Training Course’ to make it clear that this is just a training profile.

Now click the add button and create a new need item with the following properties.

name, description, units (in singular, plural and abbreviated forms), specify maxima and minima for the quantity of item allowed, a default and a frequency.

screen shot 2015-07-22 at 08 37 31

Now press the save resource button

cchristelis:
Fixes #2142


safe/impact_functions/inundation/flood_vector_building_impact/impact_function.py
```diff
@@ -65,7 +65,9 @@ def notes(self):
                     'Buildings are said to be inundated when in a region with '
                     'field "%s" in "%s" .') % (
                         self.hazard_class_attribute,
-                        ', '.join(self.hazard_class_mapping[self.wet]))
+                        ', '.join([
+                            str(_) for _ in self.hazard_class_mapping[self.wet]
```
timlinux: is this unicode friendly? Also since the var is actually used it might be nice to call the ```_``` var ```hazard_class``` or something similar to make it more readable?

timlinux:
LGTM when above comments are taken into account - thanks!

Issue #2171, opened 2015-08-11
We should disable organisation logos in the dock by default


timlinux:
This was actually agreed a long time ago - lets hide the logos by default. While I was doing training, many users were hampered by low resolution screens and the logos takes so much of the dock they could not see important information presented there.

PR #2191, opened 2015-08-20
Disable dock organisation logo for default installations - see #2171

timlinux:


timlinux:
I'm self reviewing this as the change is trivial

PR #2208, opened 2015-08-21
Realtime fix:

lucernae:
- Fix errors in realtime where hazard name is read as int instead of str
- Fix some errors in realtime tests


timlinux:
Thanks @lucernae LGTM

Issue #2221, opened 2015-08-24
When an exception occurs, we should be sure to revert the wait cursor


timlinux:
At the moment if something goes on we often get left with a  busy cursor when in fact the operation is completed (albeit with an error).

# Proposed solution

Add a flag as property to dock to keep track of the cursor state, and roll it back if needed.

timlinux:
See also http://www.programcreek.com/python/example/55362/PyQt4.QtGui.QCursor - particularly example 4 which uses a different approach for cursor management: 

```
def setActive(self, active):
        self.active = active
        if self.active:
            self.setCursor(QtGui.QCursor(QtCore.Qt.PointingHandCursor))
        else:
            self.setCursor(QtGui.QCursor())

```

PR #2223, opened 2015-08-24
Try to restore override cursor - see #2221

timlinux:


timlinux:
Merging - only failing test is relating to overpass being down.

PR #2226, opened 2015-08-24
Added new templates from Iwan - see #1314

### LINKED ISSUE #1314: A3 Map template
Create A3 map template in QGIS 2.6

timlinux:
fix #1314 - need A3 templates


timlinux:
Test fails due to upstream issue with overpass. I am going to go ahead and merge. @ismailsunni can you disable that test for now so that we can see when other tests failures cause a break?

PR #2227, opened 2015-08-25
Fix #2156

### LINKED ISSUE #2156: IF Issues : Volcanic Ash - Point Structure
# Problem
Thank for @cchristelis as your fix the issue #2105.
But, there is a new problem when I try to run the volcanic ash impacted to centroid building (Point structure). The IF is shown the error when I try to run this function.
image

In the past, when I run this IF, the process is running until show the report. I try this by changing the Inasafe develop version to SHA 10c28e7 (a day before I post the issue #2015), and this function is run.
selection_088

# Detail
I run use Linux Ubuntu 14.04 and Windows 7, also use the QGIS 2.8.2 and QGIS 2.10, the result is same.

# Proposed Solution
Make it happy

# cc
@Charlotte-Morgan 
@timlinux 
@ismailsunni

ismailsunni:
Fix #2156

### DIFF
--- a/safe/gis/qgis_vector_tools.py
+++ b/safe/gis/qgis_vector_tools.py
@@ -121,8 +121,8 @@ def create_layer(vector):
     """
     crs = vector.crs().toWkt()
     if vector.geometryType() == 0:
-        msg = "Points cant' be split"
-        raise WrongDataTypeException(msg)
+        # We can create layer from Point. Do not need to split it.
+        uri = 'Point?crs=' + crs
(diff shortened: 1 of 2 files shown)


lucernae:
LGTM

On a side note, we don't have point layer as hazard right? So that should be fine.

The test stills fails though, perhaps some problem with data location?

timlinux:
@lucernae hazard layer could be volcano point layers ....

timlinux:
LGTM @ismailsunni

PR #2232, opened 2015-08-25
Roads report mixin

cchristelis:
The last of the mixins.


cchristelis:
@timlinux, could you have a look at this RP when you have a chance please.

timlinux:
Thanks @cchristelis LGTM

One comment (for future revision) is that I think it might be good to investigate using our messaging API for reporting. There may be some advantages like

* uniformity of core
* ability to render to different formats (text, html, json) easily
* more readable code

PR #2234, opened 2015-08-26
Fix 2153 flood on road

ismailsunni:


lucernae:
LGTM, tested using Flood polygon + roads in test layers in Qgis Pisa (2.10)
Things that concerns me is we have to manually initialize the plugin and it is less futureproof because we are hacking the library (even reimplement a function which supposed to be handled by the plugin).

Btw, just to make sure, did you enable processing plugin in your Qgis via plugin management menu before running the IF? cc @ismailsunni 
It could be because it's not activated in your qgis? perhaps.

ismailsunni:
@lucernae Yeah, you are right. After I enable the processing plugin, it runs correctly.

So, we don't really need this PR if we have enabled the processing plugin. Next question, should I apply this PR or just simply give error message that the user needs to enable the processing plugin in QGIS?

I prefer to apply this PR, so that the user doesn't need to do anything.

cc @timlinux

safe/impact_statistics/aggregator.py
```diff
@@ -139,7 +139,9 @@ def __init__(self, extent, aggregation_layer):
         # aggregation polygons (one list for one polygon)
         self.impact_layer_attributes = []
 
-        self.processing = processing
+        # Notes(Ismail): Need to initialize Processing in QGIS 2.8.x
```
timlinux: For supporting older versions it may be good to put this in a QGIS version switch (I dont know if initialised call is available in < 2.8)

timlinux:
@lucernae and @ismailsunni good diagnostic work guys! I think that this approach is good because we want inasafe to work even if for some reason the user has disabled processing on their sytem, so I would +1 to go ahead with this patch. And shame on you for disabling your processing plugins @ismailsunni its one of the nicest plugins in QGIS hahahaa :-)

ismailsunni:
@timlinux I don't ever remember why I disabled it. That's why I jump to the code, not the plugin manager :dancer:

PR #2248, opened 2015-08-26
fix #2245 UI: new look is missing bold

### LINKED ISSUE #2245: UI: new look is missing bold
# problem
The new look is great but we have lost the bold in the layer key words and report headings
eg 
 - [ ] Layer keywords are not as easy to read without bold
image
 # solution 
use bold for the key word categories (not the user entered values) ie Layer_purpose is in bold but hazard is not

- [ ] report headings in the dock are missing bold
image

but show in bold in web interface
image

# proposed solution
use bold in the dock too

timlinux:


Charlotte-Morgan:
LGTM

Issue #2217, opened 2015-08-24
We should hide single / multi event choices in the wizard


timlinux:
For now we should not ask the user until we have IF's that are dependent on one or the other. @charlotte-morgan to confirm this.

Charlotte-Morgan:
see #1961

adissadis:
+1 i agree with that. Its really confusing to us when we tried to use IF wizard

ismailsunni:
Hi @Charlotte-Morgan by removing the radio button, I will also set all IF to have multiple_events and single_event.

Charlotte-Morgan:
hi @ismailsunni - i'm not really sure what you mean - except i think the IF is now reading this from the key words so its not needed in the IFCW

ismailsunni:
I have set all IF to have single and multiple event. So, we only need to remove the radio button now. Under the hood, the radio buttons do not have any effects.

PR #2251, opened 2015-08-26
Set all hazard categories

ismailsunni:
This PR will set all IF to have multiple event and single event. It will needed if we addressing #2217


timlinux:
LGTM thanks @ismailsunni

Issue #2262, opened 2015-08-26
Impact Layer Merge with Unicode Aggregation Layer


lucernae:
Related with #2229 #2233 
It occured when we uses aggregation layer which contains unicode area.
The area name from aggregation layer somehow different than the keys in ```self.summary_report```
See this on generate_report method when trying to use aggregation layer.

This could be a big bug. The problem could be either on the report generation or when reading aggregation layer from source.

Issue #2233, opened 2015-08-26
Impact Layer Merge Tool bug and spec


lucernae:
See issue #2229.

While investigating it. I found another issue and perhaps some part of the code which isn't clear for me. Can someone clarify this? @ismailsunni @timlinux 

- [x] ```self.entire_area_mode``` is not reset back to False if we change aggregation layer from Entire area to an aggregation layer without closing the dialog. This is probably just a bug, I just have to add some code to reset it back.
- [x] If the first and second layer were results from un-aggregated analysis. Do we still able to merge it using aggregation layers?
- [x] If the first and second layer were results from aggregated analysis. Do we still able to merge it using 'Entire Area' aggregation?
- [x] If one of the layer is aggregated and the other doesn't. What should we do? :D
- [x] I found that in ```self.merge``` method, we get the dictionary of aggregated report from the dom in the impact report. Is this safe? What if the report DOM structure was changed? Is there any alternatives?
- [x] Related with above issue. Because we get it from the DOM, some of the keys possibly not match with aggregation layer if it contains Unicode. This is perhaps an issue with the postprocessor when generating the report.
- [ ] Generated PDF should display the map with relevant aggregation boundaries.

timlinux:
>  self.entire_area_mode is not reset back to False if we change aggregation layer from Entire area to an aggregation layer without closing the dialog. This is probably just a bug, I just have to add some code to reset it back.

Yes I think this is a bug

> If the first and second layer were results from un-aggregated analysis. Do we still able to merge it using aggregation layers?

No - we need them to be aggregated to the same aggregation layer.

> If the first and second layer were results from aggregated analysis. Do we still able to merge it using 'Entire Area' aggregation?

If both were created using entire area, I think we should be able to merge them

> If one of the layer is aggregated and the other doesn't. What should we do? :D

Warn the user nicely that it is not supported please :-) 

> I found that in self.merge method, we get the dictionary of aggregated report from the dom in the impact report. Is this safe? What if the report DOM structure was changed? Is there any alternatives?

Impact merge was very hacky - the idea was to rewrite it once we have proper IF metadata in e.g. json format so that we can parse the outputs in a more rigourous way - something for 3.3 maybe


> Related with above issue. Because we get it from the DOM, some of the keys possibly not match with aggregation layer if it contains Unicode. This is perhaps an issue with the postprocessor when generating the report.

Hmm ... perhaps forcing the keys to ascii here might work?

Issue #2229, opened 2015-08-25
Impact Layer Merge Tool does not work with Aggregation [bug]


harrymahar:
# Problem
Impact Merge Layer Tool does not work using aggregation

# Detail
I want to merge the results of the Jakarta flood impact function both for the building and population using Impact Layer Merge tool in InaSAFE. Unfortunately, It was only can managed to merge the impact function without aggregation. When I want to merge those impact using aggregation function by using the district boundaries of Jakarta, the results box said it created successfuly. However, it does not created in the selected folder.

# Screenshot
![impact merge layer_agregasi_error](https://cloud.githubusercontent.com/assets/2534179/9462937/63131374-4b43-11e5-8b4f-ca94ae5f4271.PNG)

![impact merge layer_agregasi_error2](https://cloud.githubusercontent.com/assets/2534179/9462945/71000988-4b43-11e5-8372-fc9a43e5f3f5.PNG)

![impact merge layer_agregasi_error3](https://cloud.githubusercontent.com/assets/2534179/9462961/8b181b62-4b43-11e5-9c31-1bd4dd3c6ec3.PNG)

# Platform
Windows 8.1 64 Bit
RAM 4 GB
QGIS 2.8.3 Wien
InaSAFE 3.2.0 dev-master

# Data used
Hazard : Jakarta_Flood_18113__WGS84.shp (Raster)

Exposure : 
*Java_Population_WGS84.tiff (Raster)
*Jakarta_Building_WGS84.shp (Vector)

Agregation : Jakarta_District_Boundary.shp

timlinux:
Thanks @harrymahar @lucernae can you look into this please?

lucernae:
@timlinux it seems the merge only happens when the first and second layer's are already aggregated.
So, it had no effect when the first and second layer is processed using 'entire area' aggregation but it is trying to be merged using another aggregation layer.

Perhaps we should offer relevant aggregation layer based on first and second layer in order to make the user less confused?

harrymahar:
Hai @timlinux and @lucernae , Today I try to merge the result again. I start again using same data using Flood Impact Function in DKI Jakarta. The different thing that I did when I merge the impact function results was I match the extent of the results. So, they had similar extent view and positition in QGIS. Finally, when I merge those data, The PDF results was created based on aggregation layer which is Jakarta_District_Boundary. Eventhough it succeed to create the result, All PDF has same area display despite the result has divided based on district boundary and has different number of result. 

[...]

PR #2263, opened 2015-08-26
bugfix for issues #2229 and #2233.

lucernae:
What is being addressed #2229 and #2233:

See #2233 for reference

- Stuck in entire area mode if choosing aggregation layer without closing the dialog
- User should be able to merge in entire area mode, even if one of the layer is aggregated, both or none
- Hacky fix for naming bug in reports where it can't match unicode area names. Should be fine for ASCII
- Additional checks to make sure both layers are aggregated with the same aggregation layers with what the user choose.
- Additional checks to make sure that the user uses aggregated postprocessor in both layers if user choose aggregation layer

### DIFF
--- a/safe/gui/tools/impact_merge_dialog.py
+++ b/safe/gui/tools/impact_merge_dialog.py
@@ -556,28 +556,6 @@ class ImpactMergeDialog(QDialog, FORM_CLASS):
                         'Keyword aggregation attribute not found for '
                         'aggregation layer.'))
 
-    def normalize_dict_keys(self, input_dict):
-        """Normalizing to force keys to ascii.
-
-        Possible bugs: This function can't handle if we had two or more keys
-        with the same normalized form. It will collide. The next keys will
-        override previous keys.
-
-        Example, if somehow we had two keys 'District a' and 'District b',
-        and after forced to become ascii, it both became 'District '
-        (because a and b in unicode), we had a collision.
-
-        :param input_dict: input dictionary with unicode keys
-        :type input_dict: dict
-        :return: normalized dict
-        :rtype: dict
-        """
-        output = OrderedDict()
-        for k, val in input_dict.iteritems():
-            normalized_k = k.encode('ascii', 'ignore')
-            output[normalized_k] = val
-        return output
-
     def merge(self):
         """Merge the postprocessing_report from each impact."""
         # Ensure there is always only a single root element or minidom moans
@@ -663,13 +641,6 @@ class ImpactMergeDialog(QDialog, FORM_CLASS):
                 area = f[aggregation_attr_index]
                 if area not in aggregation_keys:
                     aggregation_keys.append(area)
-
-            # normalized aggregation_keys. hacky fix for unicode issues
-            # #2233 #2229
-            report_keys = [
-                k.encode('ascii', 'ignore') for k in report_keys]
-            aggregation_keys = [
-                k.encode('ascii', 'ignore') for k in aggregation_keys]
 
             is_subset = True
             for k in report_keys:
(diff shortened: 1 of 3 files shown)


lucernae:
Don't merge it yet.
Also trying to fix #2262

safe/utilities/clipper.py
```diff
@@ -246,7 +246,7 @@ def _clip_vector_layer(
 
     writer = QgsVectorFileWriter(
         file_name,
-        'UTF-8',
+        None,
```
lucernae: Please look at this line.

I don't understand why, if we uses 'UTF-8', the new test failed because the unicode character doesn't match.

But if we change it to None, or even ascii, the test succeed.

Well, any feedback here?

safe/utilities/test/test_clipper.py
```diff
@@ -666,6 +668,40 @@ def test_adjust_clip_extent(self):
             expected_adjusted_extent, adjusted_extent)
         self.assertEqual(adjusted_extent, expected_adjusted_extent, message)
 
+    def test_clip_vector_with_unicode(self):
+        """Test clipping vector layer with unicode attribute in feature.
+
+        This issue is described at Github #2262 and #2233
+        """
```
lucernae: I added a new test here.

This is to test that the clipped layer should contains possible values from original layer, to make sure it is consitent. Originally this test fails when the attribute compared contains unicode char, which in turn make Impact Layer Merge fails.

But the test succeeded if we write the attribute using QgsVectorFileWriter by specifying None in the encoding.
Should we just put None for the encoding?

lucernae:
@ismailsunni @timlinux could you review this?

ismailsunni:
LGTM. 
For the encoding, perhaps @timlinux knows better than me. I have no idea also.

safe/gui/tools/impact_merge_dialog.py
```diff
@@ -546,6 +556,28 @@ def validate_all_layers(self):
...
                         'aggregation layer.'))
 
+    def normalize_dict_keys(self, input_dict):
+        """Normalizing to force keys to ascii.
+
+        Possible bugs: This function can't handle if we had two or more keys
+        with the same normalized form. It will collide. The next keys will
+        override previous keys.
+
+        Example, if somehow we had two keys 'District a' and 'District b',
```
timlinux: Isn't a better example if we have e.g. 'District' and 'Dîstrict' ...

safe/gui/tools/impact_merge_dialog.py
```diff
@@ -546,6 +556,28 @@ def validate_all_layers(self):
                         'Keyword aggregation attribute not found for '
                         'aggregation layer.'))
 
+    def normalize_dict_keys(self, input_dict):
```
timlinux: Need ``.. versionadded:: 3.2`` in docstrings for new methods and e.g. ``.. version changed:: 3.2 added parameter foo``  for methods whose api has changed

safe/utilities/test/test_clipper.py
timlinux: Yes for now use None and add a TODO: This is a lousy hack, see above ticket, then add your above annotations to the ticket please.

timlinux:
LGTM - lets solve this on a best effort basis - I have some ideas about a different approach to doing IF merging but it is something to discuss in our workshop in JK - I don't know how important this feature is for @Charlotte-Morgan and how much time we want to invest in it...

So I can remind myself later - my idea is basically to do an architectural change so that during a model run we use one or more Exposure datasets and generate the merged report right at the point of analysis.

safe/gui/tools/impact_merge_dialog.py
lucernae: No, that way it becomes District and Dstrict. I want to use french accent for a and b, but I'm just too lazy to find the character map :p

Issue #2258, opened 2015-08-26
[Bug] Buildings count discrepancy between impact report and detailed building type report


cchristelis:
# Problem 

As per @adissadis's comment https://github.com/AIFDR/inasafe/issues/2235#issuecomment-134889994:
"""
This problem [count discrepancy] also happen in Flood Raster with Building. I think detailed report produced more bigger result than the general report, which i think its really strange.

![ac77a6fe-4c03-11e5-8bd1-0635c640ffca](https://cloud.githubusercontent.com/assets/2319504/9496123/23067a42-4c0f-11e5-88bd-c1b31eab1d8f.png)
![b6c3f7f2-4c03-11e5-910c-832d6a236372](https://cloud.githubusercontent.com/assets/2319504/9496126/24b1ac40-4c0f-11e5-9503-4c2ccecc2fbf.png)

I give example in sport facility where in the general report its only 6 building but in detailed report it become 50 buildings
"""

# Proposed solution

Investigate the discrepancy.

PR #2264, opened 2015-08-26
Bugfix/2258

cchristelis:
Since we iterate over buildings (features) we should only count them once if they have a non-zero class.

fixes: #2258


timlinux:
LGTM

PR #2274, opened 2015-08-28
fix inasafe-2268 - use nicer table formatting for keywords in dock

### LINKED ISSUE #2268: We need better presentation for the keywords in dock
Currently our keywords look like this:

screenshot from 2015-08-27 13-33-41

There are a number of problems here:

* The text is not nicely laid out
* The titles have underscores and odd casing
* We could do a better job of ordering the information into a logical sequence

I am proposing to create a table rather and clean up the keyword names. @adissadis I don't want to mess with your work, so I would like to know if you would rather that I wait until 3.3 to make this change?

timlinux:
fix #2268 - use nicer table formatting for keywords in dock


cchristelis:
LGTM

Issue #2288, opened 2015-08-30
Exception running flood on buildings


timlinux:
```
Details

These additional details were reported when the problem occurred.

An exception occurred when running the impact analysis.
in method 'DataSource_CreateLayer', argument 4 of type 'OGRwkbGeometryType'
Diagnostics

In file "/Users/timlinux/.qgis2/python/plugins/inasafe/safe/utilities/analysis.py", line 969, in run_analysis self.impact_layer = calculate_impact(self.impact_function)
In file "/Users/timlinux/.qgis2/python/plugins/inasafe/safe/engine/core.py", line 254, in calculate_impact result_layer.write_to_file(output_filename)
In file "/Users/timlinux/.qgis2/python/plugins/inasafe/safe/storage/vector.py", line 681, in write_to_file self.geometry_type)
In file "/Library/Frameworks/GDAL.framework/Versions/1.11/Python/2.7/site-packages/osgeo/ogr.py", line 604, in CreateLayer return _ogr.DataSource_CreateLayer(self, *args, **kwargs)
```

Scenario:

```
[tandale_flood_14_singlepart]
exposure = ../../Users/timlinux/gisdata/DarEsSalaam/CaseStudies/roads.shp
hazard = ../../Users/timlinux/gisdata/DarEsSalaam/CaseStudies/tandale-modelled-floods/tandale_flood_14_singlepart.shp
function = FloodVectorRoadsExperimentalFunction
extent = 39.231364, -6.800884, 39.251790, -6.792065
extent_crs = EPSG:21036
```

PR #2294, opened 2015-08-31
Gracefully handle cases where zero impact occurs for flood polygons on buildings.

### LINKED ISSUE #2221: When an exception occurs, we should be sure to revert the wait cursor

timlinux:
See #2288 
This PR also fixes #2221


timlinux:
@ismailsunni please review

ismailsunni:
LGTM, thanks

PR #2308, opened 2015-09-02
[IFCW] Fix #2217

### LINKED ISSUE #2217: We should hide single / multi event choices in the wizard

borysiasty:
Remove the Single/Multiple Scenario radiobuttons

### DIFF
--- a/safe/gui/tools/wizard_dialog.py
+++ b/safe/gui/tools/wizard_dialog.py
@@ -66,7 +66,6 @@ from safe.definitions import (
     layer_purpose_exposure,
     layer_purpose_aggregation,
     hazard_category_single_event,
-    hazard_category_multiple_event,
     layer_geometry_point,
     layer_geometry_line,
     layer_geometry_polygon,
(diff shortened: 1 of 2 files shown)


safe/gui/tools/wizard_dialog.py
```diff
@@ -1980,34 +1979,13 @@ def on_tblFunctions1_cellDoubleClicked(self, row, column):
...
-            self.populate_function_table_1()
 
     def populate_function_table_1(self):
         """Populate the tblFunctions1 table with available functions."""
-        if self.rbHazSingle.isChecked():
-            hazard_category = hazard_category_single_event
-        else:
-            hazard_category = hazard_category_multiple_event
+        # The hazard category radio buttons are now removed -
+        # make this parameter of IFM.available_hazards() optional
```
borysiasty: @ismailsunni Could you please look if we can just remove this parameter?

ismailsunni: As we discussed in gitter, I will leave it just like this for now. I have updated all IF to support single and multiple events. So, no need to update the IFM (for now). We can do it later if needed.

Btw, LGTM.

Issue #2075, opened 2015-06-30
Opening function centric wizard makes QGIS crash on exit


wonder-sk:
If I just open function wizard and then close it, QGIS will crash on exit. Without opening the wizard QGIS terminates correctly. Tested with 2.8 and 2.10 on Linux 64bit.

From my investigation:
- crash happens when InaSAFE tries to handle layersWillBeRemoved() signal on exit - at this point the plugin should have been unloaded
- the problem is that WizardDialog is created with main window as a parent, therefore it is not deleted automatically when the plugin is unloaded (and it keeps reference to the dock, which is not deleted either - and that's why its slots will be called)

One of the solutions would be in unload() method in safe/plugin.py add the following code to explicitly delete the wizard:

```
for wizard in self.iface.mainWindow().findChildren(QWidget, "WizardDialogBase"):
    wizard.deleteLater()
```

Other way to delete it would be to call `wizard.setParent(None)`

Also, instead of searching for the children of the main window, it is possible to store the created wizard in a member variable, e.g. `self.wizard` - this is safer than searching for a particular object name of within main window.

borysiasty:
Thanks @wonder-sk for this detailed investigation! Unfortunately it seems it only helps if you just open the Wizard without going to the Analysis tab. Once you launch the analysis on the last tab, the crash still occurs - probably because of signals connected to the MessageViewer. Well, now QGIS master crashes anyway (regardless of InaSAFE), so I have to wait with further tests.

borysiasty:
@ismailsunni: Seems the MessageViewer embedded in the wizard doesn't play well. Besides damaging the window resizing policy (if not inside a QScrollArea), there is the problem above - I've just got lost in the tangled signals in AnalysisHandler interfering with the dock. Seems rewriting the AnalysisHandler to handle independently (and analogically) the Dock and the Wizard will be the must sooner or later.

Gustry:
Let's close this ticket now inasafe v4.

PR #2317, opened 2015-09-03
[Wizard] Better prevention from creating multiple wizard sessions. Fi…

borysiasty:
…x some simple cases of #2075


ismailsunni:
LGTM

Issue #2313, opened 2015-09-03
UI: We need consistent user facing terminology


Charlotte-Morgan:
We have made some massive steps in developing terminology to help users understand their data and the impact functions but we still have a bit of 'geek speak' that may not make sense to users.

I propose we tweak the words so that user facing terminology is consistent and may not be the same words used by developers in the code. This should have no impact on code or how things work.

I propose that we match the question asked in the KW wizard with the keywords report in the dock
that we match the metadata and training materials too :)

Below is a sample of the current wording and the proposed wording for 3 examples. there are also minor changes to order of items in the dock report.

Why - I think it will make the link between steps in the KW Wizard and the key words more intuitive and make it easier to write training materials

![image](https://cloud.githubusercontent.com/assets/5743135/9649523/465ff29a-5220-11e5-9811-5f48f9668f13.png)

# cc
@samnawi @fredychandra @iyan31 @adissadis

ismailsunni:
Hi, we can do it by simply changing the name value of each item in definitions.py

Charlotte-Morgan:
oh - so i can do this. but @timlinux could change the order in the dock?

samnawi:
I've been tried to create the keyword wizard diagram.
It's like as flow chart which could guide to user to know about the whole process in the keyword wizard.
![selection_162](https://cloud.githubusercontent.com/assets/8871958/9653765/14789716-5250-11e5-8eb1-75845c5b4cea.png)

# cc
@Charlotte-Morgan @iyan31 @adissadis

ismailsunni:
Hi @samnawi 
Before creating the wizard, @borysiasty @timlinux and I also had a discussion about the workflow. And we created a similar diagram like yours for the wizard also.

Do you remember where do we put it @timlinux @borysiasty ?

timlinux:
Hmm - Borys made his in a desktop UML app I think - We also made a version which is here:

https://www.lucidchart.com/invitations/accept/463ccb7f-6494-465e-95eb-832bd4722f67

timlinux:
I just discussed this with @Charlotte-Morgan  - what we are going to do is:

* @Charlotte-Morgan is going to ensure that the ``name`` field in definitions.py are consistent with the terms that she would like to see used
* @timlinux will update the keywords table generator so that it looks up the names instead of showing the keys with underscores removed
* @ismailsunni (really sorry to throw last minute translations at you) to update translations for changed names in definitions.py

[...]

PR #2323, opened 2015-09-03
Wizard updates

borysiasty:


borysiasty:
1. XML files filtered out from the browser
===============================
The browser is no longer cluttered by .xml files


Added an explanation why the chosen layer is unsuitable
===========================================
bad_layer

All this label is grayed out in order to indicate the layer is not selectable. The headers are bold, the mismatching keywords italic. Please note in the case above the layer is classified, while the continuous mode is required. The next line contains Continuous Hazard Unit - because my intention was to only display the **expected** unit/classification, not the one found in the layer.

Charlotte-Morgan:
hi @borysiasty (i found the pull request :))
this looks good and provides users with helpful info +1
 - [ ] I have some reservations about barring people loading data with the correct geometry but incorrect key words (what if someone else assigned the wrong key words last time). 
- [ ] in the screen that shows there are no valid key word; can we tweak the text from "In the next step you will be able to register this layer" to In the next step you will be able to assign keywords to this  this layer" 
- [ ] i like the table approach to presenting the info - please see #2313 for some additional ideas on converting the terminology to more user friendly user facing words. I'm assuming these will also appear here in your solution

timlinux:
> I like the table approach to presenting the info - please see #2313 for some additional ideas on converting the terminology to more user friendly user facing words. I'm assuming these will also appear here in your solution

It looks like @borysiasty is using the name property from the ```definitions.py``` so it should automatically update with tweaks you make to the presentation layer names.

timlinux:
Ok in poking through the code it appears that my above statement is incorrect. @borysiasty we would like to standardise on using the names tor keywords terms from ```definitions.py``` so that as @Charlotte-Morgan improves them they get updated automatically in the UI.

Charlotte-Morgan:
@timlinux - see my comments in #2313 - i did not find all the terms i need in definitions.py

timlinux:
@Charlotte-Morgan ok I will look thanks

@borysiasty for those terms that *are* in ``definitions.py`` we can programmatically fetch them using a pattern like this (this just lists all keys).

```
import safe.definitions

for item in dir(safe.definitions):
    if not item.startswith("__"):
        var = getattr(safe.definitions, item)
        if isinstance(var, dict):
            if var.has_key('name'):
                print var['name']
```

I think we can also get rid of some of the eval statements in your code using a similar pattern. I will add a helper method to inasafe to let you get a definition whose key matches a given key.

FOr the *missing* ones @Charlotte-Morgan mentioned, I will come up with something...

borysiasty:
Thanks @timlinux - I created another issue to not forget about the evals

timlinux:
@borysiasty ok cool - see https://github.com/AIFDR/inasafe/pull/2328/files#diff-844316bf2c24e3915be331fde33f5059R679 for the final (re-usable) method I made to get a definition given a key. I will go ahead and apply this PR.

PR #2324, opened 2015-09-03
Updated minimum needs post processor to fix nan bug.

cchristelis:


timlinux:
Sometimes the smallest changes make the biggest differences! - thanks @cchristelis 

BTW it would be nice if you could add before and after screenies (or even just after) to your PR so that we can see at a glance the improvements made

On OSX you can just do ```command-shift_4``` then drag a region and it will drop the screenie onto your desktop, which you can then just drag and drop into the ticket...

timlinux:
LGTM, please merge

PR #2328, opened 2015-09-04
Better keywords table generation

timlinux:
Implementation for #2313 - let the keywords table keys be shown from definitions list rather. Also refactored the keywords table generation into keywords.io so that it can be re-used elsewhere...

I also made some other tidy ups to definitions.py


ismailsunni:
LGTM

PR #2337, opened 2015-09-04
fix_pylint: with some tentative modification

lucernae:
Some of the changes need reviews

### DIFF
--- a/safe/metadata/property/character_string_property.py
+++ b/safe/metadata/property/character_string_property.py
@@ -11,15 +11,13 @@ Contact : ole.moller.nielsen@gmail.com
      (at your option) any later version.
 """
 from types import NoneType
+from safe.metadata.property import BaseProperty
 
 __author__ = 'marco@opengis.ch'
 __revision__ = '$Format:%H$'
 __date__ = '27/05/2015'
 __copyright__ = ('Copyright 2012, Australia Indonesia Facility for '
                  'Disaster Reduction')
-
-
-from safe.metadata.property import BaseProperty
 
 
 class CharacterStringProperty(BaseProperty):
@@ -35,8 +33,8 @@ class CharacterStringProperty(BaseProperty):
         super(CharacterStringProperty, self).__init__(
             name, value, xml_path, self._allowed_python_types)
 
-    @staticmethod
-    def is_valid(value):
+    @classmethod
+    def is_valid(cls, value):
         # any string sequence is valid.
         return True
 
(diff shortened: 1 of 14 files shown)


safe/metadata/property/character_string_property.py
```diff
@@ -35,8 +33,8 @@ def __init__(self, name, value, xml_path):
         super(CharacterStringProperty, self).__init__(
             name, value, xml_path, self._allowed_python_types)
 
-    @staticmethod
-    def is_valid(value):
+    @classmethod
+    def is_valid(cls, value):
         # any string sequence is valid.
         return True
```
lucernae: Please review this changes, is this correct?
The old one raises pylint because it had different args with the base class

safe/metadata/property/url_property.py
```diff
@@ -37,8 +37,8 @@ def __init__(self, name, value, xml_path):
         super(UrlProperty, self).__init__(
             name, value, xml_path, self._allowed_python_types)
 
-    @staticmethod
-    def is_valid(value):
+    @classmethod
+    def is_valid(cls, value):
         # TODO (MB): this check could be a bit stronger
         if value is None or value.isValid():
             return True
```
lucernae: Same comment with above, please review this.

safe/metadata/provenance/provenance_step.py
```diff
@@ -54,7 +54,9 @@ def __str__(self):
         :return: the string representation
         :rtype: str
         """
-
+        # RMN:
+        # FIXME: Mismatched arguments. Should we delete the extra tags?
+        # pylint: disable=too-many-format-args
         return "%s: %s\n%s" % (
             self.time, self.title, self.description, self.data)
```
lucernae: We had a mismatch between placeholder and arguments. 3 placeholders and 4 args.
Should we delete one of the arg?

safe/metadata/property/date_property.py
```diff
@@ -38,8 +38,8 @@ def __init__(self, name, value, xml_path):
         super(DateProperty, self).__init__(
             name, value, xml_path, self._allowed_python_types)
 
-    @staticmethod
-    def is_valid(value):
+    @classmethod
+    def is_valid(cls, value):
         # the date types constructors already complain if a date is not valid.
         return True
```
lucernae: Same comment with url_property.py

lucernae:
Could you guys review the line I commented?
I'm not too sure about that.

@ismailsunni @timlinux

safe/metadata/provenance/provenance_step.py
timlinux: Yes ... I assume self.data is not needed anymore...

timlinux:
@lucernae your changes LGTM - for the class property / static method change, I don't know what impact it will have, but I think it will affect us only in 3.3 so lets merge, but flag those for @mbernasocchi 's attention someday if he could take a little look...

PR #2356, opened 2015-09-11
Make parameter widgets look more beautiful

### LINKED ISSUE #2355: We need to make the parameters widgets more attractive
Currently there is a lot of padding in the parameters widgets and the widgets are not always laid out following 'normal' ui conventions. For example the checkbox items appear to the right of their labels instead of to the left. There are also too many frames which makes the parameters visually cluttered.

# Proposed solution

* Use NoFrame style for container widgets
* Move labels to right of checkboxes
* Remove padding around widgets
* Generally tweak to make it look more beautiful

timlinux:
fix #2355 - this fix:

* remove excess padding making parameters more compact
* removes zebra striping of widgets
* puts checkbox labels to the right of checkboxes
* hides frame borders to flatten everything out.

Here is the before view:

screenshot from 2015-09-11 13-35-09

and here is what it looks like with my PR:

screenshot from 2015-09-11 13-31-58

### DIFF
--- a/safe_extras/parameters/qt_widgets/boolean_parameter_widget.py
+++ b/safe_extras/parameters/qt_widgets/boolean_parameter_widget.py
@@ -25,12 +25,18 @@ class BooleanParameterWidget(GenericParameterWidget):
         """
         super(BooleanParameterWidget, self).__init__(parameter, parent)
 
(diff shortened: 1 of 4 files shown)


Charlotte-Morgan:
nice :)

Issue #2322, opened 2015-09-03
BUG : Layer order of the impact result in the map


samnawi:
# Problem
When we're doing some testing, we found that the layer impact in the result sometimes in the above, sometimes in the bottom of the result. Apparently, if occur because of the selected layer in the table of content.

1.If there is no layer that selected, the result automatically will appear in the above of all layers. 
![selection_170](https://cloud.githubusercontent.com/assets/8871958/9655831/20d46928-525e-11e5-8de7-0d26196b51af.png)

2.But, if the user click a layer in the table of content, the result automatically will appear in the above of the selected layer. If the position of a layer that selected is in the bottom of exposure layer, so the result will not show in the map.
![selection_172](https://cloud.githubusercontent.com/assets/8871958/9656192/b14efe62-5260-11e5-9fa4-d77e53b81f4a.png)

# Proposed Solution
Could we make the result of the impact function always appear on top of all layers, altough there is layer that selected as the example in above?
Make it happy.
 
# CC
@timlinux 
@Charlotte-Morgan 
@fredychandra 

Thank you.

timlinux:
So I will add some logic to always insert the impact layer 1 layer above the exposure layer.

PR #2364, opened 2015-09-15
Fix for #2322 - ensure impact layer is added just above exposure layer

timlinux:
Note I expect a couple of failed tests - will fix that in a follow up PR. In the mean time, it would be great if @adissadis and @samnawi could try this out and see if it works ok for you.

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -30,7 +30,8 @@ from qgis.core import (
     QgsMapLayerRegistry,
     QgsCoordinateReferenceSystem,
     QGis,
-    QgsProject)
+    QgsProject,
+    QgsLayerTreeLayer)
 # noinspection PyPackageRequirements
 from PyQt4 import QtGui, QtCore
 # noinspection PyPackageRequirements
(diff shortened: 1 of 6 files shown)


samnawi:
@timlinux LGTM
+1 for this

safe/gui/widgets/dock.py
```diff
@@ -1319,6 +1322,77 @@ def prepare_analysis(self):
 
         return analysis
 
+    def add_above_layer(
```
ismailsunni: I think it's enough to put the argument in one line ( < 80 chars)

safe/gui/widgets/dock.py
```diff
@@ -1319,6 +1322,77 @@ def prepare_analysis(self):
...
+
+        .. note:: This method works in QGIS 2.4 and better only. In
+            earlier versions it will just add the layer to the top of the
+            layer stack.
+
+        .. seealso:: issue #2322
+
+        :param existing_layer: The layer which the new layer
+            should be added above.
+        :type new_layer: QgsMapLayer
```
ismailsunni: should it be `existing_layer`?

ismailsunni:
LGTM with some minor comments

PR #2371, opened 2015-09-17
Fix potential bug with realtime extract process:

lucernae:
It can happen when the source wasn't finished yet copied.
So the extracted file is not finished.
This fix recopy the original file when the file is different.


timlinux:
LGTM thanks

PR #2377, opened 2015-09-17
fix #2313 - Use sensible ordering for keywords in dock.

### LINKED ISSUE #2313: UI: We need consistent user facing terminology

timlinux:
fix #2313 - Use sensible ordering for keywords in dock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ System Requirements
 History
 =======
 
+* In March 2011, Risk In A Box (the predecessor to InaSAFE) was built as a web 
+  application running as a GeoNode Plugin.
 * InaSAFE development as a QGIS plugin started in January 2012.
 * In October 2012, version 1.0 of InaSAFE was launched at the [5th Asian ministerial conference on disaster risk reduction (AMCDRR)](
   http://www.unisdr.org/we/inform/events/23540) where it was demonstrated to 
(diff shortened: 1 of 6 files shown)


Charlotte-Morgan:
looks good - can we be a little more specific to separate data fields from metadata fields?
See suggestion in #2313

samnawi:
LGTM, but with some note that I've write in #2313.
And I prefer to make "Field" and "Attribute values" closest / adjacent, not like this picture
selection_204

Thank you

timlinux:
Here is a little video showing the keywords rendered for each of the standard test datasets:

https://youtu.be/SJVmzJfgRLA

timlinux:
@ismailsunni please review

ismailsunni:
LGTM

PR #2378, opened 2015-09-17
fix #2210 - Use context based feedback during analysis to properly re…

### LINKED ISSUE #2210: UI: Dock message does not match user selected analysis extent
# Problem
the user can now select analysis extent based on the intersection of data, window, book mark etc but in the dock, gets a message saying the analysis is based on window extent

# proposed solution
change the user message text in the dock to say that the analysis extent is as set by the user in the analysis extents selector

timlinux:
fix #2210 - Use context based feedback during analysis to properly reflect what clipping behaviour is in place.


timlinux:
@ismailsunni please review

ismailsunni:
LGTM

PR #2412, opened 2015-09-24
update label typo for ISO metadata (s/19155/19115/)

tomkralidis:
Per subject.  Note that there binary files `i18n/inasafe_es_ES.qm` and `i18n/inasafe_id.qm` also need updating, but I'm not sure how to do that/whether that is done by some other automated process.  If there are docs on how to do this, I can add to this PR.

### DIFF
--- a/i18n/inasafe_af.ts
+++ b/i18n/inasafe_af.ts
@@ -5789,7 +5789,7 @@ to
     </message>
     <message>
         <location filename="options_dialog_base.ui" line="422"/>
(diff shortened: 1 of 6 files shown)


timlinux:
Thanks @tomkralidis. @ismailsunni will update the ``qm`` (compiled Qt translation files) before the next release. I will back port your fixes to the 3.2.1 release we plan to put out in a few days. 

Its really great to have your contribution - thank you very much!

Issue #2389, opened 2015-09-18
Action Checklists in all of IF


samnawi:
# Problem
In Roads impacted report, there is no checklist action in the report.

![selection_229](https://cloud.githubusercontent.com/assets/8871958/9955453/69cdd5ec-5e1b-11e5-902e-efc382d92b83.png)

# Proposed Solution
Make the action checklist for the all of IF.
@fredychandra for the next version, maybe you can give the advice about the content of action checklist.

# CC
@Charlotte-Morgan @timlinux 

Thank you.

timlinux:
If you can provide input on what actions you would like to see, we can include this in 3.2.1

fredychandra:
HI @samnawi and @ismailsunni 
Here are two check lists that I propose to use in  Road Impact Analysis Result. We can add more if find something else. 

1.  Try to Identify which potential roads that able to access for evacuation and logistic distribution 
     purpose. What type of transportation reliable to use in that order?
2. Identify what equipment needed to open access of affected roads. Where the resources or  
    equipment to open access affected roads located? who/what department responsible to mobilize? 

Thank you

Charlotte-Morgan:
@fredychandra - thank you for your suggestion
Does this 'plain english version" fit with the other action list styles?
Which roads can be used to evacuate people or to distribute logistics?
What type of vehicles can use the unaffected roads?
What sort of equipment will be needed to reopen roads & where will we get it?

fredychandra:
:+1: Charlotte, this looks fantastic :)

ismailsunni:
Sample result:
<img width="377" alt="screen shot 2015-09-29 at 4 54 58 pm" src="https://cloud.githubusercontent.com/assets/1421861/10161043/81dbc3e8-66cb-11e5-930c-88e49fb7a90f.png">
The branch that I work on it: https://github.com/ismailsunni/inasafe/tree/fix_2388_2389

Please check @samnawi @fredychandra @Charlotte-Morgan

ismailsunni:
Closed in #2421

fredychandra:
HI @ismailsunni 
I just checked again the action list on IF : Flood on Road. I realize that the action 1 & 2 already consist in action no 2 - 5. We should erase no 1 & 2 action list. 

![capture](https://cloud.githubusercontent.com/assets/6345285/11706217/28c25592-9f29-11e5-9739-2c9a0a8cd8b9.PNG)

thank you

fredychandra:
# Problem

[...]

Issue #2388, opened 2015-09-18
Add a Note in All Impact Function


samnawi:
# Problem
When we doing systematics testing, we found that there is no Note in the roads impact (Both of flood and tsunami).
![selection_228](https://cloud.githubusercontent.com/assets/8871958/9955336/545fa72c-5e1a-11e5-8f15-2766c6175dd2.png)

# Proposed Solution
Give notes in here, to explain about the data and the process.

# CC
@Charlotte-Morgan @fredychandra @timlinux 

Thankyou

timlinux:
Can you provide some input on what you would like to see in the notes please?

fredychandra:
Propose notes for road IF, Flood and Tsunami 

This is for Flood impact report
1. Roads are flooded when flood levels exceed **1.0 m**
2. Roads are closed if they are flooded 
3. Roads are open if they are not flooded

This is for Tsunami Impact report
1. Roads are inundated when water levels exceed **1.0 m**
2. Roads are closed if they are inundated 
3. Roads are open if they are not inundated

**This value** should be read from user threshold in the impact function

ismailsunni:
Hi @fredychandra 
Thanks for the notes.

Btw, we also have flood polygon on road. And the notes above can't be applied in this IF since we use attribute value to identify which polygon is inundated. For example, in attribute FLOODPRONE, and the value can be YES / NO.

If you have notes for this IF, I can add it to the IF on version 3.2.1.

We can use the same notes for point 2 and 3. But for point 1, we can't. I suggest something like:

Roads are flooded when it's located on hazard polygon that has [flood value] in [attribute].

May be you can give a better alternative.

Thanks.

samnawi:
@ismailsunni My propose for point1, can we use like in the building?
So, the note will written like this :
1. Roads are said to be inundated when in a region with field "FLOODPRONE" in "YES" .

cc : @Charlotte-Morgan how about this?

ismailsunni:
@samnawi That's better. I will change it.

PR #2421, opened 2015-09-29
Fix 2388 2389

ismailsunni:
For #2388 and #2389


samnawi:
Hi @ismailsunni I've tried your PR.
LGTM.

PR #2431, opened 2015-10-03
[IFCW] Don't propose impact layers as IF inputs. Fixes #2414.

### LINKED ISSUE #2414: Keywords wizard should not offer to use impact layer as hazard
While showing InaSAFE to @akbargumbira the other night we were playing with the wizard and noticed that it will try to use the impact layer from a previous analysis as a possible hazard layer. Here is a screen sequence to illustrate:

screen shot 2015-09-24 at 09 41 16

screen shot 2015-09-24 at 09 41 26

screen shot 2015-09-24 at 09 41 50

screen shot 2015-09-24 at 09 42 01

screen shot 2015-09-24 at 09 42 14

On the last screen you can see that it fails to load the impact layer keywords and wants to offer the user the chance to use this layer as a hazard layer, registering keywords in the process.

I could see in some distant future where we may want to rerun an analysis using the impact layer as the exposure layer, but I don't see it ever making sense to offer it for use as the hazard layer. **For the short term though I want to suggest that you simply do not offer impact layers as either hazard or exposure candidates to the user in the wizard workflow.**

borysiasty:


timlinux:
Thanks @borysiasty !

PR #2433, opened 2015-10-04
Bumped version to 3.2.1 for release

timlinux:


samnawi:
I've been tested in my Windows 7-64 Bit PC, using QGIS 2.8.3.
I've got the problem with keyword. The InaSAFE says if my keyword is not relevant with the 3.2.1 version, but when I try to reiterate make the keyword, the problem is still occurred.

PR #2439, opened 2015-10-06
Fix pr 2433

ismailsunni:
This contains fix for PR #2433
We only check minor version of keyword and InaSAFE version.


samnawi:
LGTM

PR #2443, opened 2015-10-19
Revamped context help

timlinux:
This PR updates all dialogs to use the new context help system. It would be great if you could also review @Charlotte-Morgan . We have tried to implement the system in a way that makes it easy to dip in and tweak the wording without getting bogged down in a lot of code. The help texts are not intended to be an exhaustive treatise on the functions of each dialog, but rather to provide enough context that the user can understand what the various options do. I expect that we will augment and fine tune these descriptions over time - especially when it is evident that users were not able to comprehend the options presented in a given dialog.


timlinux:
Currently this test is failing:

```
======================================================================
FAIL: Test the IFCW mode.
----------------------------------------------------------------------
  File "/home/travis/build/AIFDR/inasafe/safe/gui/tools/test/test_wizard_dialog.py", line 1594, in test_input_function_centric_wizard
    self.check_current_step(step_fc_extent, dialog)
  File "/home/travis/build/AIFDR/inasafe/safe/gui/tools/test/test_wizard_dialog.py", line 115, in check_current_step
    self.assertEqual(expected_step, current_step, message)
AssertionError: Expected 28 but I got 24
-------------------- >> begin captured logging << --------------------
```

timlinux:
@ismailsunni any idea why our changes would result in the above error?

ismailsunni:
```
step_fc_agglayer_origin = 24
step_fc_extent = 28
```
Looking at the test's code, it seems, the next button is not enabled (or something like that). So that it stays in the same step (step 24, choosing aggregation). I need to try to run the test in real steps first though to check it.

But, I don't know why help will disturb this test.

Charlotte-Morgan:
Hi @timlinux - this looks good. 
At the recent training i was asked about context help and said that it was coming :)
Definitely some tweaking required to make sure terminology remains consistent with concepts documents. Will this include help for the wizards?

timlinux:
@Charlotte-Morgan Yes I we can tweak more as needed, no problem.

I wasn't planning to write context help for the wizards - not including help was a design choice because
 the nature of a wizard is such that it should guide you step by step through the process (essentially it *is* the help).

If they are not meeting this goal I think it would be more valuable to improve the information shown on the wizard dialog itself so that the choices the user needs to make are well described.

If you feel strongly that we should actually have user documentation for the wizards, I am happy to revisit my above thinking and add context help to the wizards.

timlinux:
@borysiasty can you look at the failing test?

ismailsunni:
One last failed test:
```
======================================================================
ERROR: Test local help returns no error if the file is not found.
----------------------------------------------------------------------
  File "/home/travis/build/AIFDR/inasafe/safe/utilities/test/test_help.py", line 51, in test_local_help_better
    _show_local_help(context='keywords')
  File "/home/travis/build/AIFDR/inasafe/safe/utilities/help.py", line 81, in _show_local_help
    raise HelpFileMissingError('Help file not found: %s' % base_url)
HelpFileMissingError: Help file not found: /home/travis/build/AIFDR/inasafe/docs/en/keywords.html
```

timlinux:
Thanks @ismailsunni - strange that it only shows that error now....

PR #2447, opened 2015-10-21
Add support for OSM backdrop layer

### LINKED ISSUE #2446: We should provide an alternative for OSM backdrop support in QGIS
# Problem

Currently many users are using the 'open layers' plugin to add google and other tile set backdrops to their maps. They however always encounter issues with printing when using this plugin because of the asynchronous way in which it renders via webkit.

Whilst supporting a user in Malawi, I created a GDAL wrapper around OSM which I think alleviates the printing issue allowing large format and atlas maps to be generated with OSM as a backdrop. Unfortunately for licensing reasons we can't use google imagery / bing imagery in this way, but I think it would be good to add a button to our toolbar that lets the user easily add the OSM backdrop layer to their map.

# Proposed solution

* Drop in the goal xml service definition for OSM tiles into resources
* Add a button to the toolbar that adds the OSM tile layer to the map

# Caveats

* The user needs to be online in order to use the tile sets
* The map should ideally be printed at one of the predefined OSM scales for good quality

timlinux:
This will fix #2446 - it adds a button to the toolbar that lets you drop in the OSM backdrop layer as a gdal tiled layer. @vdeparday and @adissadis FYI


timlinux:
Here is what it looks like if you add it to a project with no other layers:

screenshot from 2015-10-21 11-30-39

And here is what it is like when zoomed in:




screenshot from 2015-10-21 11-44-53

samnawi:
LGTM @timlinux 
But, I think it will be good if the OSM layer can be put automatically in the bottom from all of existing layer.
image

Thank you.

timlinux:
Thanks very much for your feed back @samnawi - it is done in 56750c1 - please tell me if it works for you as expected.

samnawi:
:+1: for this @timlinux 

image

Thank you

Charlotte-Morgan:
This is a good concept and i agree with it being under the other data; however I have been unable to generate a pdf with this layer included. QGIS stops responding  - and i get impatient and crash out :)
Oh - in case you are wondering, i tested with Jakarta Flood.

timlinux:
@Charlotte-Morgan OK I will test the pdf creation here.

timlinux:
I tested here with QGIS 2.8.3 on Win 10 and it worked fine for me. I'll take a look at your specific case to see if I can see a reason for it crashing out post merge ok?

PR #2465, opened 2015-11-17
fix #2464 - support point buildings when doing polygon flood analysis

### LINKED ISSUE #2464: We should make the polygon flood on buildings IF support point buildings too
Currently you can't do polygon flood impact on buildings even though the logic is basically the same as polygon flood on polygon buildings. We should update the safe/impact_functions/inundation/flood_vector_building_impact/impact_function.py to support point datasets too.

timlinux:
fix #2464 - support point buildings when doing polygon flood analysis

screen shot 2015-11-17 at 07 07 50

@ismailsunni @Charlotte-Morgan FYI


ismailsunni:
LGTM

ismailsunni:
I have re-run scrutinizer test, it pass now @timlinux

timlinux:
Thanks @ismailsunni

PR #2478, opened 2015-11-25
Fix #2475 - support on the fly projection for flood on buildings

### LINKED ISSUE #2475: Polygon flood on buildings fails if hazard and exposure are in different CRS
We have an issue reported byRick working in Moçambique where his flood analysis fails. I inspected the data and the issue seems to be caused by the fact that his building points are in UTM while his flood data is in EPSG:4326.

# Proposed solution:

Visit every vector on vector IF and transform the geometry of the exposure whenever any intersection is done.

timlinux:
This patch will retroject the exposure to the hazard CRS ensuring that the analysis will work even if the two layers are in different CRS's

screen shot 2015-11-25 at 15 46 52


ismailsunni:
the code itself is LGTM, but I didn't really check the logic behind it.

timlinux:
Thanks @ismailsunni I am going to merge it and then do another PR with more invasive changes for support OTF everywhere from the base class.

Issue #2485, opened 2015-11-30
Extra Keywords Step is Missing in Wizard


ismailsunni:
Problem
-----------

Previously, we have this extra keyword step in wizard for example set the name attribute in volcano IF. I found that this step is missing when I want to check @Samweli 's PR #2442. We need to take it back since it's an important thing for the wizard.

For bug reproduction, just use wizard with any volcano IF. It will not ask you about the extra keywords.

cc @borysiasty @timlinux

ismailsunni:
This is invalid. The additional keywords step indeed not showing up in IFCW, since it shows up in keywords wizard.

Have a good day :D

borysiasty:
Exactly, just started writing the same :)

Charlotte-Morgan:
yay - everything is working well

PR #2486, opened 2015-11-30
Update unit test for additional keywords

ismailsunni:
This PR only update unit test related to additional keywords for IF. To make sure that the problem is not in IF manager for problem #2485


lucernae:
LGTM

PR #2487, opened 2015-11-30
Wizard updates

borysiasty:


safe/gui/tools/wizard_dialog.py
```diff
@@ -167,25 +169,26 @@
...
-step_fc_agglayer_origin = 24
-step_fc_agglayer_from_canvas = 25
-step_fc_agglayer_from_browser = 26
-step_fc_agglayer_disjoint = 27
-step_fc_extent = 28
-step_fc_extent_disjoint = 29
-step_fc_params = 30
-step_fc_summary = 31
-step_fc_analysis = 32
+step_kw_summary = 14
```
timlinux: For when you have time - could we not get these numbers by querying the Qt object tree and so keep the code below a little more future proof?

timlinux:
LGTM thanks

PR #2492, opened 2015-12-02
Formatting cleanups to wizard

timlinux:


timlinux:
@borysiasty I generally prefer not to use messages like that because when you get them you don't see the discrepancies between the two variables being compared. So if you need the message for clarity then you should concatenate it onto the assert message some how I guess.

timlinux:
Note to self to freshen this to get it mergable...

PR #2506, opened 2015-12-06
Fix #2504

### LINKED ISSUE #2504: to_html() got an unexpected keyword argument 'wrap_slash'
When I add a layer, I often got this error message : 
layer

* QGIS 2.12
* InaSAFE develop

borysiasty:
This is an dirty and easy fix of #2504 - see my comment there.

The code is just copied from safe.messaging.item.text.Text.html(). As the table contains html tags with slashes, the parameter is definitely not meant to be set to True. However, Maybe the parameter should take no action in this class, and the docstring should say this parameter is abstract?  @timlinux could you please decide which solution is less bad? :)

### DIFF
--- a/safe/messaging/item/table.py
+++ b/safe/messaging/item/table.py
@@ -60,8 +60,14 @@ class Table(MessageElement):
         else:
             raise InvalidMessageItemError(item, item.__class__)
 
-    def to_html(self):
+    def to_html(self, wrap_slash=False):
         """Render a Table MessageElement as html
+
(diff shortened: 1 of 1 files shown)


timlinux:
Doh' sorry its my fault! I'm taking a look to see which is better.

timlinux:
Thanks @borysiasty I made an alternative fix - see #2508 - your fix gave me the pointer to what to do though so thanks very much!

Issue #2499, opened 2015-12-04
Use Date Picker in Keywords Wizard


ismailsunni:
Problem
-----------
In the next version of metadata keyword that use ISO 19115, we will store the date or datetime as python datetime object. So, instead ask the user to input the date in plain string, it would be better to use date picker.
And when the new metadata keywords is ready, we will store it as a datetime object.

PR #2509, opened 2015-12-07
Use Date Picker in Keywords Wizard. Get rid of eval() statements.

### LINKED ISSUE #2329: Wizard: ged rid of eval() statements
As @timlinux suggested, getattr should be rather used for accessing dictionaries from definitions.py.
A helper method will be written to get a definition whose key matches a given key.

```
import safe.definitions

for item in dir(safe.definitions):
    if not item.startswith("__"):
        var = getattr(safe.definitions, item)
        if isinstance(var, dict):
            if var.has_key('name'):
                print var['name']
```

borysiasty:
1. Get rid of eval() statements. Fixes #2329
2. Use QgsDateTimeEdit for the `date` keyword. Fixes for #2499. 

In the existing test dataset I found the date keyword is usually missing, but when present, it's always stored as 'dd-MM-yyyy HH:mm'. So I made the wizard always using that format (no localisation!). I hope it's ok.

Please ignore the other reverted commits - I just didn't know how to squash commits that are already not on the top of the stack.

### DIFF
--- a/safe/gui/tools/test/test_wizard_dialog.py
+++ b/safe/gui/tools/test/test_wizard_dialog.py
@@ -499,7 +499,7 @@ class WizardDialogTest(unittest.TestCase):
         message = 'Source Url should be empty'
         self.assertEqual(dialog.leSource_url.text(), '', message)
         message = 'Source Date should be empty'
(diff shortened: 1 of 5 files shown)


timlinux:
Thank youfor this @borysiasty !

Issue #2469, opened 2015-11-19
RFC: do we need an OSM downloader query specifically for Jakarta Flood data


Charlotte-Morgan:
# Problem
There are some great flood related data in OSM we should make it easy for people to use

# Proposed solution
I propose that we create some specific OSM downloader queries to download flood hazard data and evacuation centers for Jakarta.  These data can then be used as 'real world' data to test flood evacuation functions and to build smarter flood impact functions.

* Evacuation Centers.
Overpass turbo query is evacuation_center=yes
![image](https://cloud.githubusercontent.com/assets/5743135/11268778/4b7bca36-8ee5-11e5-91b9-f8ecdbfb94e6.png)
retain only polygons (current data)
these data include attributes on the capacity, water source and number of toilets available
more advice on fields to retain and filtering to create derived attributes later

* Flood prone areas
Overpass turbo query is flood_prone=yes
admin_level 8 and 9 have attributes for flood depth and duration
![image](https://cloud.githubusercontent.com/assets/5743135/11268996/b1eafb06-8ee6-11e5-90e1-f7d6db69e542.png)
more advice on fields to retain and filtering to create derived attributes later

# Institutional outcomes
By making better use of these data in maps and analyses, we will also encourage their maintenance.

# cc
@Gustry @timlinux @iyan31 @fredychandra @samnawi

timlinux:
@Charlotte-Morgan can you put the raw overpass queries in here as text too please?

Charlotte-Morgan:
I hope this is what you want

evacuation_center=yes
``
/*
This has been generated by the overpass-turbo wizard.
The original search was:
“evacuation_center=yes”
*/
[out:json][timeout:25];
// gather results
(
  // query part for: “evacuation_center=yes”
  node["evacuation_center"="yes"]({{bbox}});
  way["evacuation_center"="yes"]({{bbox}});
  relation["evacuation_center"="yes"]({{bbox}});
);
// print results
out body;
>;
out skel qt;
``

and 

flood_prone= yes
``
/*
This has been generated by the overpass-turbo wizard.
The original search was:
“flood_prone=yes”
*/
[out:json][timeout:25];
// gather results
(
  // query part for: “flood_prone=yes”
  node["flood_prone"="yes"]({{bbox}});
  way["flood_prone"="yes"]({{bbox}});
  relation["flood_prone"="yes"]({{bbox}});
);
// print results
out body;
>;
out skel qt;
``

[...]

PR #2510, opened 2015-12-08
replace "potential IDP" by "evacuation center"

Gustry:
inasafe openstreetmap downloader_002

### DIFF
--- a/safe/gui/tools/osm_downloader_dialog.py
+++ b/safe/gui/tools/osm_downloader_dialog.py
@@ -303,8 +303,8 @@ class OsmDownloaderDialog(QDialog, FORM_CLASS):
             feature_types.append('building-points')
         if self.flood_prone_flag.isChecked():
             feature_types.append('flood-prone')
-        if self.potential_idp_flag.isChecked():
-            feature_types.append('potential-idp')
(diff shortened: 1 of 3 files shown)


timlinux:
LGTM thanks

Charlotte-Morgan:
Hi @timlinux & @Gustry 
I'm reopening this because I'm not sure on the intention here. The label in the OSM downloader has changed but the features returned are still schools, hospitals etc. ie the outcome is still misleading.
In Jakarta, evacuation centres are tagged as evacuation_centres - see #2469 but this tag is not used elsewhere in the world.
Changing the label in this case is still misleading. maybe a better label would be lifeline buildings (and we should check with HOT as to which building types meet this description).

It would also be good to update the context help to describe the additional options.

@iyan31 @raniedwianugrah @adissadis - can you please comment on life line buildings?

Gustry:
Hi @Charlotte-Morgan,
The label in the OSM downloader and the result have changed. It's the same legend (for now), but it requests **only** `evacuation_center=yes` and the attribute table is very different (much more relevant).

Before, it requested hospital, sport centre, school, public building, place of worship ...

PR #2511, opened 2015-12-08
inform the user about the network status in the OSM downloader

Gustry:
About the ``ContentNotFoundError``, this error shoudn't happen because it means than osm-reporter is not running the latest version.
Only the first two errors might happen in production.


timlinux:
LGTM thanks

Issue #667, opened 2013-09-16
OSM Downloaded Does not work with internet proxy


wonderchook:
# Problem

When using proxy with OSM downloaded data never downloads and 'socket error' occurs.

# Expected results

When using a internet proxy OSM data should still download for use in InaSAFE. Would expect InaSAFE to use QGIS proxy configuration.

Gustry:
Instead of using [QNetworkAccessManager](https://github.com/inasafe/inasafe/blob/develop/safe/utilities/osm_downloader.py#L25), perhaps we should use [QgsNetworkAccessManager](http://qgis.org/api/classQgsNetworkAccessManager.html) ?

Gustry:
InaSAFE is now using the QNetworlAccessManager provided by QGIS. Any bug should be raised in the QGIS bug tracker.

PR #2516, opened 2015-12-11
#667 Use the network manager provided by QGIS and move it to the file…

Gustry:
#667 Use the network manager provided by QGIS and move it to the file_downloader file


safe/utilities/file_downloader.py
```diff
@@ -35,12 +35,9 @@
 
 class FileDownloader(object):
     """The blueprint for downloading file from url."""
-    def __init__(self, manager, url, output_path, progress_dialog=None):
+    def __init__(self, url, output_path, progress_dialog=None):
         """Constructor of the class.
```
timlinux: .. versionchanged:: 3.3 removed manager parameter.

Gustry: Done

timlinux:
Thanks will you go ahead and merge? By the way did you check which version of QGIS this was introduced? Originally we were supporting back to 2.0. Now our officially supported version is the LTR. If we are going to move the requirement up, we need to edit metadata.txt to indicate the new minimum version.

Gustry:
I'm checking in QGIS 2.8.4, but it should be fine as QgsNetworkManager is older than 2.8 in the QGIS repository.

Gustry:
Tested on QGIS 2.8.4, it's ok.
I'm going to merge even if I don't have any proxy. But it should work as we are going to use the QGIS instance.

PR #2522, opened 2015-12-16
add styles to flood test data and fix data audit for census

Gustry:


timlinux:
Can you include a screenie of the rendered styles for those looking at this PR?

Thanks

Gustry:
Sorry, I forgot that you want the flood prone in blue. I've fixed that.

This is only a QML for test layers so the aim of this PR is only for us. But maybe we should define some common colors : flood prone YES/NO ...

style

timlinux:
Thanks that looks good @Gustry - perhaps put it to 50% transparency too?

Gustry:
I will merge. I have already set transparency to 30%.

PR #2528, opened 2015-12-21
WIP : refactoring analysis

Gustry:
- [ ] safe_layer.py : docstring, implementation


Gustry:
I noticed that there a few methods [show/send]_dynamic_message, static_message ... in safe. Do you think I can continue the refactoring using safe/common/signals.py as in my PR ?

CC @timlinux @lucernae @ismailsunni

ismailsunni:
From me, I think it's ok to do it. Perhaps, we can use it in another place.

Gustry:
Thanks @ismailsunni. Yes this code is duplicated in inasafe, I can use it in some places if I continue this small refactoring.

timlinux:
@Gustry can you freshen your PR so it is mergable? Can you update on the status?

Gustry:
This PR has been split into many small PR. I'm going to close this one.

Issue #2419, opened 2015-09-28
BUG : Wrong Classification in the inundated building


samnawi:
# Problem

When I run InaSAFE using flood raster VS building, I found some building that get wrong classification in the result. As in this picture :
![selection_219](https://cloud.githubusercontent.com/assets/8871958/10127591/5e1f5184-65ce-11e5-8770-0c2e49f3ea25.png)

Based on the legend, the dark blue has >= 1 m depth, the light blue has 0.5 - 1 m depth, and the cyan has 0.1 - 0.5 m depth.
Some of building that 100% located in depth < 1 m, is categorized to inundated (>= 1.0 m) - red color.
Some of building that 100% located in depth >= 1 m, is categorized to wet ( 0 m - 1.0 m) - orange color.

So, I think this issue will be influence the result, and the result possibly has incorrect value.
# PC Platform

Linux Ubuntu 14.04 and Windows 7 64-Bit
QGIS 2.8.3
InaSAFE 3.2 release version
# Proposed Solution

Make it Happy
# CC

@Charlotte-Morgan @fredychandra @timlinux 

Thank you.

Charlotte-Morgan:
I think the "make it happy", needs to look at the analysis extent. Looks like buildings are being included in the impact assessment when they are not coincident with the hazard data.
@ismailsunni - i've assigned this to you, but feel free to pass it on :)

ismailsunni:
It seems @timlinux will know better than me. So, I will pass it to him :). It can be related with analysis extent or interpolation

timlinux:
Probably something that will be fixed by switching to pixel lookups rather than interpolation...

adissadis:
I got this same issue like @samnawi . This issue not only affected to Building polygon but also building points. See my pictures below:
![image](https://cloud.githubusercontent.com/assets/2520159/11947535/d3969616-a89a-11e5-82c2-16b7eeee37a2.png)

It's look like the hazard raster data create some virtual buffer around the hazard area (probably around 8 meters) from pixel who categorize as affected area. This will cause building around those pixel will categorized as affected:
![image](https://cloud.githubusercontent.com/assets/2520159/11947557/4fcf4174-a89b-11e5-8eae-1812346fb551.png)

# Platform
Windows 7
QGIS 2.8.4
InaSAFE 3.2.5 dev-master (sync with Github at 21 December 2015)

timlinux:
[...]

PR #2542, opened 2015-12-31
WIP - Tsunami raster building IF

ismailsunni:


ismailsunni:
Hi @samnawi can you try this new IF?

samnawi:
:+1: 

LGTM

ismailsunni:
Update:
- Adding dry zone
- Change color

screen shot 2016-01-04 at 11 51 48 am

samnawi:
LGTM for add dry class and update the color.
But, there is some note for it :
* Update the note also for dry class
* I found some wrong calculated in this picture

image

Some building that I pointed with the arrow is 100% position in 0 meters, but the classified in low class, not dry class.

Thank you.

timlinux:
@samnawi wrote:

> Some building that I pointed with the arrow is 100% position in 0 meters, but the classified in low class, not dry class. Maybe it's similar with issue #2419.

I think this issue will always be present whilst we are using interpolation. As soon as we can move to direct value lookups (with building classified according to the largest cell value it touches) this issue will go away.

ismailsunni:
@timlinux 
Yes, that's what I thought. Should I continue working on this, or wait for the interpolation removal first?

Issue #2425, opened 2015-10-01
Volcano IF summary needs to show results with units


timlinux:
![screenshot from 2015-10-01 16-34-42](https://cloud.githubusercontent.com/assets/178003/10223653/720fa610-685a-11e5-8f96-c386fa201ea7.png)

So better to have 3km, 5km, 10km in that table

ismailsunni:
Fix in #2544

Issue #2530, opened 2015-12-22
All Population be Affected in Flood IF Result [Bug]


harrymahar:
# Problem
No Population be affected in Flood IF result

# Detail 
I try to run flood IF using flood Jakarta (vector) and Java Population (raster). The result shows number of population affected, people need evacuation and unaffected population. However, there is strange result where number of unaffected population is zero. It means, all people in Jakarta affected by flood eventhough the floodprone only happened in certain area.

# ScreenShoot
![unaffected population](https://cloud.githubusercontent.com/assets/2534179/11947670/0b5c1178-a89d-11e5-892e-4ea5bac85642.PNG)

# Platform
Windows 8.1 64 Bit
RAM 4 GB
QGIS 2.8.4 Wien
InaSAFE 3.2.5 dev-master

# Data
Hazard : Jakarta_Flood_18113__WGS84.shp (Vector)

Exposure : 
*Java_Population_WGS84.tiff (Raster)

Agregation : Jakarta_District_Boundary.shp	

Cc : @timlinux @ismailsunni @samnawi

samnawi:
Hi @harrymahar I get this problem too. Either in WIndows 7 64 Bit and Ubuntu 14.04. Using QGIS 2.8.4 and 2.12.1.

![image](https://cloud.githubusercontent.com/assets/8871958/12031344/6684010a-ae3d-11e5-8ca2-724ed5c5e26c.png)

In the picture above, when I try to doing zonal statistics process using flood polygon as vector layer zone, generate the number of population 1,120,024. If the number is rounded up 1,000, similar with total affected population 1,121,000. It means that the IF just calculate people in the flood polygon area, whereas there is raster population data inside the analysis area, but outside the flood polygon.

adissadis:
@samnawi @ismailsunni i think this will make people confused reading the result. People will think all the people is affected by flood and there is 0 people unaffected. While there are people who not in flooded area but the result counted as 0 people unaffected.

In my opinion, i think we better to remove information "Unaffected Population" for this IF to avoid this misunderstanding.

ismailsunni:
@adissadis can you try my PR for this ticket above? I merely remove the unaffected population from the impact report.

adissadis:
@ismailsunni  Its still appear in impact report:
![image](https://cloud.githubusercontent.com/assets/2520159/12256320/02d6f7e6-b92f-11e5-9a34-01e1cfe2f03e.png)

ismailsunni:
[...]

PR #2544, opened 2016-01-05
Fix 2425 if volcano unit

ismailsunni:
For #2425

### DIFF
--- a/safe/engine/core.py
+++ b/safe/engine/core.py
@@ -280,7 +280,7 @@ def buffer_points(centers, radii, hazard_zone_attribute, data_table=None):
     :param centers: All center of each point (longitude, latitude)
     :type centers: list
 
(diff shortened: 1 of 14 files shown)


ismailsunni:
@samnawi Can you try this PR?

samnawi:
LGTM

selection_011
But from the picture in above, there is some note for it :
* It does not have consistency --> see the red box
* Unaffected population is zero? whereas the unaffected building is not zero. I think the unaffected is the exposure data outside the radius and inside the extent analysis? --> see the blue box (Maybe it's relevant with issue #2530)
* The information order. In the Population IF, Total information is above the radius information, while in the building IF, Total information is under the radius information. --> see the black arrow

samnawi:
LGTM

PR #2546, opened 2016-01-06
Fix 2385 sorting hazard classification

ismailsunni:


samnawi:
@ismailsunni It work for some IF, but I still found a wrong order in Volcano IF (polygon)
selection_026

Also for the consistency, the report will sort from the worst severity / high hazard or low hazard?
selection_030

samnawi:
LGTM
But just a little bit issue about a consistency of the classification. Will we use just like this 

selection_036

Or like this

selection_035

samnawi:
I've test again.
LGTM

PR #2549, opened 2016-01-08
WIP Put impact summary in separated xml file.

ismailsunni:
WIP Put impact summary in separated xml file.


timlinux:
LGTM thanks @ismailsunni just two questions:

1. Can you check that when doing a ``Layer`` -> ``Save as`` operation that the summary table is correctly copied.
2. Can you clarify if in the metadata xml a relative path is saved to the summary table? I know we can do this implicitly, but I think it is better to do it explicitly so that e.g. if you are using a third party metadata viewer you can see a link and click through to the summary table.

ismailsunni:
> Can you check that when doing a Layer -> Save as operation that the summary table is correctly copied.

yes, it is copied.

> Can you clarify if in the metadata xml a relative path is saved to the summary table? I know we can do this implicitly, but I think it is better to do it explicitly so that e.g. if you are using a third party metadata viewer you can see a link and click through to the summary table.

So, we store the path to the html file in the xml, right?. Ok.

Btw, this PR is on hold, since in our last discussion we want to do the generation of the report in impact function.  Anyway, I need to wait for @Gustry 's refactor first.

Makefile
```diff
@@ -185,7 +185,7 @@ testdata:
 	@echo "Updating inasafe_data - public test and demo data repository"
 	@echo "Update the hash to check out a specific data version        "
 	@echo "------------------------------------------------------------"
```
timlinux: This is a great change

timlinux:
Hi @ismailsunni yes agreed - holding out for the html generator will be much cleaner, thanks!

ismailsunni:
I think this PR is no longer needed since we have generated html report from json.

Issue #2543, opened 2016-01-04
Generic Population Vector IF Revisited


ismailsunni:
Based on your @timlinux 's message in gitter:
>  Ismail can you look into the flood on population IF from @Samweli - it needs to be changed to use ‘population’ instead of ‘area’ and the exposure should be continuous, not classified I think. Also can you please test it with @samnawi 's test dataset (not redistributable) and make sure it works properly with that?

I revisited the IF, and come up some possible improvements:
- [x] The value in population field of impact layer should represent the calculated number of population in the polygon. See below:
<img width="857" alt="screen shot 2016-01-04 at 8 57 24 am" src="https://cloud.githubusercontent.com/assets/1421861/12082205/3f9aa6ce-b2c1-11e5-8991-86610cbe4ff2.png">
- [x] Change exposure type from exposure area to exposure population.
- [x] Show action list and notes and other common information in the impact layer. See another IF for reference.
- [x] Change exposure layer mode from classified to continuous.
- [x] Show area name (currently, it shows area id in the impact report)
- [x] Show number of population in each hazard level
- [x] Move area_name_field from additional_keyword to exposure_class_field (if possible)

cc @Samweli

Samweli:
Ok @ismailsunni, i'm working on the improvements, i will get back once finished.
Thanks.

ismailsunni:
Hi @Samweli 
Check also #2519

Samweli:
Ok @ismailsunni.

Charlotte-Morgan:
please review feedback and suggestions in #2519

ismailsunni:
All issues have been addressed in #2551, closing this...

PR #2551, opened 2016-01-08
Generic Population Vector IF Improvements

Samweli:
## This Pull Request addresses #2543

The following are the changes are made
 1. Changed exposure type from exposure area to exposure population.
 2. Added  action list and notes for the IF.
 3. Changed exposure layer mode from classified to continuous.
 4. Showing area name in impact_report instead of area_id
 5. Added new table for showing number of population in each hazard level
 6. Using the right value for population field in the impact layer


safe/definitions.py
```diff
@@ -772,18 +763,18 @@
     'type': 'field',
     'description': tr('Attribute for the area name ')
 }
-area_type = {
+area_type_field = {
     'key': 'area_type_field',
```
timlinux: Maybe ``area_name_field`` would make more sense here?

safe/definitions.py
```diff
@@ -772,18 +763,18 @@
...
 
-area_population_field = {
-    'key': 'area_population_field',
-    'name': tr('Population field'),
+area_id_field = {
+    'key': 'area_id_field',
+    'name': tr('Id field'),
     'type': 'field',
-    'description': tr('Attribute for the population on the area')
+    'description': tr('Attribute for the id on the area')
```
timlinux: Attribute for the **name** of the area probably makes more sense? Unless I misinterpreted, but we are reporting on area names right, not their ids?

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -59,14 +65,40 @@ def __init__(self):
...
     def run(self):
-        """Risk plugin for classified polygon hazard on area with population.
+        """Risk plugin for classified polygon hazard on polygon population.
 
-        Counts areas exposed to hazard zones and then computes the the
-        proportion of each area that is affected. The population in each
-        area is then calculated as the proportion of the original population
-        to the affected area.
+        Counts population in an area exposed to hazard zones and then
+        computes the the  proportion of each area that is affected.
```
timlinux: ``the the`` -> ``the``

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -168,7 +200,7 @@ def run(self):
         extra_keywords = {
             'impact_summary': impact_summary,
             'target_field': self.target_field,
-            'map_title': tr('Affected Areas'),
+            'map_title': tr('Affected Populations'),
```
timlinux: Probably need @Charlotte-Morgan 's input here, but 'people' makes more sense to me than 'Populations'

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -226,10 +258,12 @@ def evaluate_impact(
         :rtype: []
         """
         # Taking area necessary attributes
-        area_id_attribute = self.exposure.keyword('field')
-        type_attr = self.exposure.keyword('area_population_field')
+        area_id_attribute = self.exposure.keyword('area_id_field')
```
timlinux: Can you clarify if you are expecting the user the specify an Area Name or and Area ID? If the former then for consistency, calling the variables area_name_field rather than area_id_field might be good.

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -226,10 +258,12 @@ def evaluate_impact(
...
+        type_attr = self.exposure.keyword('area_type_field')
         area_population_attribute = self.exposure.keyword(
-            'area_population_field')
+            'field')
+        area_name_attribute = self.exposure.keyword(
```
timlinux: ah ok this partly answers some of my queries above - you have a separate property for area_name_attribute....do we need area_id too? Are you assuming that area names may not be unique?

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -257,16 +291,18 @@ def evaluate_impact(
...
             if area_id not in self.all_areas_ids:
                 self.all_areas_ids[area_id] = 0.
 
-            area = geometry.area()
-
-            all_areas[area_type] += area
             self.all_areas_ids[area_id] += geometry_area
+            all_areas[area_type] += area
+
+            if area_id not in self.areas_names:
```
timlinux: Also adds to doubt as to whether we are dealing with id's or names.

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -364,33 +400,55 @@ def assign_impact_level(
...
-
         else:
-            impacted_feature.setGeometry(impact_geometry)
             unaffected_feature.setAttributes(feature.attributes() + [1])
             impacted_feature.setAttributes(feature.attributes() + [3])
 
+        # passing the number of affected population
+        # to new resulted(impacted or unaffected) features
+
+        unaffected_population_number = self.calculate_population_number(
```
timlinux: Please note our 'pull left' formatting rule so:

```
unaffected_population_number = self.calculate_population_number(
                                             unaffected_geometry,
                                             feature,
                                             area_population_attribute)
```

would rather be:


```
unaffected_population_number = self.calculate_population_number(
      unaffected_geometry,
      feature,
      area_population_attribute)
```

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -364,33 +400,55 @@ def assign_impact_level(
...
+                                            unaffected_geometry,
+                                            feature,
+                                            area_population_attribute)
+        unaffected_feature.setAttribute(
+            3,
+            unaffected_population_number)
+        impacted_population_number = self.calculate_population_number(
```
timlinux: pull left

safe/impact_functions/generic/classified_polygon_people/impact_function.py
```diff
@@ -429,3 +487,38 @@ def evaluate_affected_people(self):
         if total_affected_population == 0:
             message = no_population_impact_message(self.question)
             raise ZeroImpactException(message)
+
+    def calculate_population_number(self,
```
timlinux: pull left

safe/impact_functions/generic/classified_polygon_people/test/test_classified_polygon_people.py
```diff
@@ -55,6 +55,7 @@ def test_run(self):
         rect_extent = [
             extent.xMinimum(), extent.yMaximum(),
             extent.xMaximum(), extent.yMinimum()]
+        print rect_extent
```
timlinux: We should use 'prints' in production code (probably in non production code too since we have logging framework). Search InaSAFE code for 'LOGGER' for examples of how to use. Log messages will show in QGIS log panel.

safe/impact_reports/polygon_population_exposure_report_mixin.py
```diff
@@ -506,3 +607,96 @@ def total_population(self, total_population):
         :type total_population: int
         """
         self._total_population = total_population
+
+    def area_name(self,
```
timlinux: No need for line break if all args can fit in a single line of < 80 chars

safe/impact_reports/polygon_population_exposure_report_mixin.py
```diff
@@ -506,3 +607,96 @@ def total_population(self, total_population):
...
+
+        :returns area_name: name of the area
+        :rtype area_name: string
+        """
+
+        area_name = self.areas_names[area_id]
+
+        return area_name
+
+    def hazard_table(self,
```
timlinux: No need for line break if all args can fit in a single line of < 80 chars

timlinux:
@Samweli thanks for this patch. I have some mostly minor comments in the PR review above. If we can address these we can merge!

Samweli:
Thanks @timlinux, i will look into them :).

ismailsunni:
Thanks @Samweli 

@samnawi can you try this PR?

samnawi:
I've tried in my PC, both of Win7 64bit and Ubuntu 14.04 64bit.
I got some error message when I run vector population VS vector flood.
selection_047

@Samweli I've a question for the vector population keyword wizard, What is the meaning of "*Attribute for the type of the area*"? Is the geomerty type or population type or what?
selection_046

Thank you.

Samweli:
Hi @samnawi, the type attribute stands for the area type(ward or district or sub-ward) and from the exposure data you have used, is the population values type range? The IF expects integer type population values, this can be one of the reason it is raising an exception.

ismailsunni:
Ouch, write long comment, and it's gone. Accidentally click back button.

1. Is it right that the unit is `density`? Because in https://github.com/Samweli/inasafe/blob/develop/safe/impact_functions/generic/classified_polygon_people/impact_function.py#L513, there is multiplication between area ratio and the population. So, I think it should be `population count`.
2. Please don't use hardcoded attribute name like in https://github.com/Samweli/inasafe/blob/develop/safe/impact_functions/generic/classified_polygon_people/impact_function.py#L384. The best practice would be get the `field` attribute from hazard layer's keywords
3. Please use the similar approach that I use for classifying the hazard level. See https://github.com/inasafe/inasafe/pull/2546/files#diff-a53a88821aeba07646367921f6a9e767R112. Currently, you did like https://github.com/Samweli/inasafe/blob/develop/safe/impact_functions/generic/classified_polygon_people/impact_function.py#L402. It will be a problem if the layer has other attribute value, (e.g. HIGH LEVEL FLOOD)
4. Can you put the table with hazard level in top of the impact report, then followed by the total affected population, then the breakdown of each area.

ismailsunni:
Thanks @Samweli 
Point 1, 2, and 3 are solved.
But, point 4 missed the total affected, unaffected, and total population. See below image from another IF:
screen shot 2016-01-16 at 6 58 23 am
Compared to this IF
screen shot 2016-01-16 at 6 52 52 am
Perhaps you can add those three number. It should be easy, just refer to `population_exposure_report_mixin.py.`
Still in the same aspect, can you take the `name` from the `definitions.py` (not the `key`). See the report above, it says `High Hazard Zone`, `Medium Hazard Zone`. We use name since they are translatable, changeable, and more human friendly. In the other side, we can't change the `key` because it will break the keyword.

I will add another comment after I try with the data from @samnawi 

Thanks @Samweli

ismailsunni:
Hi @samnawi 
After I tried your data, the error is caused by `Null` value in the Population field. If I change it to a number, it works properly.
So, @Samweli, we need to add a handler or checker to see if the data is all a number, and give beautiful error message.

I added another comment on @timlinux 's comment above

safe/definitions.py
ismailsunni: Hi @timlinux if there are several Polygon, with the same Kelurahan Name, should we merge the result or we do it in aggregation? Currently, it will list all different id.

samnawi:
LGTM
image

But I still have some question about it :
1. This IF generate 3 class in the result. How to explain that class, meanwhile as I know, the hazard data in vector flood IF just inform wet and dry (2 class)? --> *see the result from building vs vector flood, it just generate inundated and not inundated*.
2. Should we show the breakdown (based on admin level) as default? because it will produce much pages in the report.

timlinux:
@ismailsunni wrote:

> Hi @timlinux if there are several Polygon, with the same Kelurahan Name, should we merge the result or we do it in aggregation? Currently, it will list all different id.

Hmm tough call. If we merge the result do you mean for the report only? Clicking on an individual polygon should still show the result for only that polygon. My feeling would be to merge the result in the IF for the report, as the aggregator might be working at a different level e.g. district or village.

Maybe also we could warn the user that there are duplicate area names....

ismailsunni:
Hi @samnawi 
> 1. This IF generate 3 class in the result. How to explain that class, meanwhile as I know, the hazard data in vector flood IF just inform wet and dry (2 class)? --> see the result from building vs vector flood, it just generate inundated and not inundated.

I think this IF is not for flood specific only. If we only have two classes (e.g. dry and wet), we need to map those two values in hazard's keyword via keyword wizard. For example: wet goes to high hazard zone, and we will left dry without class. Then, we will get the number of people in high hazard zone (which the same as wet zone).

> 2. Should we show the breakdown (based on admin level) as default? because it will produce much pages in the report.

That's what I asked @timlinux above. Currently, it breakdowns by the polygon. So, will it be better to have the breakdown is aggregated by the name of the area name?

ismailsunni:
After the last commit e913246, I think this PR is ready to be merged. What do you think @timlinux @samnawi ?

samnawi:
@ismailsunni Thank's for update this PR. I've been tried it and this is the result.
image
The symbology in the result was similar with the hazard classification. But for the color, it's correct if the inundated area (YES) is presented with green color? because in the previous we use red color for inundated / affected area.

I agree with your propose if this IF not only for flood, so we can replicate this method to the other hazard and it's good if the user can choose the proper hazard in the keyword, to create the classification.

Samweli:
@samnawi I incorrectly switched values that is why we are getting this, will correct them shortly.
Thanks

samnawi:
LGTM
selection_064

Maybe tomorrow @Charlotte-Morgan will add some comment about it.

Thank you.

Charlotte-Morgan:
Observations:
1. Problem with assigning KW to exposure data - see #2596
2. The IF does not cope with NULL values in the population data (spins out).
NULL should be an acceptable value and when counts are being reported can be 0 automatically.
A note should be added to the notes section of the impact report to say that NULL values in the population data have been replaced with 0.
3. there are inconsistencies with terminology between the map and report; I would prefer to see 'affected people' not 'affected populations' for the map title. Similarly the report heading could be "number of people affected'. Also check hazard level vs hazard zone
4 on the map the legend could read "people affected by each hazard zone" . The no and yes colours do not show on the map (need to turn off the hazard layer). This map is not very useful as it reflects the flooded an populated extents and not the analysis results. Might it be more useful to show counts of people affected??
5 I do not understand the general area of the impact report. The total number of people affected in the yes hazard zone is greater than the total of the affected and unaffected values. I think there is an error in the way the values are being added.
image
The value at yes should be the same as the affected population ??
Not sure where the no value is coming from.

I have not looked at the detailed report.

Great to see this IF progressing and able to work with data from different continents. I think the comments above are relatively small and easy to fix :)

timlinux:
@ismailsunni I'd like to get the release out this week. Can you fill in the gaps for @Samweli if he is not available starting tuesday? @Samweli and @ismailsunni it would be great to address @Charlotte-Morgan 's concerns above.

Charlotte-Morgan:
Hi @ismailsunni  - as requested more info on point 4 above.
a) the output layer is called "populations affected by each hazard zone" . This is then in the map legend. Please change the name of the default impact layer to "people affected by each hazard zone"
NOTE: this should be consistent with all hazard zone on people impact functions

b) the results in the output layer are displayed by the attribute safe_ag which is coded to mean yes or no. I think the results would be more meaningful if the label was the number of people in the hazard zone. No might be 0 and yes should be greater than 0.
image

c) The map title should also be "affected people" - with capitalisation as per standards.

image

Charlotte-Morgan:
re PR #2600 & point 5 above - results still look dodgy to me. We now have negative unaffected people ??  The total numbers also look incorrect to me.

image

The notes and assumptions also need to be updated to use the same standard wording and changes made by @timlinux  - i'm pretty sure we removed the human fractions statement previously

Samweli:
Hi @Charlotte-Morgan thanks for your comments, i'm looking at the problems you have highlighted. Will get back once i have fix them.

Charlotte-Morgan:
hi @ismailsunni @Samweli - I see that some of the issues above are addressed but the results are still confusing for me. I do not understand what you are trying to present with the results. I'm assuming that you are reporting on affected people. The number of people affected in the no  zone is still HUGE.

Suggestion: Would it be possible to style the top of the report in a similar way to the flood vector on raster population and just show the number of people in the flood affected area (yes hazard zone)
image
This solves the HUGE value at no :)
@Samweli - if this works, you may want to think about what you want to show on the map; maybe its the number of people in the flood affected (yes) area? not just where the flooded area is?

Another issue I just noticed: is this IF calculating the min needs for all affected people or only a certain percentage? ie in flood vector on raster population we use the 1% rule but we also state this in the impact report and notes.  
It would be useful to use similar notes if they are relevant:
image
@timlinux did a lot of work tidying up reports and notes & it would be great to get that work to persist in all impact reports (please).
Just leave out the second dot if you are not using a percentage. If there is a percentage state what it is.

Sorry - I think this gives you more things to tweak, but its getting there and it runs, even on data with NULL values so you have made great progress. :) And this IF will be so useful next week when it floods in Jakarta.

PR #2556, opened 2016-01-11
Refactoring signals and tr

Gustry:
This is smaller PR about #2528.
 - use tr() method from safe.utilities.i18n
 - refactors signals in  safe/common/signals.py
 - add docstrings

## CC
@timlinux @ismailsunni


ismailsunni:
LGTM

PR #2562, opened 2016-01-12
Comment out unaffected population from report.

ismailsunni:
For #2530


samnawi:
@ismailsunni This is the result
image

What is the different between *People within hazard field ("FLOODPRONE") of value "YES"* and *Total affected population*?

ismailsunni:
replaced by #2581

PR #2563, opened 2016-01-12
fix missing spaces

Gustry:


safe/defaults.py
```diff
@@ -203,7 +203,7 @@ def age_postprocessor():
         'Youth ratio defines what proportion of the population have not yet '
         'achieved financial independence. The age threshold for youth can '
         'vary by region - please consult with your local census bureau to find'
-        'out what the relevant threshold is in your region. InaSAFE does not '
+        ' out what the relevant threshold is in your region. InaSAFE does not '
```
ismailsunni: Hi @Gustry 
Perhaps, it will be better to make them uniform by putting the space in the end of line.

PR #2564, opened 2016-01-12
Fix #2468 : Set custom threshold in the building postprocessing

### LINKED ISSUE #2468: [Bug] Building type not appear in InaSAFE result
# Problem
My InaSAFE result didn't provide detailed building type eventhough i already select attribute that have building type inside it

# Description
I have several data from field survey. I already define all the data using keyword editor (you need to open qgis several time because it will crashed when you tried to define the second time - issue #2467 ).

For exposure data, i set 'amenity or type' for detailed building type.
image

When i run InaSAFE, i got 'unknown type' and 'other' not as i expected (breakdown by school, clinic, restaurant, place_of_worship, etc).

image

# Platform
InaSAFE 3.2.3 from QGIS repository
Windows 7
QGIS 2.8

# Data used
1. ancaman_banjir (flood hazard)
2. marunda_inasafe (exposure)
3. batas_rt (aggregation)

You can download these data in:
https://drive.google.com/folderview?id=0B3X6jqdE4HueSzFyT0oyc2E0aGc&usp=sharing

Gustry:
Fix issue #2468
If I understood, the mixin  concept in python, we need to "duplicate" code because the ```__init__``` in the ```BuildingExposureReportMixin``` can't be called in the IF.

Maybe, someone should review my Frenglish :)

qgis3

# CC
@ismailsunni @timlinux


safe/impact_reports/building_exposure_report_mixin.py
```diff
@@ -360,10 +361,9 @@ def total_buildings(self):
 
     def _consolidate_to_other(self):
         """Consolidate the small building usage groups < 25 to other."""
```
ismailsunni: Update the docstring also

safe/impact_functions/volcanic/volcano_polygon_building/impact_function.py
```diff
@@ -46,6 +46,9 @@ def __init__(self):
         self.volcano_names = tr('Not specified in data')
         self._target_field = 'Hazard'
 
+        # From BuildingExposureReportMixin
+        self.building_report_threshold = 25
```
ismailsunni: Why do you need to initialise it? Because in the `run` method, it will be replaced by the value from the postprocessor

ismailsunni:
I tried this PR, with standard test data EQ vs building, and I got:
```
Problem

KeyError
Suggestion

You can try the following to resolve the issue:

Check that you have the latest version of InaSAFE installed - you may have encountered a bug that is fixed in a subsequent release.
Check the InaSAFE documentation to see if you are trying to do something unsupported.
Report the problem using the issue tracker at https://github.com/AIFDR/inasafe/issues. Reporting an issue requires that you first create a free account at http://github.com. When you report the issue, please copy and paste the complete contents of this panel into the issue to ensure the best possible chance of getting your issue resolved.
Try contacting one of the InaSAFE development team by sending an email to info@inasafe.org. Please ensure that you copy and paste the complete contents of this panel into the email.
Details

An exception occurred when running the impact analysis.
impact_total
Diagnostics

In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe-dev/safe/utilities/analysis.py", line 774, in run_analysis self.run_aggregator()
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe-dev/safe/utilities/analysis.py", line 740, in run_aggregator self.run_post_processor()
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe-dev/safe/utilities/analysis.py", line 754, in run_post_processor self.postprocessor_manager.run()
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe-dev/safe/impact_statistics/postprocessor_manager.py", line 517, in run value.setup(parameters)
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe-dev/safe/postprocessors/building_type_postprocessor.py", line 84, in setup self.impact_total = params['impact_total']
```

But, it seems it works fine with other type of hazard.

safe/impact_functions/volcanic/volcano_polygon_building/impact_function.py
Gustry: Yes, it will be replaced by the value from the postprocessor, but don't we need to initialise every var which are in ```self``` in the ```__init__``` of the current class ?

ismailsunni: Ah yes, I usually, use `self.var = None`. Ok, no problem then.

timlinux:
In the gui next to the spin box, ``feature`` -> ``features``

and ``which are inferior than this value`` to ``which are within this threshold``

Gustry:
Thanks Tim for the english.
For the plural, I put the plural form, but it doesn't seem to work : https://github.com/Gustry/inasafe/blob/fix-2468/safe/defaults.py#L321

safe/impact_functions/volcanic/volcano_polygon_building/impact_function.py
ismailsunni: After some thought, perhaps we can put the initialisation in `building_exposure_report_mixin.py`.

Gustry: In case of a mixin, python will look for the first ```__init__``` method from the left to right. For instance, in our code : 
```
class VolcanoPolygonBuildingFunction(
        ClassifiedVHClassifiedVE,
        BuildingExposureReportMixin):
    """Impact Function for Volcano Point on Building."""
```
Only the ```__init__``` in ```ClassifiedVHClassifiedVE``` will be executed. That's why I had to duplicate that part of code in every IF. But it's true that this value is going to be overwrite. What should I do ? :)

Gustry:
I have fixed the comment above.
However, I have disabled this custom parameter threshold in ```earthquake_building``` because this IF is using a special postprocessor, not the ```building_type_postprocessor``` and has a special ```statistics_type```. Is-it OK like this or do you want me to add the threshold also to this IF ?

safe/impact_functions/volcanic/volcano_polygon_building/impact_function.py
ismailsunni: Ah, yes. I read this http://python-history.blogspot.co.id/2010/06/method-resolution-order.html after reading your comment. Hmm, I am ok with the initialisation. :)

ismailsunni:
@Gustry is it possible to add the threshold? If so, perhaps we need to add the threshold also, to make them uniform.
Actually, I am wondering what this post processor does, and what's the different with building type postprocessor. It's only used once also.

samnawi:
@Gustry Please correct if I'm wrong. For this PR, Is it right if I set the threshold, all of the value under the threshold will be labeled as **other**?

I got the result when I set the threshold in 25 (default) like this :
selection_055

Then, when I tried to set the threshold in 500, not all of value under 500 become to other.
selection_057

Is it because the threshold also applied to *total* and *not affected* too?

Gustry:
@samnawi Yes, the threshold is applied to the ```Total``` column only. In your second screenshot, you don't have any value less than 500.

Gustry:
I've updated the earthquake IF to use the threshold. I needed to change some other postprocessing statistics. It doesn't seem to break something, but the impact report has now one new item : ```closed buildings``` as you can see on the screenshot.
I'm not enough familiar with the statistics output in InaSAFE. Did my last commit could break something else ?
Before :
qgis3
After :
qgis3

ismailsunni:
I don't remember anything about the statistic type or statistic classes. Perhaps we need to ask @timlinux or another older people in InaSAFE.
Anyway, this PR is LGTM.

timlinux:
LGTM too  - I think we should remove that statistics_type logic in the underlying aggregator code too since this change will mean it is not used anywhere.

timlinux:
Oh please just check the detailed PDF reports are equivalent before merging

ismailsunni:
@timlinux, great, we can remove it from xml metadata attributes also then.

Gustry:
Both PDF are the same, except with this PR we get more details informations in the second table ``Notes and assumptions``. It's better !

To remove the ```statistics_type```, I've created another pull request. We will merge #2593 after the release, but we need to merge this one before.

PR #2566, opened 2016-01-12
Manage layers in analysis.py

Gustry:
- Rename layers like in base.py
 - Clip_parameters returned useless parameters (exposure, hazard and geoextent), so I've removed them.
 - Use self.hazard, self.exposure and self.aggregation everywhere in analysis.py


ismailsunni:
LGTM

Issue #2529, opened 2015-12-22
Flood IF Analysis Layer Result did not Match with Hazard and Exposure Data [bug]


harrymahar:
# Problem
Flood IF Analysis Layer Result did not Match with Hazard and Exposure Data

# Detail 
I try to run flood IF using flood Jakarta (vector) and Java Population (raster). The result generally shows good result but when I zoom in to certain area some layer is missing. There is a hole between flood layer and the result layer. It should be at least be classified in analysis layer and overlap with the flood layer but unfortunately it is empty.

# ScreenShoot
![output layer flood_population2](https://cloud.githubusercontent.com/assets/2534179/11947246/e0e6450e-a896-11e5-8c4d-69572e50a827.PNG)

![output layer flood_population](https://cloud.githubusercontent.com/assets/2534179/11947250/e665a63c-a896-11e5-8466-f0def87839d8.PNG)

# Platform
Windows 8.1 64 Bit
RAM 4 GB
QGIS 2.8.4 Wien
InaSAFE 3.2.5 dev-master

# Data
Hazard : Jakarta_Flood_18113__WGS84.shp (Vector)

Exposure : 
*Java_Population_WGS84.tiff (Raster)

Agregation : Jakarta_District_Boundary.shp	

Cc : @timlinux @ismailsunni @samnawi

samnawi:
I think it's because the symbology. If we look at the picture in the bottom, the hole in the result has a value < 1. 
![selection_004](https://cloud.githubusercontent.com/assets/8871958/12033306/1a9c1086-ae54-11e5-9cf2-5e784209f0a6.png)

When I try to change the symbology, the hole area filled by color also.
![selection_005](https://cloud.githubusercontent.com/assets/8871958/12033321/4b47fcea-ae54-11e5-963f-ed9ac4d6b39a.png)

# Problem
In the symbology, value 0 - 1 will represented with white color, but in the display the value < 1 represented by transparent color.

ismailsunni:
@samnawi do you have a suggestion for a better symbology?

samnawi:
@ismailsunni I just want to make sure if the symbology in the legend is white, there should be no hole / transparent symbol in the map.

samnawi:
@ismailsunni has been created the PR to solve this problem.

@harrymahar From your perspective, what are your proposed to solve it?
[...]

PR #2567, opened 2016-01-13
Set transparency to 0 for all style per #2529

ismailsunni:
For #2529


samnawi:
mmm.... why all of the unaffected area become to white color?
image

Charlotte-Morgan:
I don't think this is a good solution because we need to be able to draw the raster over other layers and  'see through' the areas with 0

ismailsunni:
@Charlotte-Morgan @samnawi 
So, we only set transparent = 100 for 0 value in the impact layer? And, if the value is not 0, we set transparent = 0

Charlotte-Morgan:
i have another solution which i put in the ticket #2529

PR #2574, opened 2016-01-13
Use SafeLayer in Analysis

Gustry:
- Use SafeLayer in Analysis
- Remove keywords when we setup the analysis as the keywords will be read by SafeLayer : 
 - command line
 - dock
 - analysis_handler

### DIFF
--- a/safe/storage/safe_layer.py
+++ b/safe/storage/safe_layer.py
@@ -94,8 +94,8 @@ class SafeLayer(object):
     def keywords(self, keywords):
         """Setter for keywords property.
 
-        :param keywords: The actual layer.
-        :type keywords: QgsMapLayer, Layer
+        :param keywords: Keywords as a dictionary.
+        :type keywords: dict
         """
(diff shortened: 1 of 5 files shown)


Gustry:
@ismailsunni This PR is now ready to review.
Thanks in advance.

safe/storage/safe_layer.py
```diff
@@ -94,8 +94,8 @@ def keywords(self):
     def keywords(self, keywords):
         """Setter for keywords property.
 
-        :param keywords: The actual layer.
-        :type keywords: QgsMapLayer, Layer
+        :param keywords: Keywords as a dictionary.
+        :type keywords: dict
```
Gustry: I think it was a mistake in the docstring, isn't it ?

ismailsunni: Hmm, yes. It seems a bad copy-paste from other method. Thanks :)

ismailsunni:
LGTM

PR #2580, opened 2016-01-15
Set transparent for raster value = 0. Per #2529.

### LINKED ISSUE #2529: Flood IF Analysis Layer Result did not Match with Hazard and Exposure Data [bug]

ismailsunni:
Fix #2529. It will replace my PR in #2567

### DIFF
--- a/safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
+++ b/safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
@@ -367,11 +367,7 @@ class ITBFatalityFunction(
             style_class = dict()
             style_class['label'] = create_label(interval_classes[i])
             style_class['quantity'] = classes[i]
-            if i == 0:
-                transparency = 100
-            else:
-                transparency = 30
-            style_class['transparency'] = transparency
+            style_class['transparency'] = 30
             style_class['colour'] = colours[i]
             style_classes.append(style_class)
 
--- a/safe/impact_functions/generic/classified_polygon_population/impact_function.py
+++ b/safe/impact_functions/generic/classified_polygon_population/impact_function.py
@@ -216,15 +216,10 @@ class ClassifiedPolygonHazardPopulationFunction(
             else:
                 label = create_label(interval_classes[i])
 
-            if i == 0:
-                transparency = 100
-            else:
-                transparency = 0
-
             style_class['label'] = label
             style_class['quantity'] = classes[i]
             style_class['colour'] = colours[i]
-            style_class['transparency'] = transparency
+            style_class['transparency'] = 0
             style_classes.append(style_class)
 
         # Override style info with new classes and name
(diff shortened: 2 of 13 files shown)


samnawi:
As the proposed from @Charlotte-Morgan in the issue #2529 
image

I still find the cell with a value greater than 0 symbolized by the transparent color.
image

timlinux:
LGTM

ismailsunni:
Hi @samnawi 
Ah, yes. I accidentally revert my changes. I have updated it again. Please check :)

samnawi:
@ismailsunni this is the result 
**Still in Develop Branch**
image

**When I tried your PR**
image

As the proposed from @Charlotte-Morgan, the first point is LGTM.
But for 2nd point, the 0 value should presented by transparent.

Then, I try to change the properties
image

The map will show like this 
image

Charlotte-Morgan:
I agree with @samnawi - great explanation thank you!

ismailsunni:
Hi @samnawi @Charlotte-Morgan 
Yes, that's what I did. My code works in qgis python script https://github.com/ismailsunni/scripts/blob/master/qgis_raster_transparency.py
but apparently, it doesn't work when I put it in InaSAFE. Need to find out. Thanks for the feedback :)

ismailsunni:
Hmm, now I understand, the transparent is indeed not working in InaSAFE.

samnawi:
@ismailsunni The transparent that you mean cannot work in InaSAFE is just for value = 0 ?
Because in the current, The transparent can work for 0 - 1 cell value.
selection_066

ismailsunni:
Did you set the transparency from layer properties or doing nothing (from InaSAFE) ?

samnawi:
The picture in above is properties come from InaSAFE result

ismailsunni:
Update,
After some discussion (and confusion) with @samnawi, we found that there is a problem in InaSAFE in OSX. In OSX, the transparency set doesn't work properly. But it works on Windows and Ubuntu.

I will make it work in Ubuntu first, then try to fix it on OSX.

ismailsunni:
Result comparison Ubuntu vs OSX (ignore the number, since it's different analysis extent ):

1. OSX (Not transparent)
osx


1. Ubuntu 14.04 (Transparent)
ubuntu-14 04

Both are using develop, and the same scenario


@lucernae @timlinux can you try the scenario also in OSX? to make sure, it's not me.

samnawi:
For Transparent issue :+1: 
LGTM in my Ubuntu and Windows machine.

PR #2582, opened 2016-01-15
move clip_parameters function in IF

Gustry:
Moving the clip_parameters function into the IF.

I created some getters and setters in analysis.py. I missed some docstring as it should be very temporary in analysis.py. It's only to make a kind of bridge between the dock and the IF until we remove analyse.py.

As @ismailsunni told me, I had to change the DOCK in the test.


safe/impact_functions/base.py
```diff
@@ -591,6 +629,7 @@ def set_if_provenance(self):
         )
 
     def message_pre_run(self):
```
timlinux: ``message_pre_run`` -> ``emit_pre_run_message``

safe/impact_functions/base.py
```diff
@@ -652,3 +692,197 @@ def message_pre_run(self):
             # KeyError is for when ['postprocessors'] is unavailable
             pass
         send_static_message(self, message)
+
+    def get_clip_parameters(self):
```
timlinux: ``get_clip_parameters`` -> ``clip_parameters`` (generally I think it is better to avoid using ``get_`` prefixes for methods as it is verbose and usually does not add meaning).

timlinux:
Thanks @Gustry the analysis refactor is coming on nicely, LGTM. I made a few small (pedantic) comments.

Gustry:
This PR is ready to be reviewed @timlinux and @ismailsunni.
Thanks !

safe/impact_functions/base.py
```diff
@@ -208,6 +234,26 @@ def actual_extent_crs(self):
         return self._actual_extent_crs
 
     @property
+    def map_canvas(self):
```
ismailsunni: Is it ok to put map_canvas in impact function base? Please correct me, but in my mind, map_canvas is more related to the QGIS not the impact function it self.

Gustry: Yes, I have though the same too. But I need to extent of the map canvas in order to compute the best extent between exposure, hazard, user extent and the map canvas extent.
To be more consistent with other properties like ```requested_extent``` and ```actual_extent```, I can use ```map_canvas_extent```. Is-it ok ?

ismailsunni: > To be more consistent with other properties like requested_extent and actual_extent, I can use map_canvas_extent. Is-it ok ?

Do you mean that we will only store the extent not the whole map canvas object? If so, I agree with it :)

Gustry: It's done ;-)

ismailsunni:
LGTM

PR #2584, opened 2016-01-20
move optimal_clip into the IF

Gustry:


Gustry:
As the feature freeze ended, we can merge into develop. Is-it possible to have a look @ismailsunni or @timlinux ?
Thanks

ismailsunni:
LGTM

Gustry:
Thanks for the review @ismailsunni

PR #2590, opened 2016-01-22
Fix 2235 huge difference

ismailsunni:
Please check this @samnawi @adissadis


samnawi:
LGTM.
Both of Ubuntu and WIndows PC.

Issue #2410, opened 2015-09-22
BUG: Flood raster on building - error in report


Charlotte-Morgan:
# problem
The total number of  affected items listed in the report actions is incorrect - it is in fact the total number of items in the analysis area.

eg
Jakarta flood raster on buildings - whole analysis area
action list mentions 2016 closed schools and 128 closed hospitals
but this is the total number of schools and hospitals not just closed (affected)
![image](https://cloud.githubusercontent.com/assets/5743135/10017573/db4b4b0a-6159-11e5-9a14-a4f8a5992386.png)

The same issue occurs with the value quoted in the action list for flood raster on people

# proposed solution
Action lists should quote values for affected features not total analysed features

Charlotte-Morgan:
another proposed solution is to not quote the affected numbers in the action list.
see #2389 and apply same solution there too

PR #2591, opened 2016-01-22
Better building usage count. Add categories as parameter.

ismailsunni:
Addressing #2410 
Result:
screen shot 2016-01-22 at 3 06 30 pm
See that closed hospital and closed school is equals to the total affected, not the total building.

Please check, @Charlotte-Morgan or @samnawi


samnawi:
LGTM
image

Issue #2594, opened 2016-01-24
Now it is hard to see error log in Travis


lucernae:
The log is too long to be displayed in Travis.
Perhaps something happen with the verbosity of the log?

Gustry:
@timlinux Can we do something on the QGIS side ?
@lucernae We can still use the raw log, it works.

ismailsunni:
Fix in #2597

PR #2597, opened 2016-01-28
Standardise style version in all qmls

timlinux:
Also updates changelog. Don't merge yet

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,494 @@
+Changelog for version 3.3.0
+================================
+
+759fde8 (HEAD -> develop, upstream/develop, origin/develop, origin/HEAD) Merge pull request #2595 from ismailsunni/fix_failing_test
+8485a67 PEP8
+51bc9d8 Fix failed test.
+42b4463 Merge pull request #2564 from Gustry/fix-2468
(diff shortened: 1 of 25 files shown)


timlinux:
@ismailsunni can you review - I basically updated all QML's so we get rid of all those nasty messages clogging up the stdout (about outdated styles).

ismailsunni:
Ah, shorter stdout.... Now, I can see some message, e.g. https://travis-ci.org/inasafe/inasafe/builds/105497288#L1693.

And of course, LGTM.

Related issue : #2594

timlinux:
For the record the impetus behind this PR is that the QML version causes lots of debug messages in QGIS if it is older than the current version - so much so that our tests exceed the 10000 line maximum for travis logs and it is basically impossible to see useful information between all the log 'spam'. Here is the before example:

selection_002

And here are two pics with my changes applied:

selection_001


selection_003

timlinux:
I also fixed it upstream in QGIS: https://github.com/qgis/QGIS/commit/e15c7fb96cf25e3f716b28eebd48a9719642c2a4

https://github.com/qgis/QGIS/pull/2734

PR #2608, opened 2016-02-04
fix #2588 Remove the network request if we get the data to avoid the timeout

### LINKED ISSUE #2588: QgsNetworkManager shows useless warning message
# Problem
Since InaSAFE is using QgsNetworkManager in the OSM downloader, even if the user got the shapefile correctly, QGIS shows some warnings like this after a few seconds : 
qgis3
You can change the timeout in the QGIS settings : 
selection_001

I tried to have a look in the QGIS code source to understand : 
http://qgis.org/api/qgsnetworkaccessmanager_8cpp_source.html line 203, we can see a timer is starting and it will show a message bar if the timer is finished. But I can't see where the request should stop the timer.

# CC
@timlinux, do you think it's a QGIS problem or did we miss something ?
We are not the only one with this problem : https://github.com/nextgis/osminfo/issues/17

Gustry:


Gustry:
This fix can work only for QGIS > 2.11

samnawi:
LGTM in QGIS 2.12

PR #2611, opened 2016-02-05
Fix 2609 target field

### LINKED ISSUE #2609: BUG: aggregation error with tsunami impact on buildings
# problem
When running tsunami impact on buildings i get an aggregation error.

# environment
windows
QGIS 2.8.3
InaSAFE 3.3.0 dev

# data
Tsunami continuous raster
OSM buildings with type attribute
key words on  buildings layer (these were auto updated from 3.2)
image
no aggregation data

note: these are the standard maumere scenario training data. the data contain many records with NULL for the type attribute. However there are some other values.

# error message in dock
image

# proposed solution
IF should run and generate results and users should be happy :)

ismailsunni:
- Fix #2609 
- PEP8

### DIFF
--- a/bin/inasafe.py
+++ b/bin/inasafe.py
@@ -325,7 +325,8 @@ def run_impact_function(command_line_arguments):
     impact_layer = analysis.impact_layer
     write_results(command_line_arguments, impact_layer)
     try:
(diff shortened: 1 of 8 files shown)


Charlotte-Morgan:
LGTM - this fixes the bug.

Issue #2612, opened 2016-02-05
Tsunami impact on buildings: add values to map legend


Charlotte-Morgan:
The tsuanmi impact on buildings has great notes explaining the different classes. There values are dynamic as users can change the class thresholds.

Is it possible to also include these value on the map legend :+1: 
Dry zone: 0 m
low hazard zone: 0.1 - 1.0 m
medium hazard zone: 1.1 - 3.0 m
high hazard zone: 3.1 - 8.0 m
very high hazard zone > 8.0 m

Issue #2610, opened 2016-02-05
PAPERCUT: minimum needs wording


Charlotte-Morgan:
# problem
the current minimum need statement : "relief items to be provided single" sounds strange

# proposed solution
would it be possible to replace the word 'single' with the word 'once' ?
This would make the statement in reports more relevant

ismailsunni:
The word is editable by user. But, for convenience, I have update the default value in my PR #2614

PR #2614, opened 2016-02-05
Fix 2612 and 2610

ismailsunni:
For #2612 and #2610

### DIFF
--- a/resources/minimum_needs/BNPB_en.json
+++ b/resources/minimum_needs/BNPB_en.json
@@ -52,7 +52,7 @@
             "Default": "0.05",
             "Minimum allowed": "0.02",
             "Maximum allowed": "0.05",
-            "Frequency": "single",
+            "Frequency": "once",
             "Resource name": "Toilets",
             "Resource description": "",
             "Unit": "",
--- a/resources/minimum_needs/BNPB_id.json
+++ b/resources/minimum_needs/BNPB_id.json
@@ -52,7 +52,7 @@
             "Default": "0.05",
             "Minimum allowed": "0.02",
             "Maximum allowed": "0.05",
-            "Frequency": "tunggal",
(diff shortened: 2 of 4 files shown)


ismailsunni:
Sample result of the new legend:
screen shot 2016-02-05 at 3 02 00 pm

samnawi:
LGTM

Charlotte-Morgan:
thanks @ismailsunni - this looks better, normally we use mutually exclusive values but I will wait for someone else to ask for that :)

ismailsunni:
Hi @Charlotte-Morgan 
Yes, my math-sense also thinks they should be mutually exclusive. In the calculation, they are indeed mutually exclusive. So, I prefer something like this:

Dry Zone : depth = 0 m
Low Hazard Zone : 0 < depth <= 1.0
Medium Hazard Zone : 1.0 < depth <= 3.0
High Hazard Zone : 3.0 < depth <= 8.0
Very High Hazard Zone : 8.0 < depth

Charlotte-Morgan:
have a look at #2612 - I give specific values in that.
Or in your example above; IIRC, the bottom of the range is normally greater than the top of the previous range.

so if low goes up to 1.0
medium will be greater than 1.0 up to 3.0
high will be greater than 3.0 up to 8.0 
etc

ismailsunni:
In the example #2612, there is no legend for 0.05. So, it doesn't cover all values.

Do we need to update the legend now?

Charlotte-Morgan:
i think you need to tweak it as per my comment above

ismailsunni:
Hi @Charlotte-Morgan 
Sorry, I think I miss-understood your comment. So, I made another one:
screen shot 2016-02-09 at 8 37 30 am
see my PR on #2622

Issue #2603, opened 2016-02-02
Update Metadata XML Template


ismailsunni:
Change from `English` to `eng`.
See here : https://github.com/inasafe/inasafe/blame/develop/safe/metadata/iso_19115_template.xml#L7

ismailsunni:
Today I learned that we need to add new line to text file, not only in code. http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline. Therefore, I added it.

ismailsunni:
Addressed in #2616, closing this now...

PR #2616, opened 2016-02-05
Update LanguageCode from English to eng

ismailsunni:
The command that I use: `find . -type f -name '*.xml' -exec sed -i '' s/English/eng/ {} +`

For issue #2603 , please check @lucernae


Gustry:
LGTM

PR #2620, opened 2016-02-08
update keywords for the OSM tile layer

Gustry:
If we use the OSM tile layer, InaSAFE updates its keywords every time (XML).
I need to use ```git checkout resources/osm/WorldOSM.xml```

@ismailsunni Is it OK if we update the XML file ?


ismailsunni:
Yes, i think it's ok. It seems the previous one is from the old scheme of xml metadata (still use CDATA).

And, we also need to make sure, whenever we run a test, it should not modify the test data. If so, there must be something wrong.

Thanks @Gustry 
It seems there is a conflict, please fix it and you can merge it.

PR #2621, opened 2016-02-08
fix #2604 check if the user_extent is well defined

### LINKED ISSUE #2604: BUG: User extents error message
# Problem
When I choose the analysis extent as the intersection of the hazard and exposure layers (Flood vector and population vector); i get the warning that there are no overlapping extents.

image

However, this warning is not correct, the data do have overlapping extents.

# proposed solution
Is it possible that the extent checker --> warning issuer needs to be tweaked to recognise vector on vector analysis??

Gustry:
Fix https://github.com/inasafe/inasafe/issues/2604

### DIFF
--- a/safe/utilities/analysis.py
+++ b/safe/utilities/analysis.py
@@ -18,7 +18,7 @@ import logging
 # noinspection PyPackageRequirements
 from PyQt4.QtCore import QSettings
 
-from qgis.core import QgsMapLayer
+from qgis.core import QgsMapLayer, QgsRectangle
 
 from safe.impact_statistics.postprocessor_manager import (
     PostprocessorManager)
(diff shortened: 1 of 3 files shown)


safe/utilities/analysis.py
```diff
@@ -125,12 +125,16 @@ def user_extent(self):
...
-            user_extent.yMaximum()]
-        self.impact_function.requested_extent = extent
+        if isinstance(user_extent, QgsRectangle):
+            extent = [
+                user_extent.xMinimum(),
+                user_extent.yMinimum(),
+                user_extent.xMaximum(),
+                user_extent.yMaximum()]
+            self.impact_function.requested_extent = extent
+        else:
```
ismailsunni: Will this ever happen?
And if so, is the exception handled properly, and beautifully?

Gustry: It should never happen in the current state of InaSAFE.
The check is done before everytime. I put this exception **in case of** this function doesn't get a `QgsRectangle` or `None`. As I said this `analysis.py` will be removed soon.

timlinux:
LGTM thanks @Gustry

PR #2622, opened 2016-02-09
Fix legend to use greater number than preious range.

ismailsunni:
From discussion in PR #2614 
screen shot 2016-02-09 at 8 37 30 am


timlinux:
LGTM thanks @ismailsunni

PR #2640, opened 2016-02-23
Fix issue with styles for flood on polygon pop when class field is an…

timlinux:
… int


ismailsunni:
LGTM.
One thing though, look at the result of find usage in PyCharm, it seems impact_functions/styles.py is not used anywhere else except in test_styles. Should we just remove it?

timlinux:
@ismailsunni that would be good - death to dead code! Hey that doesn't make logical sense :-)

PR #2649, opened 2016-03-01
use staging osm-reporter if developer mode is ON and if it is not a f…

Gustry:
use staging osm-reporter if developer mode is ON and if it is not a final release

@timlinux Is-it ok to use K5 in InaSAFE ?


ismailsunni:
LGTM

timlinux:
LGTM too

> @timlinux Is-it ok to use K5 in InaSAFE ?

Yes sure!

PR #2661, opened 2016-03-08
Refactoring impact function run

Gustry:
This PR deletes the file `analysis.py` and makes the IF working by itself.

Client classes can use IF like in the dock : 
```
impact_function = self.prepare_impact_function()
impact_function.run_analysis()
```

CC : @timlinux @ismailsunni @lucernae


ismailsunni:
LGTM

PR #2668, opened 2016-03-11
fix #2361 count roads length in post processing instead of features

### LINKED ISSUE #2361: BUG : Different value at the report of roads inundation (Flood and Tsunami)
# Problem
There is a different value that appear in the report, between in the "Breakdown by road type" and in the "Detailed road type report".
selection_197

# Data
Hazard : Tsunami Maumere
               Jakarta Flood Polygon (RW Boundary)
               Jakarta Flood Raster (HKV - Modeling)
Exposure : OSM Roads

# Proposed Solution
Make it happy

# CC
@Charlotte-Morgan @timlinux 

Thank you

Gustry:


samnawi:
LGTM
selection_097
But @Gustry , should we change the column name of temporarily closed (in the closed roads) with total affected?

Gustry:
Thanks for the review.
It makes sense `total affected`, which one is better ? I was thinking about that comment : https://github.com/inasafe/inasafe/issues/2361#issuecomment-170837103 when I put `Temporarily closed`

Gustry:
@samnawi Should I let like this or do you want me to change ?
As I need to backport it to 3.3.1 after.

PR #2680, opened 2016-03-14
Tsunami on roads IF

Gustry:
PR against master, I will backport it to develop later.

Work in progress


Charlotte-Morgan:
Hi @Gustry - great progress thank you :)
A few comments:
we use the word inundation for tsunami impact and flooded for flood impacts (changed at 3.2)
the road IF runs slowly, even on a small area
In the road report, can you please use the word inundated and make the report similar to the tsunami impact on buildings. The current road report looks like this:
image

However for buildings affected by tsunami we report against each class rather than affected and not affected - see like this:
image

Is it possible to make the road report similar to the building report for Tsunami?

Also - in the wizard and dock, the user still has the option to use the flood impact on tsunami impact function. It would be good when you are done that they only have the tsunami impact function available.

Apologies if i have commented on things you are already working on or aware of.

Charlotte-Morgan:
hi @Gustry - i've reviewed the latest push with impact report by hazard zone.
This is looking good. do you think we need 'unaffected' as this is not included in the building report and should be the dry zone within the extent of the hazard data.

We probably do not need to report on all the roads outside the hazard extent / analysis extent. This may also make it a bit faster?

We have also introduced the idea that if a table has all 0 values we omit that item. I think Christian did the work on these table tweaks previously.

Charlotte-Morgan:
@samnawi will provide some advice on road classes and the order they are listed tomorrow

Gustry:
> We have also introduced the idea that if a table has all 0 values we omit that item.

It's one of the thing to review :) I think you are talking about the `unaffected` column, right ?
Which class do you want considered as unaffected ? dry and low hazard zone ? (If I understood the code with @ismailsunni )

> we use the word inundation for tsunami impact and flooded for flood impacts (changed at 3.2). In the road report, can you please use the word inundated and make the report similar to the tsunami impact on buildings. 

This should be fine now.

>the road IF runs slowly, even on a small area

It's on my todo list, I think I know why.

> Also - in the wizard and dock, the user still has the option to use the flood impact on tsunami impact function. It would be good when you are done that they only have the tsunami impact function available.

Yes, I would like to do it at the end. I can compare both results in that way.

> I see the notes and assumptions need some tweaking

I didn't touch them. They come from flood raster on buildings IF I think.

Thanks for the review.

Charlotte-Morgan:
hi @Gustry 
Thanks for the update. its looking good. I make a few more comments - though most I think can be considered bug fixes for after the release if necessary.

In this case, my understanding is that Unaffected = Open = Dry zone. There is no need to report unaffected in this case.In other IF - we have asked for an unaffected column. That is because the 'dry zone" equivalent does not exit.  I checked with @samnawi - we agree on this point.

Still is slow - I understand why - some nice warnings in the future and maybe something in the change log for now will cover this. I also have some ideas on making the analysis faster that we can discuss when you are here.

Thanks for removing flood IF option in dock and wizard +1

As for tweaking notes and assumptions - sorry, was not implying you had changed anything. We can do that sometime. It will affect all road IF probably 

Only other point is the mismatch between closed road reporting - not sure what unit this is in

I think that is all - this looks good to me - thank you

Gustry:
Good morning @Charlotte-Morgan,
> There is no need to report unaffected in this case.

Ok, I can remove that column.

Yes, I know it's slow. I tried to implement this IF for the release soon. We can improve this IF, but as we have multiple thresholds, we have to expect it to be slower than the Flood on Road IF. We plan to add a progress bar and add some messages to inform the user that the process is running.

> Only other point is the mismatch between closed road reporting - not sure what unit this is in
I think, I need to merge this PR and backport it to the master branch : https://github.com/inasafe/inasafe/pull/2668
Is-it possible to have a look ?

Thanks for the review

timlinux:
LGTM Gustry. I'd still like to see the aggregation in m and some other small tweaks to the report (I'll make a separate ticket for those)

timlinux:
Please also backport to master

Gustry:
@timlinux : 

> I'd still like to see the aggregation in m

It should be ok with that PR : https://github.com/inasafe/inasafe/pull/2688

> Please also backport to master

It's already a PR against the master branch. I will have to "frontport" it to develop :)

Gustry:
selection_060

I have fixed the unit for each row and removed the unit on headers.

selection_061
I've merged master in the branch so as to have my fix about road length in the post-processing, but it wrong on that case.

timlinux:
@Gustry ok awesome work thanks. LGTM

PR #2685, opened 2016-03-17
Remove not slow filter in travis.

ismailsunni:


safe/engine/test/test_engine.py
```diff
@@ -1550,8 +1544,7 @@ def test_polygon_to_roads_interpolation_flood_example(self):
         # assert I_attributes[]['polygon_id'] ==
         # assert I_attributes[]['parent_line_id'] ==
 
-    test_polygon_to_roads_interpolation_flood_example.slow = True
-
+    @unittest.skip('Skipped by default')
     def Xtest_polygon_to_roads_interpolation_jakarta_flood_example1(self):
```
Gustry: remove X

safe/engine/test/test_engine.py
```diff
@@ -165,6 +165,7 @@ def test_polygon_hazard_raster_exposure_clipped_grids(self):
                    'and full grids, Got %s and %s' % (x, y))
             assert numpy.allclose(x, y)
 
+    @unittest.skip('Slow test')
```
Gustry: Instead of always skipping, can we use a condition ? `unittest.skipIf(condition, reason)`
For the condition, we can use a environment variable or a QSettings (depending if are on travis or in QGIS UI)

Gustry:
I think it might be good to merge this PR asap, it will be useful for the new testing in docker.

Gustry:
Thanks Ismail, LGTM.
I don't think we need to wait 3.6. This PR doesn't change anything. We can have another round with slow tests after the docker PR.

.travis.yml
```diff
@@ -10,6 +10,9 @@ python:
 virtualenv:
   system_site_packages: true
 
+env:
+- ON_TRAVIS=true
```
Gustry: Hum, did we enable slow test on travis ? :)

Issue #2691, opened 2016-03-17
Improvements to Tsunami impact on roads report


timlinux:
### Problem

Tsunami on roads still has a few issues in its reporting:

![selection_006](https://cloud.githubusercontent.com/assets/178003/13848411/56487cde-ec5a-11e5-96fa-be02d0ac1629.png)

![selection_007](https://cloud.githubusercontent.com/assets/178003/13848413/590b555e-ec5a-11e5-9d4a-555bac885d2f.png)
### Expected Behaviour
- Units should be explicit
- Table headings should span all columns
- Aggregation column headings should be in bold
- Aggregation report should be in (m) not number of features
### CC

@Charlotte-Morgan

timlinux:
See also #2680

timlinux:
For the first image above the (m) might be better asa suffix to column headers rather than row headers.

Gustry:
See comment : https://github.com/inasafe/inasafe/pull/2680#issuecomment-198193826
Some fixes are in master : https://github.com/inasafe/inasafe/commit/56a7ef1965de2f63dced204a19b297e34c50280e

Need to frontport to develop

Gustry:
Good evening @Charlotte-Morgan and @timlinux 
How would you like to highlight some columns in the table ? (inundated roads)

Gustry:
So everything has been fixed, except the highlight.

timlinux:
Discussed with Etienne - we will wait for Qt table implementation for dock reporting for doing things like highlighting the flooded columns.

ismailsunni:
I think we will move it to 3.5 @timlinux @Gustry ?
For the Qt Table

Gustry:
We will not do Qt Table anymore. We just need to highlight columns which are affected in the HTML.

timlinux:
Also make group headings span multiple columns:  e.g. "Breakdown by road type"

Charlotte-Morgan:
Here is my attempt at what a table should look like. 
Title with units at the top.
Affected and not affected clearly separated with headings and colour if possible.
Notes further explain things.

If it's not possible to highlight things now - that's fine; the use of headings can solve that.

![image](https://cloud.githubusercontent.com/assets/5743135/16683061/10b113bc-4528-11e6-8e0a-00ab2013eb9a.png)

timlinux:
I moved over to 3.6 - we can add column highlighting then. @Gustry has already made various other tweaks there.

[...]

PR #2694, opened 2016-03-21
Fix unit in the road report issue #2691

Gustry:
#2691

### DIFF
--- a/safe/impact_statistics/postprocessor_manager.py
+++ b/safe/impact_statistics/postprocessor_manager.py
@@ -192,6 +192,9 @@ class PostprocessorManager(QtCore.QObject):
                     # run method. But just in case there is something wrong.
                     zone_name = 'Unnamed Area %s' % null_index
                     null_index += 1
+                if name == 'road type':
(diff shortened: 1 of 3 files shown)


safe/impact_statistics/postprocessor_manager.py
```diff
@@ -192,6 +192,9 @@ def _generate_tables(self, aoi_mode=True):
                     # run method. But just in case there is something wrong.
                     zone_name = 'Unnamed Area %s' % null_index
                     null_index += 1
+                if name == 'road type':
+                    # We add the unit 'meter' as we are counting roads.
+                    zone_name = '%s (m)' % zone_name
```
timlinux: I think m should be wrapped in tr() ? Or is it used universally?

timlinux:
LGTM - see my notes about using tr() around m. Please back port to master.

Gustry:
It seems that `m` is universally. It's the official symbol for meter : https://en.wikipedia.org/wiki/Metre

PR #2695, opened 2016-03-21
Tsunami IF

Gustry:


safe/impact_reports/road_exposure_report_mixin.py
```diff
@@ -142,14 +157,16 @@ def roads_breakdown(self):
                 else:
                     affected_by_usage.append(0)
             row = m.Row()
-            row.add(m.Cell(road_type.capitalize()))
+            row.add(m.Cell('%s (m)' % road_type.capitalize()))
```
timlinux: we should probably tr() m as it is probably different in different locales

safe/impact_reports/road_exposure_report_mixin.py
```diff
@@ -163,16 +180,19 @@ def roads_breakdown(self):
             total_affected[count] = number_affected
 
         row = m.Row()
-        row.add(m.Cell(tr('Total'), header=True))
+        row.add(m.Cell(tr('Total (m)'), header=True))
```
timlinux: As above for translation

timlinux:
LGTM @Gustry please merge

timlinux:
Please back port to master after merging. Thanks!

PR #2643, opened 2016-02-24
Make Experimental happy again

ismailsunni:
For #2637

PR #2709, opened 2016-04-04
Landcover impact function

wonder-sk:
Rebased the work on land cover impact function to current state of develop branch

Implements #1216


wonder-sk:
Added a simple pivot table implementation that will allow easy generation of summary tables.

This is how the report looks now: 1) with support for aggregation zones, 2) with experimentally added bar charts for each land cover type

Report

@samnawi Feel free to comment if you have any suggestions on how to improve the report

timlinux:
This looks super nice, thanks @wonder-sk 

@Gustry could I assign this to you to resolve merge conflicts and then get it merged into develop? Please ping @samnawi in the gitter channel and see if he is ok with the proposed changes before we merge.

Thanks

samnawi:
Hi @wonder-sk 
Sorry, I just realized this PR.
LGTM

The result from aggregation zone also include in the attribute of the layer.

But, I have some question for this PR,
1. What is the function of this column
image

I think it will be redundant, because the user has been decided the field which is processed in the keyword wizard.

I think it's like deja vu, because I ever comment like this in [here](https://github.com/inasafe/inasafe/pull/2643#issuecomment-188131801)

2. For aggregation zone n the report, Could we change the highlight background? Because it similar with Analysis result and analysis detail, possibly to make confuse. Maybe it will better if just Highlight on the aggregation zone title, and then the region use the bold font.

Thank you.

wonder-sk:
Thanks @samnawi for comments. Indeed the parameter was not necessary - I have removed it and switched to using keywords from exposure layer.

For the second comment, it should be easy to change styling of header for aggregration zones - I am just a bit unfamiliar with the styling rules, probably other guys know better whether there's a recommended CSS class for sub-headings or if a new CSS class should be added. Not sure who is the best person to tackle it...

Gustry:
@wonder-sk I've updated your PR with a conflict, but there is a failing test not related to the merge.

samnawi:
Hi @wonder-sk  The PR is running very well without aggregation. But, when I tried to run with aggregation zone, there is an error message like this
image

samnawi:
@wonder-sk and @Gustry Could we make this PR fix before Wednesday? because we want to use it on Tsunami Ambon scenario arrangement in BNPB.

Thank you.

Gustry:
FYI @wonder-sk,  I moved the test dataset to Jakarta, it's more convenient if data are all in Jakarta, especially if we want to use the aggregation area provided.

I tried to launch the analysis with the aggregation area, I got the error : 
```

ValueError

An exception occurred when running the impact analysis.
max() arg is an empty sequence
Diagnostics

In file "/home/etienne/.qgis2/python/plugins/inasafe/safe/impact_functions/base.py", line 699, in run_analysis self._impact = calculate_impact(self)
In file "/home/etienne/.qgis2/python/plugins/inasafe/safe/engine/core.py", line 176, in calculate_impact result_layer = impact_function.analysis_workflow()
In file "/home/etienne/.qgis2/python/plugins/inasafe/safe/impact_functions/base.py", line 685, in analysis_workflow impact = self.run()
In file "/home/etienne/.qgis2/python/plugins/inasafe/safe/impact_functions/generic/classified_polygon_landcover/impact_function.py", line 181, in run impact_summary = impact_table = _format_report(report_data)
In file "/home/etienne/.qgis2/python/plugins/inasafe/safe/impact_functions/generic/classified_polygon_landcover/impact_function.py", line 362, in _format_report bar_chart=True)
In file "/home/etienne/.qgis2/python/plugins/inasafe/safe/impact_functions/generic/classified_polygon_landcover/impact_function.py", line 263, in format_pivot_table max_value = max(max(row) for row in pivot_table.data)
```
I'm looking about the pivot_table

wonder-sk:
OK so things should work fine now also with aggregation. Tested also with Ambon/Tsunami data from @samnawi 

For the test data (now moved to Jakarta), there is a bug in InaSAFE that causes the impact layer to be loaded with wrong CRS. This is due to Web Mercator CRS and due to the fact that internally in InaSAFE, QGIS impact layer is saved with OGR and then loaded in QGIS again, loosing extra projection information (stored in .qpj file). This should be raised as a separate issue and it is not directly related to this PR.

samnawi:
Yeah @wonder-sk I can run your PR with aggregation zone.

ismailsunni:
@samnawi can we merge this to develop then?

samnawi:
OK @ismailsunni, the PR is working properly in my PC

Charlotte-Morgan:
hi @wonder-sk 
this is looking great.
The prototype graph is good to see but probably does not need to be included in the detailed report. I think the table of numbers is fine.
Would it be possible to make the order in teh table Low Hazard Zone on the left .. Medium and High on the right.  Please look at the tsunami on roads report for recently updated styling.
We also have a convention of omitting 0 values

We need to add notes and assumptions. If 0 values are omitted then this is usually stated in the notes.

Gustry:
For the notes, there is a ticket : https://github.com/inasafe/inasafe/issues/2764
I guess we should create some new tickets as this pull request is closed now.

Issue #2673, opened 2016-03-11
We need a new panel in wizard for class assignments


timlinux:
The idea is that if we have 

* roads
* buildings
* in the future landcover

The user should map the concepts in the type column to known concepts. These will be used in aggregation.

There should be a simple ui similar to the wet / dry classification tree where the user drags type values from the left into the classification tree on the right.

borysiasty:
/me will ask @Gustry for more info next week.

ismailsunni:
The steps:
1. @borysiasty will add new step in the wizard to map the exposure. You can find the list of the mapping value in here https://github.com/inasafe/inasafe/blob/develop/safe/postprocessors/building_type_postprocessor.py#L40 (for building, you can find the other in the _similar_ python files.
2. @ismailsunni to make it sure the keywords is working properly
3. @Gustry will use the value mapping to use in the aggregator. See #2675

timlinux:
Here are the current classes we map to:

### Buildings:

* Medical
* Schools
* Places of worship (I think we should rather change this to 'Place of worship'
* Residential
* Government
* Public Building
* Fire Station
* Police Station
* Supermarket
* Commercial
* Industrial
* Utility
* Sports Facility
* Other

### Roads

* Motorway / highway
* Motorway link
* Primary road
* Primary link
* Tertiary
* Tertiary link
* Secondary
* Secondary link
* Road, residential, living street, etc

borysiasty:
@ismailsunni @timlinux @Gustry - as I understand the only solution (apart from hardcoding ;) ) is to read the maps every time from BuildingTypePostprocessor.field_values and RoadTypePostprocessor.field_values. Some day maybe we'll move them to the metadata, but for now this is the way to go, isn't it?

borysiasty:
@ismailsunni @timlinux @Gustry - After some thinking (and working) I guess we can directly use the existing classification (mapping) step. From wizard's point of view the situation is almost identical. I only see one difference: for hazards, the available classifications are defined in impact functions and returned by ImpactFuctionManager (thus there may be multiple classifications available for given layer).

For the exposure classification, the wizard has to read the mapping directly from BuildingTypePostprocessor.field_values and RoadTypePostprocessor.field_values. Please let me know if there any future plans to unify those hazard and exposure mapping/classification or not (I suppose there aren't at the time, as the purpose is different).

borysiasty:
@timlinux - in your list of classes for roads you didn't include
 - Track
[...]

PR #2719, opened 2016-04-08
[Keyword Wizard] Add structure and road classification (#2673)

borysiasty:
This PR makes the wizard showing the mapping (classify) step also for classified roads and structures. The default mapping is read from BuildingTypePostprocessor and RoadTypePostprocessor (unlike  hazard classifications, which are read from ImpactFunctionManager), and the result is put to the keywords as 'exposure_value_map' (I'm not sure if it should be the final name).

Finally, the result is ignored by KeywordIO.write_keywords(), as ExposureLayerMetadata doesn't handle the keyword 'exposure_value_map'. @ismailsunni @timlinux @Gustry could someone decide how this keyword should be stored, please?

### DIFF
--- a/safe/gui/tools/test/test_wizard_dialog.py
+++ b/safe/gui/tools/test/test_wizard_dialog.py
@@ -33,7 +33,6 @@ pardir = os.path.abspath(
     os.path.join(os.path.dirname(__file__), '../../..///'))
 sys.path.append(pardir)
 
-from safe.impact_functions import register_impact_functions
 from safe.common.utilities import temp_dir
 from safe.test.utilities import (
     clone_raster_layer,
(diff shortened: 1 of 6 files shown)


ismailsunni:
hi @borysiasty 
Thanks for the PR, looks great.

If you are able to add the exposure class mapping (like the one that you said in the email), please go on :) 

If you get some failed test regarding the metadata, please revert the adding and finalise the PR. I will help you add the new metadata keywords.

I prefer to not import anything from PostProcessor, since we want to refactor it. May be it will be placed in the definitions.py or something similar.

ismailsunni:
This PR is for #2673 *merely add reference

borysiasty:
@ismailsunni - I removed dependencies on postprocessors and made a wizard_metadata.py file instead. It's meant to be a temporary source of metadata not put to the proper places yet (the file will be removed after imports are replaced with proper ones).

The metadata are now 'structure_class_mapping' and 'road_class_mapping' and they are waiting for the final decision :)

PR #2726, opened 2016-04-15
pep8, remove deprecated methods, small UI changes in the OSM downloader

Gustry:

### DIFF
--- a/realtime/flood/peta_jakarta_api.py
+++ b/realtime/flood/peta_jakarta_api.py
@@ -17,8 +17,8 @@ class PetaJakartaAPI(object):
 
     @classmethod
     def get_aggregate_report(cls, duration, level):
-        rest_point = \
-            'https://rem.petajakarta.org/banjir/data/api/v2/rem/flooded'
+        rest_point = (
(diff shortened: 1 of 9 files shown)


timlinux:
Hi @Gustry If you make ui changes, could you include a screenshot in the PR too please?

Gustry:
Sorry @timlinux . I have just removed default checked features (roads and buildings polygons).

realtime/flood/peta_jakarta_api.py
```diff
@@ -17,7 +17,8 @@ class PetaJakartaAPI(object):
 
     @classmethod
     def get_aggregate_report(cls, duration, level):
-        rest_point = 'https://rem.petajakarta.org/banjir/data/api/v2/rem/flooded'
+        rest_point = \
```
ismailsunni: Which is better, using backslash or brackets for case like this?

In my knowledge, backslash has a risk to introduce an error since it's not easy to see a line continuation.

Gustry: I think you are right. I'm going to change.

ismailsunni:
LGTM

PR #2727, opened 2016-04-15
manage a timeout error from the overpass in the OSM Downloader

Gustry:
Need to backport to develop.

CC : @timlinux can you review and release 3.3.2 ?


timlinux:
Thanks very much @gustry LGTM

PR #2732, opened 2016-04-21
add debug helper functions

Gustry:
I might become lazy, but I like debugging geometries by displaying them in the canvas.
This PR just brings simple functions to display a QgsGeometry for instance.


timlinux:
LGTM

PR #2734, opened 2016-04-21
PEP8.

ismailsunni:
It's a small PR fixing the PEP8, while checking if I have disabled scrutinizer.


Gustry:
I have already commited a PEP 8 fix in a open PR : 
https://github.com/inasafe/inasafe/pull/2726

ismailsunni:
@Gustry ah, I didn't see it. We can close this then :D The main reason is to check the scrutinizer's status only.

PR #2741, opened 2016-04-23
Fix 2740 update exposure map

ismailsunni:
Updating value map for road and structure. Please check @samnawi


samnawi:
Hi @ismailsunni I've been tried your PR.
LGTM for the road structure
image

But, there is some bug in this PR :
1. Still could not compute the Closed Road
image
2. I could run with aggregation zone, but there is no result in the report
image
3. Group description was truncated for 1st mouseover, but full text at the next mouseover.
1st
screenshot_6
2nd
screenshot_7
4. Regrouping in the structure was fail.
image

## Propose
I just realized if there is no classification value for residence, could we add that class?

Thank you

PR #2742, opened 2016-04-24
Json impact data

ismailsunni:
Refactor impact report to impact data and saved as json file. No more impact_report in keyword.
This is still WIP to test regularly in Travis


safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
```diff
@@ -182,75 +182,66 @@ def action_checklist(self):
...
-            'The fatality calculation assumes that '
-            'no fatalities occur for shake levels below 4 '
-            'and fatality counts of less than 50 are '
-            'disregarded.'))
-
+        title = tr('Notes and assumptions')
+        fields = [
+            tr('Total population in the analysis area: %s') %
+            format_int(population_rounding(self.total_population)),
+            tr('<sup>1</sup>People are displaced if they experience and '
```
timlinux: We probably need to get rid of mark up embedded in the text like this as it won't map well to other representations e.g. plain text.

safe/impact_functions/generic/classified_polygon_population/impact_function.py
```diff
@@ -63,29 +61,26 @@ def notes(self):
...
-            'All values are rounded up to the nearest integer in '
-            'order to avoid representing human lives as fractions.'))
-        checklist.add(tr(
-            'Population rounding is applied to all population '
-            'values, which may cause discrepancies when adding values.'))
-        message.add(checklist)
-        return message
+        fields = [
+            tr('Total population in the analysis area: %s') % population,
+            tr('<sup>1</sup>People need evacuation if they are in a hazard '
```
timlinux: Same comment as above re: markup

safe/impact_functions/earthquake/itb_earthquake_fatality_model/impact_function.py
ismailsunni: Should we just remove it? Since it's very specific to html.

timlinux:
This is looking really nice. Just remember that somewhere (the wiki) we should document (at a high level) the json document so it is clear what the sections are etc. That can happen post release. I really like how this is looking @ismailsunni 

Let me know when I can do more detailed testing (I only did a code review thus far).

ismailsunni:
Update:
It's working now if we run analysis in dock and wizard. No more html, but json.

Unfortunately, the post processing json data breaks many things.... I am still working to make it all pas...

Gustry:
> Unfortunately, the post processing json data breaks many things.... I am still working to make it all pas...

@ismailsunni I'm also working on the postprocessing. Should I wait for your merge before to continue or not ? I would like to change some core things, because I have some issues.

ismailsunni:
@Gustry it depends on which part you are working on. If you need to read to post processing report (in html), you will need me to merge it first. Our previous tests was using false positive test control for checking the impact report in text (made me scratch my head a lot).

But, as far as I know, I only modified how dock and wizard get the post processing report. The method itself is not called anywhere else. So, 95% you can go on without me merge first.

Issue #2744, opened 2016-04-26
Different Result in Population IF at Master and Develop Version


samnawi:
### Problem
When I run Population IF (Tsunami vector VS Population vector), using the same data (hazard and exposure), I found a different result.

This is the result in Develop version
![image](https://cloud.githubusercontent.com/assets/8871958/14816977/31377f1e-0bde-11e6-92e3-394d45b65cd1.png)

This is the result in Master
![image](https://cloud.githubusercontent.com/assets/8871958/14817026/77e2bfaa-0bde-11e6-84aa-5217b48bbd93.png)

Which one is correct?

### Environment
Win 7 64bit
QGIS 2.14.1

### Proposed Solution
Make it happy

Thank you.

ismailsunni:
This is the data. @samnawi couldn't upload it here, so I helped him:
[Ambon.zip](https://github.com/inasafe/inasafe/files/236544/Ambon.zip)

samnawi:
When I try to compare with Tsunami evacuation IF (Tsunami Raster VS Population Raster), the results approached the InaSAFE master.
![image](https://cloud.githubusercontent.com/assets/8871958/14818388/0efe0f4a-0be7-11e6-8544-0cea6fdd12d3.png)

FYI, the data that I use in this [comment (1st comment above)](https://github.com/inasafe/inasafe/issues/2744#issue-151112249) was generated from the raster data that I use in here.

ismailsunni:
The Tsunami Raster Data that @samnawi mentioned above: [Tsunami-raster-ambon.zip](https://github.com/inasafe/inasafe/files/236629/Tsunami-raster-ambon.zip)

ismailsunni:
The regression is re-producible.
Actually, our test is not covering this regression, so I made a change to our data test for census layer. You can get the updated data test here : https://github.com/ismailsunni/inasafe/commit/23949ac09fbfdc91d86ff6c641b3019be695cedf

Here the result from master:
<img width="1304" alt="inasafe - 2744 - master" src="https://cloud.githubusercontent.com/assets/1421861/14839340/46bb1f8c-0c54-11e6-8436-0fab33bfe58c.png">
And here the result of develop
<img width="1304" alt="inasafe - 2744 - develop" src="https://cloud.githubusercontent.com/assets/1421861/14839351/692dbbf6-0c54-11e6-9abb-09cde3a0ae3e.png">

[...]

PR #2745, opened 2016-04-27
Fix 2744 regression people polygon

ismailsunni:
For fixing #2744.
Please check @samnawi


samnawi:
@ismailsunni LGTM
I've test with the Ambon Scenario, and the result was same with the Master
image

PR #2753, opened 2016-05-03
Remove duplicate metadata in the wizard

Gustry:
@borysiasty The wizard metadata has been moved to safe/definitions.py so I updated your code to use it.


ismailsunni:
LGTM

Issue #2491, opened 2015-12-02
Realtime for volcanic ash


Charlotte-Morgan:
# Background

Geoscience Australia has been working technical agencies in Indonesia to develop a volcanic ash modelling capability. After being shown InaSAFE, the volcanic ash modelers are keen to find ways to share information with DM and communities and to share hazard data for situational awareness during a volcanic eruption.

InaSAFE version 3.4 will include a new IF for ash on land cover.
there is also a request for a real-time interface.
# proposal
- [ ] volcanic ash hazard data to be provided by technical agency
- [ ] new IF for ash impact on crops - calculate area affected
- [ ] ash impact on people - number of people affected (no fatality calculation for this)
- [ ] temporal analysis, time slices with impact of cumulative ash build up over time - NEW
# impact map

First draft of what an impact map might look like - note this is a static pdf :(
![image](https://cloud.githubusercontent.com/assets/5743135/11528720/0eb7314a-991b-11e5-98d9-ccb843fc83e0.png)
# comments

There could be many similarities between flood and volcanic ash (not really), ie in the way the data may be supplied and the type of analysis and presentation of impact analyses. eg not a single event like an earthquake; more likely to be identified by a time stamp; time series analysis or reporting will be required; compounding effect of ash buildup or prolonged duration of inundation
# user requirements

Workshops and meetings will be held to determine more specific user requirements. Please add any details that are needed:
- [ ] which organisation will provide hazard data and in what format
- [ ] which organisation will host the real time system
# cc

@lucernae @samnawi @timlinux

timlinux:
@Charlotte-Morgan thanks for making a report mockup - that is very useful. We will stand by for further details on requirements from your volcano colleagues.

fredychandra:
HI @Charlotte-Morgan 
I think It will be nice if we can also adding how many crops and settlements area will be covered by ash potentially in that report.

samnawi:
:+1: for @fredychandra suggest to add the impacted for crops and settlements area. The other important thing is also the impact for the airport, as we know when the volcanic eruption happens,one of the issue is about air transportation.

Charlotte-Morgan:
This proposal is being driven very much by the technical and user community and I have simply documented what they have proposed.
They will be holding workshops to clarify user requirements; the first is scheduled for 27 January - 5 February in Melbourne. It will involve GA, BOM, and volcano experts from BG.
[...]

PR #2757, opened 2016-05-04
Classify and polygonize a raster

Gustry:
This PR brings 3 new functions : 
* reclassify
* polygonize_band
* reclassify_polygonize (the mix of these 2)

For the reclassify, I found an algorithm here : https://github.com/chiatt/gdal_reclassify
I have copy/paste the code for now. @timlinux Do you want me to re-write it ? The code seems working for now. I wrote some unit tests.

I would like to continue this PR for the landcover and maybe change every IF using a raster (except earthquake).

### DIFF
--- a/safe/gis/reclassify_gdal.py
+++ b/safe/gis/reclassify_gdal.py
@@ -23,7 +23,7 @@ from gdalconst import (
     GDT_Float32,
     GDT_Float64)
 
-from safe.gis.gdal_ogr_tools import polygonize_band
+from safe.gis.gdal_ogr_tools import polygonize
 from safe.common.utilities import temp_dir, unique_filename
 from safe.common.exceptions import FileNotFoundError, ReadLayerError
 
@@ -220,6 +220,14 @@ def _process_reclassify(
 
 
 def _classes_to_string(ranges):
+    """Function to transform the ranges dictionary as comma-delimited string.
+
+    :param ranges: The ranges.
+    :type ranges: OrderedDict
+
+    :return: List(output classes, input classes).
+    :rtype: list
+    """
     output_ranges = []
     for interval in ranges.itervalues():
         if interval[0] is None:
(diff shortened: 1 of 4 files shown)


safe/gis/reclassify_gdal.py
```diff
@@ -0,0 +1,296 @@
...
+    dst_ds.SetGeoTransform(transform)
+    if nodata in ["True", "true", "t", "T", "yes", "Yes", "Y", "y"]:
+        out_band.SetNoDataValue(default)
+
+    out_band.GetStatistics(0, 1)
+    dst_ds.SetProjection(proj)
+
+
+def _classes_to_string(ranges):
+    output_ranges = []
```
Gustry: Missing docstring

safe/gis/reclassify_gdal.py
```diff
@@ -0,0 +1,296 @@
...
+
+    :return: The file path to the reclassified raster.
+    :rtype: str
+    """
+    temporary_dir = temp_dir(sub_dir='pre-process')
+    output_raster = unique_filename(
+        suffix='-reclassified.tiff', dir=temporary_dir)
+
+    output_classes, input_classes = _classes_to_string(ranges)
+    default = 0
```
Gustry: Explain why we should avoid 0 in the ranges

timlinux:
Hi @Gustry 

Cool stuff. Definitely don't rewrite it - its great to re-use existing code where possible. One thing I am curious about is how we do multiple classifications. Bear in mind the use case is to reclassify a raster from continuous data into a single class per range e.g.

screenshot_22

So in that case we would reclassify by this scheme:

Original Value | Class
-------------------|----------
0 | 0
0.01 - 0.2 | 1
0.2 - 2 | 2
2.1 - 5 | 3
5.1 - 10 | 4
10 - 99 | 5

If that use case is satisfied by this PR can you provide a small test to illustrate it and then LGTM to merge

timlinux:
See also #2491

Gustry:
Yes, you can already do this kind of classification.
It's included in the unit test already :  https://github.com/inasafe/inasafe/pull/2757/files#diff-a254ca94a99e82265e74f0f479c44fbaR45
I will add some documentations.

That's why I would like to update IF about tsunami for instance. I had some performance issue last time with tsunami raster.

safe/gis/reclassify_gdal.py
```diff
@@ -0,0 +1,326 @@
...
+    return ranges.keys(), output_ranges
+
+
+def reclassify(input_raster, ranges, no_data=0):
+    """Reclassify a raster according to some ranges.
+
+    This function is a wrapper for the code from
+    https://github.com/chiatt/gdal_reclassify
+
+    For instance if you want to classify like this table :
```
Gustry: @timlinux To classify with multiple thresholds

timlinux:
OK great LGTM once tests pass.

Issue #2449, opened 2015-10-22
Wizard code cleanup


borysiasty:
- [x] Refactor into small modular architecture: one class per wizard tab.
- [x] We need to have an easy way to add new IF types e.g. @Samweli 's code does not 'just work' when he added his IF
- [x] Address fragility of tests

timlinux:
@borysiasty could you make some time in the next 2 weeks to look at this please?

borysiasty:
Ok!

Where can I find more details about problems with @Samweli 's code? Is it just the need to update tests in PR #2442 or doesn't it work at all?

borysiasty:
Ok @timlinux !

Where can I find more details about problems with @Samweli 's code? Is it just the need to update tests in PR #2442 or doesn't it work at all?

ismailsunni:
hi @borysiasty @timlinux 
I checked in @Samweli 's PR, all is good. It's working in wizard also.

We only need to add an image/icon for the new exposure type:
![area icon](https://cloud.githubusercontent.com/assets/1421861/11232423/9cca230c-8de5-11e5-8242-44378a3aa782.png)

Samweli:
hi @ismailsunni, i have added an icon for area, you can find it here https://thenounproject.com/term/inland/8685/

ismailsunni:
Hi @Samweli 
Thanks. Actually, if you have already the icon for area, you can add it to InaSAFE by yourself in your PR. You only need to put it under `resources/img/wizard/` and give a proper name. It should be something like `keyword-subcategory-area.svg

Samweli:
Ok, thanks @ismailsunni , i have already done that, here https://github.com/Samweli/inasafe/blob/develop/resources/img/wizard/keyword-subcategory-area.svg

ismailsunni:
Done in #2771

PR #2771, opened 2016-05-08
Wizard updates

borysiasty:
- Improve test coverage from 79% to 87%
- Split tests into smaller files
- Minor bugfixes and code cleanups


Gustry:
It seems very good, thanks !

safe/gui/tools/wizard/test/test_function_centric_wizard.py
```diff
@@ -0,0 +1,897 @@
...
+# safe.gui.tools.wizard
+QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
+
+from safe.definitions import inasafe_keyword_version
+from safe.impact_functions import register_impact_functions
+from safe.gui.tools.wizard.wizard_dialog import WizardDialog
+from safe.gui.widgets.dock import Dock
+from safe.utilities.keyword_io import KeywordIO, definition
+
+DOCK = Dock(IFACE)
```
ismailsunni: Hi @borysiasty 
Please put the Dock object inside the class, for example in setUp method and set it as class member. Example : https://github.com/inasafe/inasafe/blob/develop/safe/gui/widgets/test/test_dock.py#L81

The reason is sometimes if we initialise outside the test class, the object will keep alive when the test finished, and "disturb" another test.

safe/gui/tools/wizard/test/test_function_centric_wizard.py
```diff
@@ -0,0 +1,897 @@
...
+
+    def check_current_step(self, expected_step):
+        """Helper function to check the current step is expected_step
+
+        :param expected_step: The expected current step.
+        :type expected_step: WizardStep instance
+        """
+        current_step = expected_step.parent.get_current_step()
+        message = ('Expected %s but I got %s' % (expected_step, current_step))
+        self.assertEqual(expected_step, current_step, message)
```
ismailsunni: I think we can use the default failed assert message, rather than creating our own here.

ismailsunni:
hi @borysiasty 
I made some comments above, please address it before merging the PR. Thanks :)

safe/gui/tools/wizard/test/test_function_centric_wizard.py
timlinux: Yes I switched to using the default message as it provides a richer reflection of the variable state and is less maintenance overhead.

timlinux:
LGTM too once @ismailsunni 's comments are taken into account - thanks @borysiasty !

safe/gui/tools/wizard/test/test_function_centric_wizard.py
borysiasty: Ah great! :) I used to comment them out for debugging :-D

borysiasty:
@ismailsunni - thanks for the review, I've fixed both issues. Can we close #2449 now?

timlinux:
Thanks @borysiasty. @ismailsunni please merge when you are ready.

PR #2775, opened 2016-05-09
Fix #2724.

### LINKED ISSUE #2724: InaSAFE Intermediate Ex. errors: Python error
### Problem
I'm trying to work my way through the InaSAFE tutorial exercises, the Basic Exercise went OK, but the [Intermediate Exercise]  (http://docs.inasafe.org/en/training/socialisation/run_intermediate_inasafe.html) part 1  has thrown up a Python error and an OSM downloader error (the latter listed separately)
### Expected Behaviour
Running the application after defining the relevant keywords for 'Jakarta District' dataset should result in a set of statistics regarding the affected population relative to the districts. Instead a warning is received regarding the following Python error: 

```
2016-04-14T12:42:23	1	Traceback (most recent call last):
			  File "C:/Users/airv/.qgis2/python/plugins\inasafe\safe\gui\widgets\dock.py", line 1160, in accept
			    self.analysis.setup_analysis()
			  File "C:/Users/airv/.qgis2/python/plugins\inasafe\safe\utilities\analysis.py", line 311, in setup_analysis
			    self.setup_impact_function()
			  File "C:/Users/airv/.qgis2/python/plugins\inasafe\safe\utilities\analysis.py", line 470, in setup_impact_function
			    self.aggregator.deintersect()
			  File "C:/Users/airv/.qgis2/python/plugins\inasafe\safe\impact_statistics\aggregator.py", line 486, in deintersect
			    self.hazard_layer)
			  File "C:/Users/airv/.qgis2/python/plugins\inasafe\safe\impact_statistics\aggregator.py", line 1487, in _prepare_polygon_layer
			    int_com.difference(int_sym))
			AttributeError: 'NoneType' object has no attribute 'difference'
```

### Environment
Windows 7 Pro 64bit
QGIS 2.14.1 64bit
InaSAFE 3.3.1

### Proposed Solution
Any useful advice gratefully received

ismailsunni:
We need to write the layer first before creating the layer metadata file. If we create the metadata file, InaSAFE will assume that the layer is not file based, then the metadata will be stored in the database. However, after the layer is written to the disk, InaSAFE will assume that the layer is file based, which make InaSAFE try to find .xml for the metadata. InaSAFE doesn't find the xml, and generate default metadata (with keyword version only).
Hence, it will fail when there is a verification.

please review @timlinux

### DIFF
--- a/safe/impact_statistics/aggregator.py
+++ b/safe/impact_statistics/aggregator.py
@@ -1329,7 +1329,6 @@ class Aggregator(QtCore.QObject):
         temporary_dir = temp_dir(sub_dir='pre-process')
         out_filename = unique_filename(suffix='.shp', dir=temporary_dir)
 
-        self.copy_keywords(layer, out_filename)
         shape_writer = QgsVectorFileWriter(
             out_filename,
             'UTF-8',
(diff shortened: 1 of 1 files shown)


timlinux:
Ha that is tricky! Good catch @ismailsunni ! LGTM

ismailsunni:
haha, yes, indeed. Thanks to PyCharm remote debugger. :)

PR #2791, opened 2016-05-13
fix #2790 wrong projection if we use 3857

### LINKED ISSUE #2790: Wrong projection for the impact layer if the input is 3857
### Problem
If the input layer is in 3857, the impact layer will be in 54004.
It seems to be a QGIS/OGR problems : 
https://hub.qgis.org/issues/11134

### Proposed Solution
* Fix in QGIS/OGR
* Or as a quickhack, hardcode the reprojection in inasafe

Gustry:


ismailsunni:
LGTM

PR #2799, opened 2016-05-18
remove capital letters and add the main class to the default strings

Gustry:
These strings are capitalized in the python code


ismailsunni:
LGTM

PR #2808, opened 2016-05-20
fix #2676 don't use the translated parameter name in the minimum needs

### LINKED ISSUE #2676: BUG: Can not add new item and edit for Minimum Needs Manager in QGIS Bahasa Indonesia
# Problem
Add new item using InaSAFE minimum needs manager in QGIS using Bahasa Indonesia does not work.

# Detail
In InaSAFE Training last week in The National Development University, Yogyakarta, I taught participant to **adjust Minimum needs** for Flood Impact Function using **Minimum needs manager** tool. In this training we using QGIS in Bahasa Indonesia because participants think that would be easier for them to understand all QGIS and InaSAFE tools in Bahasa Indonesia. Unfortunately, we had Python Error when we want to save new minimum need that we added into one of profiles in minimum needs manager. This problem occured not only for flood but other hazard impact function as well.

However, when I try to adjust minimum needs using QGIS in English Languange, It went well. I can save my new item and run it.

# Screenshoot
minimum needs error
minimum needs error2

# Platform
Windows 10 64 Bit
RAM 8 GB
QGIS 2.8.6 Wien (Long Term Release)
InaSAFE 3.2.6

cc : @timlinux @ismailsunni @samnawi

Gustry:


ismailsunni:
LGTM

PR #2821, opened 2016-05-20
Fix 2770 missing value map

### LINKED ISSUE #2770: Missing value_map should be handled gracefully
### Problem

I am not sure who the user was, but today we got a sentry report:

http://sentry.kartoza.com/kartoza/inasafe-desktop/group/6330/

```
KeywordNotFoundError
'value_map'

python/plugins\inasafe\safe\storage\safe_layer.py ? in keyword
 
            raise KeywordNotFoundError(e)
python/plugins\inasafe\safe\impact_functions\generic\classified_polygon_population\impact_function.py ? in run
 
        self.hazard_class_mapping = self.hazard.keyword('value_map')
python/plugins\inasafe\safe\engine\core.py ? in calculate_impact
 
    result_layer = impact_function.run()
python/plugins\inasafe\safe\utilities\analysis.py ? in run_analysis
 
            self.impact_layer = calculate_impact(self.impact_function)

```
### Expected Behaviour

When value_map is missing, I think we should default at the minimum to putting everything in the **other** bucket and still completing the analysis successfully.

### Proposed Solution

Trap exceptions where the value_map is not found in the keywords.

# CC

@Charlotte-Morgan @samnawi

ismailsunni:
For fix #2770

I have added more test about the bug. Anyone want to test this?

cc @timlinux


safe/impact_functions/base.py
```diff
@@ -1408,7 +1408,10 @@ def _calculate_impact(self):
         start_time = datetime.now()
 
         # Run the IF. self.run() is defined in each IF.
-        result_layer = self.run()
+        try:
+            result_layer = self.run()
+        except KeywordNotFoundError as e:
+            raise KeywordNotFoundError(e)
```
Gustry: Don't you want `raise` only ? If you raise a exception, you will loose the previous traceback in the `run()`

timlinux:
LGTM @ismailsunni - please resolve the discussion with @Gustry before merging.

safe/impact_functions/base.py
ismailsunni: TIL...
So, I was always thinking that I need to raise the same exception if I want to catch it later.
To sum up, if I have already raise certain exception, I don't need to catch it, unless I need it to do more process (e.g. giving error message), is it right?
And the exception will be passed until someone catches it, right?

Let me know I am wrong.

Thanks @Gustry

Gustry: You are right now ;-)
You don't need to catch and re-raise the exception at every levels.
Something like this will work : 
```
try:
    func1()
except Exception:
    print 'error'


def func1():
    func2()


def func2():
    func3()


def func3():
    raise Exception('test error')
```

To sum up, in InaSAFE, we should just raise the exception where the problem occurs, and then finally catch it in UI to show the problem to the user.

Gustry:
Hi Ismail, I answered your comment.
You need to resync your branch, probably because of the landcover.

ismailsunni:
@Gustry thanks for the reply, I read some articles about it this morning also :D

PR #2822, opened 2016-05-20
[IFCW]Reset previously set aggregation layer. Fixes #2818

### LINKED ISSUE #2818: [Bug]  IFCW: Cannot set "No Aggregation" if the previous action is using Aggregation
### Problem
This minor bug appear when i previously run analysis using IFCW with aggregation. When i tried to run the second analysis without close the QGIS first (using the same QGIS and the same data that already loaded in QGIS layer list) and i don't using aggregation in this second analysis, the IFCW will read the aggregation layer in the end of the summary information before run the analysis.

Here's step to produce this error:

1. Click IFCW
2. Select hazard type, add hazard, add exposure and select using aggregation first.
3. The InaSAFE will successfully run Analysis using aggregation
4. Without close the QGIS first, click IFCW again and try to run the same hazard analysis
5. Select hazard & exposure that already in the layer list
6. Do not select the aggregation layer. Choose "No thanks, I'm happy to aggregate results for entire analysis window" and click Next.
7. Select analysis area and postprocessor
8. In the IFCW analysis summary, you will realize that IFCW will still using aggregation layer even though you already define not to use aggregation 

This bug only occur when you run IFCW using aggregation first and then you run IFCW for the second time for the same hazard but without aggregation.

### Expected Behaviour
If the user run IFCW first time using aggregation and in the second time the user run IFCW without aggregation, the IFCW will do the analysis based on the user input in the aggregation section (the user select "No thanks, I'm happy to aggregate results for entire analysis window"

### Environment
Windows 10 64bit 16GB
QGIS 2.8.9 InaSAFE 3.4 from github

borysiasty:


ismailsunni:
LGTM.

@adissadis can you test it? Since I can't reproduce the issue. Thanks

adissadis:
LGTM. Thank you @borysiasty !!

PR #2824, opened 2016-05-23
fix #2823 value mapping strings in the wizard

### LINKED ISSUE #2823: Wizard value mapping text needs improving
### Problem

Currently the new value map text looks like this:

screen shot 2016-05-21 at 23 22 54

Am I wrong or is it not true that it is not only the post processor that uses the value mappings, but the impact function itself? Maybe nice to say "We need to map the TYPE attributes to known categories."

Also replace "Un-mapped' with "Unclassified"

### Environment
* InaSAFE : master
* QGIS : all
* OS : all
* Data used : 
  * Exposure : /Users/timlinux/dev/python/inasafe/safe/test/data/exposure/roads.shp  
  * Hazard : 
  * Aggregagtion :

### Proposed Solution

Gustry:


timlinux:
LGTM

PR #2825, opened 2016-05-23
Fix how we count features for buildings, reordering columns in the report

### LINKED ISSUE #2819: Building postprocessor doesn't count properly
### Problem
Using the tsunami on buildings : 
selection_078

### Expected Behaviour
There is more than 2 closed buildings.

Gustry:
Fix #2819


ismailsunni:
LGTM

PR #2834, opened 2016-05-24
fix #2831 hide the unaffected column if this property is set to true

### LINKED ISSUE #2831: The road report shows unaffected column with 0.
### Problem
Running the raster tsunami on roads IF, it gives me `0` in the `unaffected column` even if the `dry zone` is not null.

### Expected Behaviour
* First, `unaffected` and `dry` zone should be equal.
* Secondly, it doesn't make sense to show `dry` and `unaffected` as it's a duplicate. We should show `affected` which should be the sum from low, medium, high and very high hazard classes.

### Environment
* InaSAFE : dev
* Data used :
  * Exposure : Roads Maumere_Roads_WGS84.shp
  * Hazard : Raster tsunami Maumere_Tsunami_WGS84.tif

Gustry:


ismailsunni:
LGTM

PR #2835, opened 2016-05-24
Ash Raster Landcover Polygon Impact Function

lucernae:
Additional commits:
- move register_impact_functions
- translation format fix


safe/gui/tools/help/dock_help.py
```diff
@@ -91,20 +91,23 @@ def content():
...
     bullets.add(m.Text(
-        tr('the '),
-        m.ImportantText(tr(
-            'questions ')),
-        tr('area')))
+        # format 'the __questions__ area' for proper i18n
+        tr('the %s area') % (
+            m.ImportantText(tr(
+                'questions')).to_html(),
+        )))
```
lucernae: @ismailsunni can you review this?
I'm having trouble to translate it with previous format because the string is separated: 'the', 'questions', and 'area'.
With this, we can just translate 'the %s area' and reorder it in Bahasa Indonesia as : 'daerah %s'

lucernae:
@ismailsunni also there are some changes I made in 'translation format fix' commit. Need some review.

safe/gui/tools/help/dock_help.py
ismailsunni: Hi @lucernae, I have tried way also, see my gitter chat last night. And it doesn't work properly since the ImportantText won't be rendered properly.

safe/impact_functions/unit_definitions.py
```diff
@@ -1,6 +1,6 @@
 # coding=utf-8
 from safe.definitions import unit_generic, unit_metres, unit_mmi, \
```
ismailsunni: Please avoid to use backslash, use bracket instead

safe/impact_functions/loader.py
```diff
@@ -0,0 +1,118 @@
+# coding=utf-8
```
ismailsunni: Any reason why we move this to `loader.py` ?

ismailsunni:
Please separate this PR into two PRs, ash IF related and translation formatting related

lucernae:
Closing this PR and split it.

PR #2836, opened 2016-05-25
Fix #2832, remove zonal stats and tidy up some tests.

### LINKED ISSUE #2832: Remove Python Zonal Stats
### Problem
Currently we have two zonal stats (python and QGIS). Since the zonal stats QGIS is more reliable, we should stop our support for python zonal stats. See issue: #2751 
### Expected Behaviour
- [x] InaSAFE should only support QGIS zonal stats
- [x] Remove related code to python zonal stat
- [x] Remove option to choose zonal stats in option dialog

### Environment
All

### Proposed Solution
See expected behaviour above.

ismailsunni:


Gustry:
Very nice to remove some code ;-)

ismailsunni:
@Gustry yeah, and make our test faster about 3 minutes in travis. It seems travis run the test slower than my laptop. There is no more than 20s for each test when I ran in my laptop.

PR #2837, opened 2016-05-25
Add Ash Raster Landcover Polygon Impact Function

lucernae:
- Add Raster Landcover Polygon Impact Function
- Move register_impact_function to new module called loader (to avoid get_qgis_app
deadlock when running individual tests of impact_functions)


lucernae:
@ismailsunni to answer your previous questions.
I moved register_impact_functions from init.py so I'm able to run nosetests for each IF individually.

ismailsunni:
@lucernae, that means, we should explicitly call loader method to load all IF, right? Ok, no problem for me.

lucernae:
@ismailsunni yeah, that's the consequence. So I wan to see if it breaks any tests first. Plugin.py will be not affected since it called register_impact_functions directly.
The problem arise when we try to run nosetests for individual impact functions, because previously when importing impact_functions, init.py is called and other IF is imported before get_qgis_app is called, so test won't run because QGIS instance isn't created yet.
However we also can't put get_qgis_app in the init.py, since doing so will break InaSAFE if used when opening QGIS normally. So I propose to seperate the loader from init.py. With consequence, you have to consciously import loader.py if you want to load all IF in tests file.

ismailsunni:
Hints for failed test:
1. Because you add ash layer in standard layer. It will be put in the top (sort alphabetically), thus get index = 0. And there is no IF for ash and building point.
2. no idea, but might be the same reason as point 1
3. Similar with point 1
4. Difference expected index since you added new layers.

Check this commit https://github.com/ismailsunni/inasafe/commit/a2db6f983eff75708a3b959e9f6533e41d082fde, you can cherry-pick it to get clearer error message.

safe/impact_functions/ash/ash_raster_landcover/impact_function.py
```diff
@@ -0,0 +1,244 @@
...
+                border_color='#000000',
+                transparency=0),
+        ]
+        style_info = dict(
+            target_field=self.target_field,
+            style_classes=style_classes,
+            style_type='categorizedSymbol')
+
+        extra_keywords = {
+            'map_title': tr('Affected Land Cover'),
```
Gustry: You need to remove `map_title` and put it in the metadata file

safe/impact_functions/ash/ash_raster_landcover/impact_function.py
```diff
@@ -0,0 +1,244 @@
...
+            'target_field': self.target_field
+        }
+
+        impact_layer_keywords = self.generate_impact_keywords(extra_keywords)
+
+        # Create vector layer and return
+        impact_layer = Vector(
+            data=impact_layer,
+            name=tr('Land cover affected by each hazard zone'),
```
Gustry: Same as above, you should move it to metadata as `layer_name`

Gustry: And remove `by each hazard zone` according to the new reporting standards.

ismailsunni:
LGTM @lucernae 
Minor comment: can you change the version added to 3.5 ?

We also need to test it by the user. cc @samnawi @ivanbusthomi @felix-yew @adissadis

PR #2838, opened 2016-05-25
Remove zonal stat code.

ismailsunni:
Sorry, I forget to remove the legacy code in the previous PR #2836


Gustry:
I think it's better now ;-)
Thanks !

PR #2841, opened 2016-05-25
Translation 3.4

lucernae:
CC @ismailsunni 
Some question for this:
- What happened when we 'tr' a dynamic text? Does the base string is generated?
- Alternative way is to extract the specially formatted strings that needs to be translated manually. (Put the generated text manually to transifex?)

### DIFF
--- a/safe/gui/tools/help/osm_downloader_help.py
+++ b/safe/gui/tools/help/osm_downloader_help.py
@@ -68,7 +68,7 @@ def content():
         'reappear.'))
     tips.add(tr(
         'Check the output directory is correct. Note that the saved '
-        'dataset will be named after the type of data being downloaded'
+        'dataset will be named after the type of data being downloaded '
         'e.g. roads.shp or buildings.shp (and associated files).'
     ))
     tips.add(tr(
@@ -98,9 +98,8 @@ def content():
     message.add(tips)
 
     message.add(m.Paragraph(
-        tr('When the'),
-        m.ImportantText(tr(' Political boundaries ')),
-        tr(
+        # format 'When the __Political boundaries__' for proper i18n
+        tr('When the %s '
             'box in the Feature types menu is ticked, the Political boundary '
             'options panel will be enabled. The panel lets you select which '
             'admin level you wish to download. The admin levels are country '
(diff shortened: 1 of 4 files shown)


ismailsunni:
@lucernae 
- AFAIK, the dynamic translation is not working for now.
- Manually adding text to transifex is prone to missing text, and I think it will be hard to manage

safe/gui/tools/help/osm_downloader_help.py
```diff
@@ -98,17 +98,19 @@ def content():
...
             'options panel will be enabled. The panel lets you select which '
             'admin level you wish to download. The admin levels are country '
             'specific. When you select an admin level, the local name for '
             'that admin level will be shown. You can change which country '
             'is used for the admin level description using the country drop '
             'down menu. The country will be automatically set to coincide '
-            'with the view extent if a matching country can be found.')))
+            'with the view extent if a matching country can be found.') %
+        (
+            m.ImportantText(tr('Political boundaries')).to_html(),
```
timlinux: But if we do this we are breaking separation between content and markup - now we will html embedded in the string....I think that is bad...

timlinux:
Ok I have some reservations about this because this approach embeds html markup into the strings which is against the design intent of the messaging framework. But I can't see a more elegant way to deal with this right now so I suggest to go ahead and merge here and then we can figure out a more elegant approach in the future.

ismailsunni:
Here the example of the problem that we found:
```
    message.add(m.Paragraph(tr(
        'There are three main areas to the dock panel:')))
    bullets = m.BulletedList()
    bullets.add(m.Text(
        tr('the '),
        m.ImportantText(tr(
            'questions ')),
        tr('area')))
    bullets.add(m.Text(
        tr('the '),
        m.ImportantText(tr(
            'results ')),
        tr('area')))
    bullets.add(m.Text(
        tr('the '),
        m.ImportantText(tr(
            'buttons ')),
        tr('area')))
```
Then in transifex we got :
screen shot 2016-05-24 at 21 46 05
It's very difficult to translate word by word in Indonesia or French. In Indonesia, we don't have `the`, and we have different order of word.

PR #2850, opened 2016-05-25
fix #2849 building report, fix align right/left

### LINKED ISSUE #2849: Report style for buildings
## Problem

Text is right aligned in impact reports where new structure mappings are used.

## Proposed solution
All text should be left aligned
All numbers should be right aligned

make it happy

Gustry:
@Charlotte-Morgan I fixed the text alignment for the breakdown and for the postprocessing, like `closed buildings`.

### DIFF
--- a/safe/impact_template/abstract_road_building_report_template.py
+++ b/safe/impact_template/abstract_road_building_report_template.py
@@ -79,10 +79,15 @@ class AbstractRoadBuildingReportTemplate(TemplateBase):
             table = m.Table(
                 style_class='table table-condensed table-striped')
             table.caption = v['caption']
+            attributes = v['attributes']
+
             header = m.Row()
-            for attribute in v['attributes']:
-                header.add(m.Cell(attribute, header=True))
-            header.add(m.Cell('Total', header=True))
+            # Bold and align left the 1st one.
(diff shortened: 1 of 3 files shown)


ismailsunni:
The code is LGTM. But, it looks strange. I can't decide it quickly, whether the column 0, 4, 7, 11 belongs to Medium hazard zone or High hazard zone.

Charlotte-Morgan:
Hi @Gustry  - based on your screenie above - I think we need to fix it some more.

I think the standard should be:
Alignment: numbers and number headings are right aligned, words and word headings are left aligned

So this means that:
Building type heading and content are left aligned
all the hazard zone headings and counts are right aligned.

Please look at all the good reports that are out there that do not need fixing and copy from them - you must have some standard code for reports that has all the basics in it. Tim did so much work on tidying up reports for 3.2

Gustry:
Thanks for the review @Charlotte-Morgan .
Sorry if I misunderstood. I was happy to update, because I was not convinced by my PR yesterday. As Ismail said, it was not very easy to read.
This is new status :

Charlotte-Morgan:
@Gustry - looking great thank you.
In some reports we have wrapped column headings too but I'm happy to wait for @easmetz  to get her head around some standards for these things.

PR #2905, opened 2016-06-03
fix #2899 impact layer should draw over the hazard layer

### LINKED ISSUE #2899: Layer order: impact layer should draw over hazard layer
## Problem
the impact layer is drawing under the hazard layer and not visible to the user immediately
the user will print a map of the hazard data and not the impact layer if they do not turn off or reorder layers

eg - this happened for jakarta flood raster on pop raster and padang earthquake raster on pop raster

image

# proposed solution
IIRC - the standard behaviour @easmetz is for impact layers to draw above hazard layers or for the hazard layer to be turned off.
This is most relevant for raster on raster analysis and where the exposure data are population rasters.

I think this is a regression.

Gustry:
Issue https://github.com/inasafe/inasafe/issues/2899

* If `hide the exposure layer` is checked : 
  * The impact layer will be above the hazard layer
  * The exposure layer will be unchecked

* If `hide the exposure layer` is not checked : 
 * The impact layer will be above the hazard layer and the exposure layer

 @Charlotte-Morgan Is-it the expected behaviour ?
For now in InaSAFE, we don't have any option to hide the hazard layer after the analysis, only the exposure layer, so the algorithm will always put the impact layer above the hazard.

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -1267,7 +1267,7 @@ class Dock(QtGui.QDockWidget, FORM_CLASS):
 
         return impact_function
 
-    def add_above_layer(self, existing_layer, new_layer):
+    def add_above_layer(self, new_layer, *existing_layers):
         """Add a layer (e.g. impact layer) above another layer in the legend.
 
         .. versionadded:: 3.2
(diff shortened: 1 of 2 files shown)


ismailsunni:
LGTM

Charlotte-Morgan:
Looks great to me.

Issue #2765, opened 2016-05-05
Using IF metadata to store the styles


Gustry:
### Problem

Currently, the style for each IF is stored in the impact layer (SafeLayer).
The code is written in the main function of the IF.
Some IF don't use the colour code.
### Proposed Solution

I would like to set the style in the IF metadata because I think that the style is not very appropriate in the `run` method of the IF. I would like to move it in the metadata section. It will allow us to return a QgsMapLayer instead of a SafeLayer from the IF.
The style will be a property in the IF and the dock will be able to read the style.

Moreover, we can create a list of colours. For now, every IF defines their owns colours (affected, not affected, dry, wet ...) We can make them more uniform across IF.

CC : @timlinux @ismailsunni

timlinux:
Yeah I really like this idea. Maybe it is also good to transfer these values to the impact layer xml metadata or to the impact_data so that if the data is provided without a style file we can actually generate the QGIS style again. I can see use cases where for example we want to generate a table with histograms as per Martin's proof of concept and we want the bars to match the colours assigned to the style of the layer classes.

Gustry:
We can now close this ticket with inasafe v4.

PR #2914, opened 2016-06-07
Review the legend and refactoring the legend in the metadata

Gustry:
This PR removes 'by each hazard zone' in the legend. See #https://github.com/inasafe/inasafe/issues/2765
And I made a refactoring to put these parameters in the metadata file instead of the IF itself.

Soon, we will be able to remove the Vector layer from the return statement.

Can you review it @ismailsunni or @lucernae ? Thanks

### DIFF
--- a/safe/impact_functions/earthquake/itb_bayesian_earthquake_fatality_model/metadata_definitions.py
+++ b/safe/impact_functions/earthquake/itb_bayesian_earthquake_fatality_model/metadata_definitions.py
@@ -80,7 +80,7 @@ class ITBBayesianFatalityMetadata(ITBFatalityMetadata):
                    'a Bayesian approach. Submitted for Journal of the '
                    'Geological Society')  # FIXME
             ],
-            'map_title': '',
+            'map_title': 'Earthquake impact to population',
             'legend_title': '',
             'legend_units': '',
             'legend_notes': '',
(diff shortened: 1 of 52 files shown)


safe/impact_functions/earthquake/itb_bayesian_earthquake_fatality_model/metadata_definitions.py
```diff
@@ -80,6 +80,11 @@ def as_dict():
                    'a Bayesian approach. Submitted for Journal of the '
                    'Geological Society')  # FIXME
             ],
+            'map_title': '',
```
ismailsunni: Perhaps we can take the value from the ITB Earthquake Fatality IF.

safe/impact_functions/earthquake/pager_earthquake_fatality_model/metadata_definitions.py
```diff
@@ -72,6 +72,11 @@ def as_dict():
             'output': '',
             'actions': '',
             'limitations': [],
+            'map_title': '',
```
ismailsunni: Perhaps we can take the value from the ITB Earthquake Fatality IF. Same as previous comment

ismailsunni:
LGTM, great...

PR #2918, opened 2016-06-08
Fix #2898 Handle special character in XML

### LINKED ISSUE #2898: Handling Special Character for XML
### Problem
See problem #2886 that caused by trying to write special character (`&`) to xml file.

### Expected Behaviour
We should handle all special character properly before writing to xml or json.

### Environment
* InaSAFE : 3.4
* QGIS : All
* OS : All
* Data used : With special character

### Proposed Solution
Alternatives:
- Catch if there is a special character, and ask user to change it.
- Catch and save it with escape character.

This will superseded issue #2886.

ismailsunni:


Gustry:
LGTM

PR #2935, opened 2016-06-09
fix #2925 set the layer visible when we add it the map canvas

### LINKED ISSUE #2925: Empty report map when using IFCW on volcano on buildings
### Problem
After running IFCW for volcano on buildings using [http://docs.inasafe.org/en/training/socialisation/run_other_hazards.html](url). The report is generated fine but the map is blank because the resultant  layer are not toggled on by default. I have my QGIS settings as not to draw any layers added.

Custom qgis configuration.

setup

Layers added are not on by default.
result

Generated report contains blank map.

empty

### Expected Behaviour
Generate report with map visible.

### Environment
* InaSAFE :  3.14
* QGIS :  2.14.1
* OS :  Ubuntu 16.04
* Data used :
  * Exposure :  `Other Hazard/Sinabung/02 Exposure Data/Sinabung_buildings_WGS84.shp`
  * Hazard :  `Other Hazard/Sinabung/01 Hazard Data/Sinabung_Hazard_Map_2015_WGS84.shp`
  * Aggregation :

### Proposed Solution
Overwrite any custom qgis settings and add any layers as toggled on in QGIS when using any inasafe process.

Gustry:


ismailsunni:
LGTM

PR #2950, opened 2016-06-13
Faster impact tsunami on roads

timlinux:


safe/impact_functions/inundation/tsunami_raster_road/impact_function.py
```diff
@@ -95,8 +99,12 @@ def _raster_to_vector_cells(raster, ranges, output_crs):
 
     # prepare coordinate transform to reprojection
     ct = QgsCoordinateTransform(raster.crs(), output_crs)
-
     rd = 0
+    # Hack because QGIS does not return the
+    # nodata value from the dataset properly in the python API
+    dataset = gdal.Open(raster.source())
+    no_data = dataset.GetRasterBand(1).GetNoDataValue()
+    dataset = None  # close the dataset
```
Gustry: Why not using `del dataset` ?

safe/impact_functions/inundation/tsunami_raster_road/impact_function.py
```diff
@@ -154,6 +167,9 @@ def _raster_to_vector_cells(raster, ranges, output_crs):
                 features = []
     # Add the latest features
     vl.dataProvider().addFeatures(features)
+    # For debugging
+    # from qgis.core import QgsMapLayerRegistry
+    # QgsMapLayerRegistry.instance().addMapLayer(vl)
```
Gustry: These lines can be rewritten easily. If we let these kind of lines, we will find a lot in a few months/years.

safe/impact_functions/inundation/tsunami_raster_road/impact_function.py
timlinux: I followed the goal coding guide for this

safe/impact_functions/inundation/tsunami_raster_road/impact_function.py
timlinux: Agreed - will will delete it

safe/impact_functions/inundation/tsunami_raster_road/impact_function.py
Gustry: Ok. It's not very pythonic :)
I just read http://gis.stackexchange.com/questions/80366/why-close-a-dataset-in-gdal-python but it's not the coding guide

PR #2962, opened 2016-06-17
Added hooks for other plugins

mbernasocchi:
This allows other plugins tohook extra functionality on top of inaSAFE. @timlinux what do you think of the idea?

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -1145,6 +1145,7 @@ class Dock(QtGui.QDockWidget, FORM_CLASS):
             of the web view after model completion are asynchronous (when
             threading mode is enabled especially)
         """
+        self.dock_accept_begin_hook()
         self.enable_signal_receiver()
         try:
             self.enable_busy_cursor()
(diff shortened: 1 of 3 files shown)


timlinux:
Hi @mbernasocchi 

Interesting idea. I wonder if it would not be better to implement this kind of logic in the impact function base class though? We are trying to bit by bit move logic away from dock and into the IF base so that the behavioural logic is all governed in the IF (and thus portable between web, desktop dock, desktop wizard, command line, celery etc. implementations).


We already have https://github.com/inasafe/inasafe/blob/develop/safe/impact_functions/base.py#L764 etc. so we could add more standard logic there if needed...

Also in terms of implantation style, I think it might be nicer to do something like:

* create a callback registry singleton object (I know @wonder-sk hates singletons but....)
* support a pattern like this 

```python 
def before_run_foo():
    print "Hello world"

def after_run_foo():
    print "Goodbye world"

registry = safe.callback_registry()
registry.add_preparation_callback(before_run_foo)
registry.add_completion_callback(after_run_foo)
```

Then the IF calls the callbacks in sequence (and maybe with a self reference so you can peek into the state of the IF) and they can implement whatever logic you like.

That way we have no nasty monkey patching (which makes the code hard to read) and something that is usable in many different contexts.

What do you think?

timlinux:
I am closing this as it is no longer relevant to our new architecture.

PR #2965, opened 2016-06-20
Only test tool package.

ismailsunni:
Don't merge it, for testing purpose only.


Gustry:
I though you were going to squash these commits before the merge :)

ismailsunni:
@Gustry I was. But, I think I was too happy to make it pass, and forgot to do it :D

PR #2974, opened 2016-06-21
select test package

Gustry:

### DIFF
--- a/safe/plugin.py
+++ b/safe/plugin.py
@@ -38,7 +38,14 @@ from PyQt4.QtCore import (
     Qt,
     QSettings)
 # noinspection PyPackageRequirements
-from PyQt4.QtGui import QAction, QIcon, QApplication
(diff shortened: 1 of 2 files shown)


timlinux:
Why don't we have a list box with a default entry of 'all tests' and then a list of all other tests that we can just select the one we want from? I think it is more sustainable and soft coded than having one action per package.

Gustry:
> I think it is more sustainable and soft coded than having one action per package.

Sorry, I didn't show what happen when we click `Select package`, there isn't one action per package : 
anim

We can type the python package we want to inspect. It should be in the python path, we can even launch the test suite from another plugin :) Your input is saved in the QSettings.

Of course I would like something more userfriendly, where we can select which tests we want, but I think this is convenient enough for this first step.

ismailsunni:
LGTM

Issue #2945, opened 2016-06-10
MetadataReadError while loading a layer


Gustry:
### Problem
I opened the file `West Sumatera/01 Hazard Data/Padang_EQ_2009_WGS84.tif` from `data.inasafe.org` : 

```
An error has occurred while executing Python code: 

MetadataReadError: No keyword version found. Metadata xml file is invalid. Layer uri: /home/etienne/kartoza_dev/inasafe/dataset/West Sumatera/01 Hazard Data/Padang_EQ_2009_WGS84.tif Keywords file: True keywords: license: Free use with accreditation title: InaSAFE analysis result url: PyQt4.QtCore.QUrl(u'http://inasafe.org') organisation: InaSAFE.org abstract: InaSAFE analysis result date: 2015-07-27 00:00:00 email: info@inasafe.org 
Traceback (most recent call last):
  File "/home/etienne/.qgis2/python/plugins/inasafe/safe/gui/widgets/dock.py", line 924, in get_layers
    self.keyword_io.update_keywords(layer, {'title': name})
  File "/home/etienne/.qgis2/python/plugins/inasafe/safe/utilities/keyword_io.py", line 214, in update_keywords
    existing_keywords = self.read_keywords(layer)
  File "/home/etienne/.qgis2/python/plugins/inasafe/safe/utilities/keyword_io.py", line 181, in read_keywords
    return read_iso19115_metadata(source, keyword)
  File "/home/etienne/.qgis2/python/plugins/inasafe/safe/utilities/metadata.py", line 110, in read_iso19115_metadata
    raise MetadataReadError(message)
MetadataReadError: No keyword version found. Metadata xml file is invalid.
Layer uri: /home/etienne/kartoza_dev/inasafe/dataset/West Sumatera/01 Hazard Data/Padang_EQ_2009_WGS84.tif
Keywords file: True
keywords:
license: Free use with accreditation
title: InaSAFE analysis result
url: PyQt4.QtCore.QUrl(u'http://inasafe.org')
organisation: InaSAFE.org
abstract: InaSAFE analysis result
date: 2015-07-27 00:00:00
email: info@inasafe.org
```

### Environment
* InaSAFE : Develop

timlinux:
This should be fixed already?

Gustry:
Thanks @ismailsunni , I can't replicate

PR #2976, opened 2016-06-22
Pass when the metadata is outdated on loading to combo box.

ismailsunni:
Fix for #2945 

The issue comes up because whenever there is a change in the list layer, InaSAFE tries to update the combobox in the dock for the hazard, exposure, and aggregation layer.

So, I just pass if the metadata is out of update (just like layer with no keyword).


timlinux:
LGTM

PR #2988, opened 2016-06-26
Rename test_data_path

Gustry:


timlinux:
I'm +0 on this - do you envisage having a 'relative_test_data_path' function too? If not the absolute prefix doesn't seem to add much value to me (we could just describe that behaviour in the doctstring of the helper...).

I'll leave it to you to decide if you want to merge it or not after taking into account my comments above - but check pep8 as it may have pushed some lines over 80 chars...thanks!

timlinux:
Ah sorry I just realised that you are trying to prevent it being run as a test. +1 from me then!

Gustry:
Sorry, I was just doing a test by changing the function name as you noticed.
`test_data_path` is executed at least 58 times in travis as a test, but this function is not a test ! I should get a counter closer than the tests in docker. I expect something like 517 tests now instead of 575 before.

Gustry:
Travis found 519 tests instead of 575 !

PR #2995, opened 2016-06-29
InaSAFE Headless cumulative PR

lucernae:
Contains cumulative changes from merging InaSAFE 3.4 to InaSAFE Headless packages.


bin/inasafe.py
```diff
@@ -32,11 +33,15 @@
     QgsMapLayerRegistry,
     QgsCoordinateReferenceSystem)
 
+
+from safe.test.utilities import get_qgis_app
+# make sure this line executes first
+QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
```
Gustry: Do you use IFACE and CANVAS outside of our test suite ? (I mean in production)

timlinux: Yeah we should probably not be importing test code into production. Conceivably we could be stripping test code out of production shipped packages.

timlinux:
LGTM thanks @lucernae

bin/inasafe.py
lucernae: I didn't use the variable specifically, but get_qgis_app needs to be called first to initialize QPaint and other Qt related object for the composer to works.
Any solution besides this is highly appreciated.

timlinux:
Can you merge this please @lucernae ? We are supposed to be in feature freeze from 1 July...

timlinux:
LGRTM :-)

PR #2996, opened 2016-06-29
remove print or logs in tests

Gustry:


timlinux:
Just curious - is that for performance or neatness sake? +1 to do the merge when passing.

Gustry:
Hum, let's say for performance issue when we will move to Python 3 :p

PR #3000, opened 2016-06-30
Layer visibility

Gustry:
* Make the layer added by the IFCW visible by default when we add a new hazard/exposure layer.
* When the analysis is finished, the dock is checking where to put the impact layer (above the hazard/exposure), enable the impact layer by default ,.... but these statements were not executed by the wizard. For now, I tried quickly to refactor common functions and I copy/paste some statements from the dock which were missing. I'm going to write a ticket to fix that later.

@NyakudyaA Can you try this PR ? Thanks to notice that the wizard does not have the same behaviour.
It's related to #2925


ismailsunni:
LGTM

NyakudyaA:
@Gustry still the behavior persist that layers are not being toggled on using IFCW

NyakudyaA:
LGTM

PR #3002, opened 2016-07-01
use unicode if the layer name has some accents

Gustry:
I couldn't launch an analysis if a layer has some accents (often in French)


ismailsunni:
LGTM

PR #3003, opened 2016-07-01
Fix old keyword problem

ismailsunni:
This PR fixes a problem when you try to launch keyword wizard on a layer that has old xml keyword or any invalid xml.


Gustry:
Just trying to open the layer included in the test and I got this message in the dock : 

No keyword version found. Metadata xml file is invalid. Layer uri: /home/etienne/kartoza_dev/inasafe/inasafe/safe/test/data/other/invalid_keyword_xml.tif Keywords file: True keywords: license: Free use with accreditation title: InaSAFE analysis result url: PyQt4.QtCore.QUrl(u'http://inasafe.org') organisation: InaSAFE.org abstract: InaSAFE analysis result date: 2015-07-15 00:00:00 email: info@inasafe.org

ismailsunni:
@Gustry updated, can you check it again? Thanks

safe/gui/tools/wizard/test/test_keyword_wizard.py
```diff
@@ -126,6 +126,21 @@ def test_get_missing_question_text(self):
...
+    def test_invalid_keyword_layer(self):
+        layer = clone_raster_layer(
+            name='invalid_keyword_xml',
+            include_keywords=True,
+            source_directory=standard_data_path('other'),
+            extension='.tif')
+
+        # check the environment first
+        self.assertIsNotNone(layer.dataProvider())
+        print layer.source()
```
Gustry: Can you remove the print or comment it ? Thanks

Gustry:
Ok, I got a better error message now. Thanks

I didn't undertsand well the test. It tests only if the datasource is valid. Is-it correct ?

ismailsunni:
@Gustry 
I have removed the print statement.
The test will raise MetadataReadError previously. I have added comment to make it clearer.

Gustry:
Ok, thanks, I could understand the test now.
To make it clear, we could do something like : 
```
try:
    dialog.set_keywords_creation_mode(layer)
except MetadataReadError:
    assertTrue(False, "The exception has been raised")
```

But I'm ok with your version

PR #3011, opened 2016-07-06
display a nice error message if MemoryError

Gustry:
- [ ] Need to check in the wizard if the exception is caught too.

selection_015


timlinux:
LGTM. That looks nice thanks @Gustry I wonder if it is worth letting the user try to run anyway - I think we used to have that option...in some cases the memory check may be very conservative (overestimate) about whether the users has sufficient resources.

Gustry:
On my laptop, I don't get any warning before that I don't have enough memory to run the IF.
So the IF is crashing a few seconds later.
In my case, the memory check under estimate my resources.

ismailsunni:
We have a mechanism for this.
See here the message https://github.com/inasafe/inasafe/blob/develop/safe/gui/widgets/dock.py#L1186

And the memory evaluation : https://github.com/inasafe/inasafe/blob/develop/safe/impact_functions/base.py#L767

Although, I didn't get it lately. May be my memory is to big.

safe/utilities/memory_checker.py
```diff
@@ -166,3 +166,24 @@ def check_memory_usage(buffered_geo_extent, cell_size):
...
+
+
+def memory_error():
+    """Display an error when there is not enough memory."""
+    warning_heading = m.Heading(
+        tr('Memory issue'), **WARNING_STYLE)
+    warning_message = tr(
+        'There is not enough free memory to run this analysis.')
+    suggestion_heading = m.Heading(
+        tr('Suggestion'), **INFO_STYLE)
```
ismailsunni: I believe there is SUGGESTION_STYLE in the styles.py. Perhaps it's more suitable to use that.

ismailsunni:
LGTM

PR #3016, opened 2016-07-11
Remove the building threshold fix #2932

### LINKED ISSUE #2932: Discrepancies in the building report about the threshold
### Problem
Since the PR https://github.com/inasafe/inasafe/pull/2564 we can set the building threshold. In the impact table, inasafe will aggregate every impacted object if there are less than 25 objects into the "Other" group. So if there is less than 25 buildings affected, the value mapping will change on the fly during the analysis for that table.

But the postprocessor (`Buildings closed`) is not aware about this threshold, so the value mapping won't be changed. This bring some discrepancies in the report.

Building layer : 
selection_080

Impact report :
selection_081

### Expected Behaviour
Tables should have the same behaviour : 

### Environment
* InaSAFE : Develop
* Data used :
  * Exposure : A building layer with less than 25 features in one category.

### Proposed solution
* remove the building threshold ? The behaviour to aggregate every groups in "Other" does not fit well with the new value mapping according to me.
* or make the building postprocessor aware about the threshold ? (If the user want to report using the value mapping defined for the layer, he must put "1" in the threshold parameter to avoid the aggregation).

Gustry:
## Before : 
etiquette1

## After : 
etiquette1

### DIFF
--- a/safe/defaults.py
+++ b/safe/defaults.py
@@ -27,7 +27,6 @@ from PyQt4.QtCore import QSettings
 
 from safe_extras.parameters.boolean_parameter import BooleanParameter
 from safe_extras.parameters.float_parameter import FloatParameter
-from safe_extras.parameters.integer_parameter import IntegerParameter
 from safe_extras.parameters.group_parameter import GroupParameter
 from safe_extras.parameters.text_parameter import TextParameter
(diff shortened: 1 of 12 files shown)


timlinux:
LGTM thanks

PR #2845, opened 2016-05-25
Modify structure class mapping. 

Charlotte-Morgan:
New definitions, new order.
#2721

PR #3023, opened 2016-07-12
Refactor how notes are handled

### LINKED ISSUE #2973: Add notes to explain report summary
### Problem
The report summary contains figures which are rounded numbers based on the impact function. There is no documentation that explains that the numbers displayed in the reports  are rounded number to significant digits.

selection_006

inas

### Expected Behaviour
Add a note on `notes and assumptions` explaining to the user that the results that are visible on the result map are rounded numbers.

### Environment
* InaSAFE :  3.5
* QGIS :  2.14.3
* OS : Ubuntu
* Data used : `.qgis2/python/plugins/inasafe/safe/test/data`
  * Exposure : `Roads from OSM`
  * Hazard : `/hazard/tsunami_wgs84.tif`
  * Aggregation : `Entire Area`

### Proposed Solution
Add notes on the report page.

timlinux:
* Fix #2973 - explain about rounding used in presented reports. 
* Fix issue introduced by #2845 which added invalid notes to generic landcover IF. 
* Fixed issue in generic landcover IF which was not inheriting the mixin and thus failing to render the notes properly. 
* Moved floods.shp test data to jk so it intersects with landcover.
* Added README for JSON examples
* Updated all impact functions to include notes helper from impact function base and return a list of notes rather than a dict

Probably we should do a similar thing for exposure specific actions and and also for hazard specific notes and actions.


safe/impact_functions/base.py
```diff
@@ -107,6 +102,12 @@
 LOGO_ELEMENT = m.Brand()
 LOGGER = logging.getLogger('InaSAFE')
 
+__author__ = 'akbargumbira@gmail.com'
```
ismailsunni: Do we now put the "docs" after the import? If so, it should be above the LOGO_ELEMENT in line 102 and we need to do for all files :)

timlinux: Yeah there is some pylon error I think that complains if the imports are not at the top of the file...I am just updating them as I encounter them....

ismailsunni:
LGTM, +1 to merge after fixing the conflict

Issue #2910, opened 2016-06-07
RFC: Thoughts on reporting standards


easmetz:
### Problem
After talking to @Charlotte-Morgan and with some feedback from @timlinux and @Gustry , I have reviewed the current reports for all hazard and exposure data and created a reporting standards document for discussion (see #2840). I have also used information from PERKA 2 and UNISDR, particularly when defining hazard classes.

### Proposed Solution

Everyone to read the google document https://docs.google.com/document/d/1rhipAU4mdykbKQ4Qx4I5XuUwSSd33IY7Iw99tvOQsMo/edit?usp=sharing and give feedback

Thanks!

Charlotte-Morgan:
This ticket is superceded by #2920 the RFC for reporting standards

Gustry:
I'm going to close this one, we have the RFC in 2920.

Issue #2840, opened 2016-05-25
We need a wiki page for InaSAFE reporting standards


Charlotte-Morgan:
## Problem
We have many IF reports that are inconsistent wrt terminology, styles, titles, etc
We reviewed this at the time of the 3.2 release
We did not record our decisions :(

## Proposed solution
We need to document the preferred style for IF reports so that:
- we can agree on a standard way we want to communicate
- we can provide guidance to developers writing new IF on standard report components

## Action
- review existing reports
- identify what is good, what is not so good
- develop guidelines for discussion / consultation
- update reporting style to match guidelines
- write a wiki page

## Useful links
DMI 3.4 test reports as PDF - current status of reports 
https://drive.google.com/open?id=0B-d6psrlP1qWNXppcjVPWDR1V0U

previous review of terminology (3.2)
https://docs.google.com/spreadsheets/d/1dnR60NPtkemkUqKxB_mhNrNCgd_zQ4-3LoCCEppxVeE/edit#gid=0

Current human interface doc
https://github.com/inasafe/inasafe/wiki/human-interface-guidelines

Blank test report
https://docs.google.com/spreadsheets/d/1TU4rZZjEFN5bcvu9tmyIOhMfR3zUZsH_biE2mdj7DdI/edit#gid=1838591513 

## cc
@fredychandra @samnawi @felix-yew @ivanbusthomi @timlinux @Gustry @ismailsunni

easmetz:
I have created a google doc for the draft reporting standards wiki (https://docs.google.com/document/d/1rhipAU4mdykbKQ4Qx4I5XuUwSSd33IY7Iw99tvOQsMo/edit?usp=sharing). I will put it in the gitter chat when it is ready for discussion/feedback.

easmetz:
Also @Charlotte-Morgan I can't access the previous review or the blank test report at the moment so I have requested access. When I click on the DMI 3.4 test reports as PDF link it takes me to my blank Drive dashboard. This is the same problem we had last week - could you please re-share the link? Thanks!

Charlotte-Morgan:
hi @easmetz, thanks for the feedback, I have fixed the links above - I hope it works now.

Charlotte-Morgan:
@easmetz 
Please make your document above open for comment now, then we can add any existing standards.

Here are some thoughts on what needs to go in the standards:
**Map**
Title
Legend order
Legend colours
Marginalia

**Report**
General report
Alignment 
Headings
Units

[...]

PR #3027, opened 2016-07-12
Reporting standards

timlinux:
Hi @easmetz and @Charlotte-Morgan 

Here is my first pass at creating the standards doc. It is written in markdown (same system we use for writing tickets) so it should be familiar. There is one caveat and that is that we cannot drag and drop images into the document like we do in a ticket. Quick workaround for that is to open an empty ticket, drag and drop an image into , the copy the created URL to your clipboard and paste it into the document.

In terms of workflow, I am proposing that we make a series of pull requests, covering one section of the document each. In my first mockup, I have synthesised content for the title for example.

Once the document has been fleshed out a bit, we can then start to take on PR's if we need to tweak the document as a whole.

We can apply these changes easily and immediately to the develop branch since they should not impact on any code so there should be low overhead in terms of test maintenance.

I'm happy to give a more detailed walk through e.g. on Skype if you need help to get the editing workflow right.

By the way, you can preview the document as it currently stands here:

https://github.com/timlinux/inasafe/blob/cd22b11a7e7720961593d9cbdd5ed8574b775621/docs/reporting-standards.md

Regards

Tim


Charlotte-Morgan:
hi @easmetz - @timlinux has made this pr with his idea for a better way to document the standards. Ie this would be the place to take the consolidated ideas from the RFC #2910 and would replace the idea of a wiki page #2840.  It would be great if you can take a look and see if this makes sense and if you think it would be easier to implement. Implementation of this approach would be a key task for you.
## cc
@RikkiWeber

Charlotte-Morgan:
Hi @timlinux - this looks like it will work to me.
Just to confirm my understanding:
-  all the content for the standards will be placed in a single mark down document as per the link above
- graphic elements should be loaded as images or entered as tables
- where should images be placed (I'm thinking of the docs work flow here)

timlinux:
Hi @Charlotte-Morgan 

> all the content for the standards will be placed in a single mark down document as per the link above

Correct

> graphic elements should be loaded as images or entered as tables

 Correct

> where should images be placed (I'm thinking of the docs work flow here)

There are two options:

1) Commit the images into the docs folder (preferably in a subfolder called images)
2) Drag and drop the image into an empty ticket, copy the resulting markdown link that gets made for the image and paste it into the standards doc, then close the empty ticket without saving.

Option 2 is probably simpler to use.

easmetz:
This looks great - can't wait to get into it! I think this structure will work.

timlinux:
Thanks @easmetz - I'm going to take your comment above as a 'Looks good to me!' and go ahead and merge this so we can get cracking! :-)

PR #3029, opened 2016-07-13
refactoring report templates,merging child class buildings and roads

Gustry:
I merged the building and road report into a single class. It's now exactly the same report.


safe/impact_template/road_building_report_template.py
```diff
@@ -41,14 +41,80 @@ def __init__(
...
+            row.add(m.Cell(format_int(int(category[1])), align='right'))
+            # For value field, if existed
+            if len(category) > 2:
+                row.add(m.Cell(format_int(int(category[2])), align='right'))
+            table.add(row)
+        message.add(table)
+        return message
+
+    def format_impact_table(self):
+        """Breakdown by building type.
```
timlinux: Is it still building type?

safe/impact_template/road_building_report_template.py
```diff
@@ -41,14 +41,80 @@ def __init__(
         :param impact_data: Dictionary that represent impact data.
         :type impact_data: dict
         """
-        super(AbstractRoadBuildingReportTemplate, self).__init__(
+        super(RoadBuildingReportTemplate, self).__init__(
```
timlinux: What happens if we need another exposure type (Foo) in the future? Will we make it RoadBuildingFooReportTemplate ?

safe/impact_template/test/test_road_building_report_template.py
```diff
@@ -20,23 +20,25 @@
...
 
 TEMP_DIR = temp_dir(sub_dir='test/template')
 TEST_DIR = os.path.dirname(__file__)
 
-JSON_FILE = os.path.join(TEST_DIR, 'data', 'road_impact.json')
+ROAD_JSON_FILE = os.path.join(TEST_DIR, 'data', 'road_impact.json')
+BUILDING_JSON_FILE = os.path.join(TEST_DIR, 'data', 'building_impact.json')
 
 
 class TestRoadReportTemplate(unittest.TestCase):
```
timlinux: TestBreakdownReportTemplate maybe better?

timlinux:
LGTM Gustry but maybe we should remove all exposure specific naming in these classes e.g. use generic BreakdownReport naming?

Gustry:
@timlinux I renamed the base class to GenericReportTemplate. Roads and buildings are using this class and Landcover, population, and polygon people have the shortest child class.

timlinux:
LGTM

PR #3039, opened 2016-07-14
Fix #2746 remove entire area

### LINKED ISSUE #2746: Repeated figures in some reports
### Problem
In the example below, the purple rectangles are exactly the same. In addition, if you don't use an aggregation layer, the red rectangles are repeated too.
selection_072

### Expected Behaviour
If we have less tables, it will make the report more readable for the final user.

### Proposed Solution
I wonder why do we have the first purple table and if we can remove it in that case.
For the red table 2, this one becomes useful only with we use an aggregation layer. If we don't use an aggregation layer, I wonder if we can skip it.

ismailsunni:
Contents
- Remove "Entire area" per #2746 
- Little bit code cleaning.


timlinux:
LGTM thanks

timlinux:
Looks good to me! thanks @ismailsunni

safe/impact_template/generic_report_template.py
```diff
@@ -62,6 +64,7 @@ def __init__(
             raise MissingImpactReport
 
         self.impact_data = impact_data
+        self.exposure = impact_data.get('exposure')
```
Gustry: Can't it go to the parent class ?

PR #3040, opened 2016-07-14
Fix #2966 - support more formats than just png for custom map north arrow and organsiation logo

### LINKED ISSUE #2966: Single image type when defining custom options in inasafe options.
### Problem
When choosing custom template options  it only allows use of the image  type  `png`  for  either north arrow or custom logo.

image_type

### Expected Behaviour
Allow different types of images to be used when defining custom options in inasafe.

### Environment
* InaSAFE : 3.5
* QGIS : 2.14.3
* OS : Ubuntu 16.04
* Data used : 
  * Exposure : 
  * Hazard : 
  * Aggregation :

### Proposed Solution
Add ability to use various image types for custom options. Examples of image types that could be supported are `jpeg`, `svg`

timlinux:
[Number_of_people_affected_in_each_class.pdf](https://github.com/inasafe/inasafe/files/363410/Number_of_people_affected_in_each_class.pdf)

### DIFF
--- a/safe/report/impact_report.py
+++ b/safe/report/impact_report.py
@@ -18,16 +18,13 @@ __copyright__ = ('Copyright 2012, Australia Indonesia Facility for '
                  'Disaster Reduction')
 import os
 import logging
-
-# noinspection PyUnresolvedReferences
 from qgis.core import (
     QgsComposition,
     QgsRectangle,
     QgsMapLayer,
-    QgsComposerPicture,
     QgsComposerHtml,
     QgsComposerFrame)
-
+# Whoaa this is ugly can we get rid of it?
 try:
     # noinspection PyUnresolvedReferences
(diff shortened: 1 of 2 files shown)


safe/report/impact_report.py
```diff
@@ -18,15 +18,13 @@
...
-
-# noinspection PyUnresolvedReferences
 from qgis.core import (
     QgsComposition,
     QgsRectangle,
     QgsMapLayer,
     QgsComposerHtml,
     QgsComposerFrame)
-
+# Whoaa this is ugly can we get rid of it?
```
ismailsunni: I think yes, it should be put in the import statement above it.

timlinux: I'm going to leave it for now until I understand why it was added there - can fix that another time...

PR #3041, opened 2016-07-14
Ash on places IF

Gustry:
* Airports

airports

* Places without population

places

* Places with population

population

With the wizard, we should check that the population is not empty and we can't have a population field with an airport.


timlinux:
LGTM - as a separate request, I think we should not show the summary table if there is only one record - what do you think?

PR #3044, opened 2016-07-18
fix #2911 - PetaJakarta layer style not working

### LINKED ISSUE #2911: Symbology of Peta Jarkarta Flood layer is not clearly visible
### Problem
After downloading a flood layer from Peta Jakarta the symbology of the layer makes it not clearly visible especially when you have a raster layer in the layer list.  In the properties of the flood layer the column that is used to style the layer is missing. Also the downloaded layer is not displayed by default.

jk

Flood layer generated is not toggled on.
jk1

Flood layer is not clearly visible.
jk3
Missing column used to style layer.
jk4

### Expected Behaviour
Make layer visible by changing the transparency levels of the symbols. 
Make sure the style is applied correctly and the column that is used to style the layer is visible.
Make the flood layer default as on in QGIS.

jk6

### Environment
* InaSAFE : 3.14
* QGIS : 2.14.1
* OS :  Ubuntu 16.04
* Data used :
  * Exposure : 
  * Hazard : Flood layer downloaded from Peta Jakarta
  * Aggregation :

### Proposed Solution
Symbolize the layer  correctly and allow the layer to be visible with respect to other layers.

timlinux:


Gustry:
LGTM

PR #3049, opened 2016-07-18
Let there be citations. 

timlinux:
Added formal citation structures to definitions and all impact function metadata. This is not exposed in any reports yet. Where citations do existing (in old citations form or as comments) they were added into the citations structure. defaults.py has not been addressed in this PR.


Gustry:
LGTM

PR #3053, opened 2016-07-21
remove clear_further_steps in exposure places classify

Gustry:
@ivanbusthomi I made a fix for your problem about the value mapping


ivanbusthomi:
Hi @Gustry, it looks good. thanks!

PR #3061, opened 2016-07-22
Add aggregation for places

Gustry:
@ivanbusthomi Can you try this PR about aggregation of places ?


ivanbusthomi:
Hi @Gustry ,
This is the map result that I got,
map
and this is the city name that appears in the report
affected

apparently, some cities are missing.

timlinux:
We can queue this for 3.5.1

Gustry:
In InaSAFE v4,  we have a new architecture for aggregation. I'm going to close this PR.

Issue #2272, opened 2015-08-28
Definitions: we need to update definitions


Charlotte-Morgan:
We need to update definitions so that they are helpful to users. 
Where possible we should use published international definitions for disaster related terms.
I propose updating definitions using this source http://www.unisdr.org/we/inform/terminology
unless anyone objects

# cc
@maning @assefay @adissadis @iyan31 @timlinux @vdeparday

Charlotte-Morgan:
hi @easmetz  - as discussed.
handing this to you. Please check our definitions and if a UNISDR definition exists, we should update and reference the source.

I think the file is definitions.py
#cc @timlinux  - please confirm the file

Charlotte-Morgan:
hi @easmetz  - this is the file you need
C:\Users\AIFDR_ADMIN\.qgis2\python\plugins\inasafe\safe\definitions.py

timlinux:
I think this is already done in 3.4? @easmetz can you close this ticket if you are done?

easmetz:
I am continuously updating definitions.py definitions as we go along and discover new words that need to be defined. There's a few more to go, I'd like to leave this open. No idea when this will be finished. I'd like to say by 3.5 but more definitions keep popping up.

easmetz:
This is a good updated UNISDR terminology paper. http://www.preventionweb.net/files/45462_backgoundpaperonterminologyaugust20.pdf

Charlotte-Morgan:
@easmetz - is this ticket and pull requests redundant now?

easmetz:
They may be more definitions we need to add in the future but this ticket is done in terms of Version 3.5 since the pull requests have been merged.

Issue #2920, opened 2016-06-08
RFC Reporting Standards


easmetz:
# Problem

Currently map and report outputs from InaSAFE are inconsistent. Titles, legend content, legend titles, terminology and many other aspects of the reports vary from report to report and IF to IF. We need to review the current reports for all hazard and exposure data and create reporting standards to help these products be more consistent.
# Duration

22 June 2016.
# Current state

Changes being implemented
# Proposers
- Libby Metz ( @easmetz )
  with input from:
- Tim Sutton ( @timlinux )
- Charlotte Morgan ( @Charlotte-Morgan )
- Etienne Trimaille ( @Gustry )
# Proposal
### Map title and content

**Map title**
Capitalise first letter only
Title should include identifying information for hazard and exposure data e.g. 
- People in need of evacuation due to tsunami
- People affected by earthquake shaking
- Buildings flooded by tsunami
- Roads inundated by tsunami
- Land cover affected by tsunami

Do not mention “by each hazard zone” in the title

For Generic hazard the following titles should be standard:
- People affected
- Buildings affected
- Roads affected
- Land cover affected

**Map image/content**
- The only layers that should be visible are the context data (i.e. administrative boundaries) and the impact layer.
- If the user wants to show the hazard data as well as the impact data, they can turn it on themselves.

**Record of votes for Map title and content section**

| Vote | Name |
| --- | --- |
### Legend

![image](https://cloud.githubusercontent.com/assets/16660099/16068549/0b08b9a4-330a-11e6-9c74-3e0b772253c7.png)

**Legend title**
Standard options for generic:
- Estimated people affected
- Estimated buildings affected
- Estimated roads affected
- Estimated land cover affected

_Do not add “by each hazard zone” to the end of the title. (Alternatively, if this is done it should be consistent across all maps.)_ 

**Legend content**
All layers shown on the map image should be shown in the legend.

_All legend entries should take the following form:_
Words in the legend directly next to the legend colour, followed by numbers in brackets e.g. like this:
![image](https://cloud.githubusercontent.com/assets/16660099/16068707/6a089608-330b-11e6-93e3-7c0508d4f0ce.png)

_Rounding_
- Population - whole numbers only
- Flood/tsunami depths - to 1 decimal place
- Land Cover - to 1 decimal place
- Roads - to 0 decimal place

[...]

PR #3067, opened 2016-07-26
Partfix #2272

easmetz:
Inserted definitions of **displaced** and **evacuated** into population in buildings/population notes (as they are the same). I am not completely sure if I have cited correctly as the link is too long for the space available. The definitions are slightly inconsistent in their wording as I have used the exact UNISDR definitions. I think it would be wise to edit them for consistency of wording (ie persons vs people, etc) but I'm open to comment.

The notes and assumptions now look like this:
image

### DIFF
--- a/docs/reporting-standards.md
+++ b/docs/reporting-standards.md
@@ -126,14 +126,14 @@ _Disclaimer_: InaSAFE has been jointly developed by the Indonesian Government-BN
 
 # Reports
 
-* Titles and text left aligned.
-* Numbers right aligned.
+* Titles, body text, table text, table text row heading and table text column headings for text content left aligned.
+* Numbers and table text column headings for number content right aligned.
 
-## Question
+## Analysis question
 
-The question appears at the beginning of each report and identifies and describes 
-the purpose of the analysis. The question should reflect the terminology used in the
-title. Using this table as a reference:
+The analysis question appears at the beginning of each report and identifies and describes 
+the purpose of the analysis. It should add more detail to the existing map title. 
+The question should reflect the terminology used in the title. Using this table as a reference:
 ![image](https://cloud.githubusercontent.com/assets/16660099/17164091/0ad0b44a-53ef-11e6-93f2-0c0241d15b7d.png)
 
 We propose the questions should use the following syntax:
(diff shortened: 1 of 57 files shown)


Charlotte-Morgan:
Hi @easmetz a couple of comments from me but maybe @timlinux can advise.

- I think its ok to leave a hyper link longer than the 80 characters ie no need to wrap it.
- I think the word people is better in this case than persons
- The notes for displaced should only be added to impact functions that mention displaced people
- The notes for evacuated should only be added to impact functions that mention evacuated people

this links back to #2920 regarding when we should use the words evacuated and displaced. Now we have definitions we can make some recommendations in the standard

not sure why the test has failed

easmetz:
I have now changed "persons" to "people" in line with your comments @Charlotte-Morgan

timlinux:
hi @easmetz and @Charlotte-Morgan 

Hi @easmetz a couple of comments from me but maybe @timlinux can advise.

> I think its ok to leave a hyper link longer than the 80 characters ie no need to wrap it.

Yes for Markdown, no need to adhere to 80 chars for links. For readability and version difference visualisation it is still generally a good idea to put line breaks in at around 80 chars where possible.

> I think the word people is better in this case than persons

+1

> The notes for displaced should only be added to impact functions that mention displaced people

+1

> The notes for evacuated should only be added to impact functions that mention evacuated people

+1 

> this links back to #2920 regarding when we should use the words evacuated and displaced. Now we have definitions we can make some recommendations in the standard

> not sure why the test has failed

That are all passing now it seems...

Charlotte-Morgan:
Hi @easmetz this is taking shape nicely.  
A few more suggestions from me that maybe @timlinux can confirm or dismiss.
- [ ] This document is the 'standard' not the RFC which was the proposal. I would encourage you to use 'stronger' language to describe the standard. ie Instead of **we propose the title** ... we could write **The title should have the following syntax**
- [ ] I would prefer that this document should have paragraphs and sentences with pictures, tables and dot point lists. It should avoid being a list of notes, we already have that in the RFC.

timlinux:
Hi @easmetz and @Charlotte-Morgan 

> This document is the 'standard' not the RFC which was the proposal. I would encourage you to use 'stronger' language to describe the standard. ie Instead of we propose the title ... we could write The title should have the following syntax

Agreed

> I would prefer that this document should have paragraphs and sentences with pictures, tables and dot point lists. It should avoid being a list of notes, we already have that in the RFC.

Agreed - as a minimum, a stranger coming to the project should be able to make sense of the reporting standards without needing a lot of *a priori* knowledge of the project.

easmetz:
I have separated these two out into two separate pull requests:

definitions.py changes = #3077
reporting standards changes = #3078

PR #3071, opened 2016-07-26
Fix link (update)

kant:


README.md
```diff
@@ -39,7 +39,7 @@ Quick Installation Guide
 ========================
 
 You first need to have [QGIS](http://qgis.org/) installed. Grab your free
-copy from [the QGIS download page](http://download/qgis.org).
+copy from [the QGIS download page](http://qgis.org/en/site/forusers/download.html).
```
timlinux: I think it will be better to use http://download.qgis.org as it will always redirect to the correct page and is also locale aware I think (versus the static link you used which is en only). So maybe just replaces the ``/`` with a ``.``.

Issue #3057, opened 2016-07-21
We need a full suite of report actions and notes


Charlotte-Morgan:
## problem

In our review of the wallpaper, we noted that in some of our impact reports we are missing information about the hazard and exposure data in the notes, and actions for the action list.
## proposed solution

we should populate definitions.py with actions and notes for all exposure definitions
we should populate definitions.py with actions and notes for all hazard definitions

In some cases notes are included in the impact functions and it may not be necessary to add notes for all hazards or exposure elements.

The aim is to ensure that in every impact report there are some notes about the hazard and exposure element.
## contributors

@fredychandra @easmetz @charlotte-morgan

Charlotte-Morgan:
hi @easmetz - I've assigned this too you. I hope we can work on this with the team in Jakarta next week

easmetz:
| Hazard | Notes/Actions |
| ------ | -------- | 
| Generic | <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Earthquake | <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Flood | <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Volcanic Ash |  <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Tsunami |  <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Volcano |  <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |

| Exposure | Notes/Actions |
| ------ | -------- | 
| Land cover | <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| People in buildings | <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Population | <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Road |  <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Structure |  <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |
| Place |  <ul><li>[ ] Notes</li><li>[ ] Actions</li></ul> |

Charlotte-Morgan:
Great list @easmetz; we can skip people in buildings for this week if you like :)
Generic hazard might also be difficult

easmetz:
[...]

PR #3077, opened 2016-07-27
Definitions partfix #2272 to replace #3067

easmetz:
Replacement pull request for #3607 after that proliferated and became changes to definitions.py and reporting-standards. This pull request is for the insertion of the displace people and evacuated people definitions into definitions.py only.

### DIFF
--- a/safe/definitions.py
+++ b/safe/definitions.py
@@ -517,16 +517,12 @@ exposure_land_cover = {
     ]
 }
 
-exposure_people_in_building = {
-    'key': 'people_in_building',
-    'name': tr('People in buildings'),
+exposure_population = {
+    'key': 'population',
+    'name': tr('Population'),
     'description': tr(
-        'The <b>people in buildings</b> exposure data is an experimental '
-        'data set that assigns the population of a specific administrative '
-        'area to the buildings with a residential function in that area.'
-        '<p>The process of assigning people to buildings assumes that all '
-        'people and buildings in the area are mapped. There are no InaSAFE '
-        'impact functions that use this exposure data yet.</p>'),
+        'The <b>population</b> describes the people that might be '
+        'exposed to a particular hazard.'),
     'notes': [  # these are additional generic notes for people - IF has more
         tr('All values are rounded up to the nearest integer in order to '
            'avoid representing human lives as fractions.'),
@@ -536,16 +532,7 @@ exposure_people_in_building = {
            'nearest 10 persons if the total is less than 1,000; nearest 100 '
            'people if more than 1,000 and less than 100,000; and nearest '
            '1000 if more than 100,000.'),
-        tr('Displaced people are people who, for different reasons and '
-           'circumstances because of risk or disaster, have to leave their '
-           'place of residence.'),
-        tr('Evacuated people are people who, for different reasons and '
-           'circumstances because of risk conditions or disaster, move '
-           'temporarily to safer places before, during or after the '
-           'occurrence of a hazardous event. Evacuation can occur from '
-           'places of residence, workplaces, schools and hospitals to other '
-           'places. Evacuation is usually a planned and organised '
-           'mobilisation of persons, animals and goods.')
+
     ],
     'actions': [  # these are additional generic actions - IF has more
         tr('Which group or population is most affected?'),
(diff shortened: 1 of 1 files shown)


Charlotte-Morgan:
Hi @easmetz - i think the detail you have added ie the text strings are fine but I am not sure they are in the right place.

I think we need to separate the exposure_population from exposure_people_in_buildings
we need to put details in for exposure_population.  @timlinux - please confirm
we do not need details for exposure _people_in_buildings because we do not use it in any IF now
this means most the content needs to move from one definition to the other.

The notes are generic for all IF with people as the exposed element
I think that the definition for  evacuated people should be in the notes for exposure_people only if the term 'evacuated' is used in the report - ie its what happens to affected people.
I think the definition for displaced people is impact function specific and possibly only relevant to the earthquake IF - see the advice in #3057

Charlotte-Morgan:
@easmetz - as per #3079
please define new concepts
go ahead and add details to concept_population or concept_people
write a separate ticket asking @timlinux to redirect population notes in all ** impact on people Impact functions to the concept_people and not to concept_people_in_buildings

easmetz:
I have now changed the focus from exposure_people_in_buildings to exposure_population. All IFs should be directed to exposure population instead of exposure_people_in_buildings. I will write a ticket for this.

easmetz:
I have also now inserted two new concepts: displaced and evacuated. I will insert affected once I get feedback on the reporting standards.

easmetz:
I have now added the concept of "affected".

easmetz:
This is also a partfix-3079

safe/definitions.py
```diff
@@ -55,6 +55,26 @@
     'functions. The area may have either <b>low</b>, <b>medium</b>, or '
     '<b>high</b> classification for the hazard.')
 
+concept_affected = tr(
+    '**Affected**: An exposure element (e.g. people, roads, buildings, land '
```
timlinux: Note that we can't use markdown here (so no **). You can use <b></b> html tags but I would suggest to avoid that too wherever possible

safe/definitions.py
```diff
@@ -559,18 +576,24 @@
...
+exposure_people_in_building = {
+    'key': 'people_in_building',
+    'name': tr('People in buildings'),
     'description': tr(
-        'The <b>population</b> describes the people that might be '
-        'exposed to a particular hazard.'),
-    'notes': exposure_people_in_building['notes'],
-    'actions': exposure_people_in_building['actions'],
+        'The <b>people in buildings</b> exposure data is an experimental '
+        'data set that assigns the population of a specific administrative '
```
timlinux: I don't think it makes sense to refer to the 'dataset' (user have their own PIB data which may not be experimental).

The <b>people in buildings</b> exposure data assigns the population of a specific administrative '

safe/definitions.py
```diff
@@ -559,18 +576,24 @@
...
+        'area to the buildings with a residential function in that area.'
+        '<p>The process of assigning people to buildings assumes that all '
+        'people and buildings in the area are mapped. There are no InaSAFE '
+        'impact functions that use this exposure data yet.</p>'),
```
timlinux: There are no InaSAFE impact functions that use this exposure data yet.

safe/definitions.py
```diff
@@ -535,7 +551,8 @@
         tr('Numbers reported for population counts have been rounded to the '
            'nearest 10 persons if the total is less than 1,000; nearest 100 '
```
Charlotte-Morgan: nearest 10 people

Charlotte-Morgan:
hi @easmetz - this is looking good.
Do you want to add a concept_affected as well.
It looks like the citation to UNISDR might be in the wrong place - please check

safe/definitions.py
```diff
@@ -559,18 +576,23 @@
...
-    'name': tr('Population'),
+exposure_people_in_building = {
+    'key': 'people_in_building',
+    'name': tr('People in buildings'),
     'description': tr(
-        'The <b>population</b> describes the people that might be '
-        'exposed to a particular hazard.'),
-    'notes': exposure_people_in_building['notes'],
-    'actions': exposure_people_in_building['actions'],
+        'The <b>people in buildings</b> exposure data assigns the population '
```
timlinux: Take away "The"

PR #3088, opened 2016-07-28
Revised definitions of 'count' and 'density' to provide more clarity for wizard users

easmetz:
fix-3069


Charlotte-Morgan:
LGTM

PR #3102, opened 2016-08-02
fix #2804 - make checking the analysis environment more efficient by not reading the full raster each time you pan or zoom

### LINKED ISSUE #2804: Can you create button to enable or disable InaSAFE: Analysis environment ready
### Problem
Currently, while we doing test to InaSAFE 3.4, there some issue about how InaSAFE work. I'm not good to explain this, but when you enable InaSAFE plugin, the InaSAFE will create/do some task to see if the InaSAFE analysis environment is ready or not to be execute. There will be some of message above map canvas like this:
image
From above message, i'm thinking that after you put some data in and those data already have InaSAFE keyword on this, InaSAFE will try to run some process to check whether those data is ready to run (based on analysis area or maybe the keyword or else)

This process will eat some RAM and while this not really affect when we doing some analysis with vector-vector data, this process really eat RAM for raster data. Here's my screenshot for the amount of RAM used for InaSAFE to checking the analysis environment for Raster data
image

And this is for another screenshot for another raster data:
image

Based on those screenshot, for big raster data like Java population or Sumatera Population from AsiaPop, InaSAFE will need those amount of ram to check the environment.

This become a problem to people who only have small amount of RAM in their laptop (for example only 2GB or 4GB). They will encounter lag or freeze or a long time and sometimes their laptop will crashed or QGIS mini dump.

Actually this problem can be solved by crop the raster layer into small area based on their area of interest but i proposed some of solution to create some option in InaSAFE's Options to enable or disable this feature.

### Expected Behaviour
If they disable the option to "Disable InaSAFE automatically check analysis environment" in InaSAFE's Options the InaSAFE won't perform automatic analysis environment check.

### Environment
Windows 10 64bit 16GB ram
QGIS 2.8.9
InaSAFE 3.4

### Proposed Solution
Create some flag to enable or disable InaSAFE automatic analysis environment check

timlinux:
fix #2804 - make checking the analysis environment more efficient by not reading the full raster each time you pan or zoom 

Ok I have made a fix (for rasters anyway):

Before optimisation: 15s to do a simple pan

qgis

After optimisation: 4s to do a simple pan

qgis

### DIFF
--- a/safe/storage/layer.py
+++ b/safe/storage/layer.py
@@ -47,7 +47,7 @@ class Layer(object):
         # Defaults
         self.sublayer = sublayer
         self.filename = None
-        self.data = None
+        self._data = None  # class private
 
     def __ne__(self, other):
         """Override '!=' to allow comparison with other projection objecs
--- a/safe/storage/raster.py
+++ b/safe/storage/raster.py
@@ -111,7 +111,7 @@ class Raster(Layer):
             # Assume that data is provided as a numpy array
             # with extra keyword arguments supplying metadata
 
-            self.data = numpy.array(data, dtype='d', copy=False)
+            self._data = numpy.array(data, dtype='d', copy=False)
 
             proj4 = self.get_projection(proj4=True)
             if 'longlat' in proj4 and 'WGS84' in proj4:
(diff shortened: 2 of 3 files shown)


timlinux:
@adissadis can you test this please and see if it resolves your issues?

ismailsunni:
LGTM

adissadis:
LGTM

PR #3110, opened 2016-08-04
Support more granular definition of notes 

### LINKED ISSUE #3109: Notes need to be more granular for hazards
### Problem

Notes should be able to cater for the specifics of the hazard data used - continuous vs classified data, single event versus multi-event

### Proposed solution

Support a structure like this in definitions.py:

```
hazard_earthquake = {
    'key': 'earthquake',
    'name': tr('Earthquake'),
    'description': tr(
        'An <b>earthquake</b> describes the sudden violent shaking of the '
        'ground that occurs as a result of volcanic activity or movement '
        'in the earth\'s crust.'),
    'continuous_notes': [  # additional generic continuous hazard notes - IF has more
],
   'classified_notes': [  # additional generic classified notes - IF has more
],

   'single_event_notes': [  # additional generic single event notes - IF has more
],

  'multi_event_notes': [  # additional generic multi event notes - IF has more
],

    'actions': [  # these are additional generic actions - IF has more

    ],
    'citations': [
        {
            'text': None,
            'link': None
        }
    ]
}
```

timlinux:
and remove duplication of content in definitions.py. fix #3109

CC @Charlotte-Morgan

### DIFF
--- a/safe/impact_functions/base.py
+++ b/safe/impact_functions/base.py
@@ -502,6 +502,7 @@ class ImpactFunction(object):
         """
         notes = []
         hazard_name = self.hazard.keyword('hazard')
+
         for hazard in hazard_all:
             try:
                 if hazard['key'] == hazard_name:
@@ -521,7 +522,7 @@ class ImpactFunction(object):
                         notes += hazard['multi_event_notes']
             except KeyError:
                 pass
-        return None
+        return notes
 
     def action_checklist(self):
         """Return the action check list.
--- a/safe/impact_functions/base.py
+++ b/safe/impact_functions/base.py
@@ -514,10 +514,10 @@ class ImpactFunction(object):
                 if self.hazard.keywords['data_type'] == 'continuous':
                     if 'continuous_notes' in hazard:
                         notes += hazard['continuous_notes']
-                if self.hazard.keywords['scenario'] == 'single_event':
+                if self.hazard.keywords['hazard_category'] == 'single_event':
                     if 'single_event_notes' in hazard:
                         notes += hazard['single_event_notes']
-                if self.hazard.keywords['scenario'] == 'multiple_event':
+                if self.hazard.keywords['hazard_category'] == 'multiple_event':
                     if 'multi_event_notes' in hazard:
                         notes += hazard['multi_event_notes']
             except KeyError:
--- a/safe/impact_functions/base.py
+++ b/safe/impact_functions/base.py
@@ -452,14 +452,22 @@ class ImpactFunction(object):
             'notes']
         :rtype: list, None
         """
+        notes = []
         exposure_name = self.exposure.keyword('exposure')
         for exposure in exposure_all:
             try:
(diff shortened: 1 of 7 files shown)


safe/impact_functions/base.py
```diff
@@ -500,14 +500,29 @@ def hazard_notes(self):
...
-                    return hazard['notes']
+                    if 'notes' in hazard:
+                        notes += hazard['notes']
+                if self.hazard.keywords['data_type'] == 'classified':
+                    if 'classified_notes' in hazard:
+                        notes += hazard['classified_notes']
+                if self.hazard.keywords['data_type'] == 'continuous':
+                    if 'continuous_notes' in hazard:
+                        notes += hazard['continuous_notes']
+                if self.hazard.keywords['scenario'] == 'single_event':
```
timlinux: I think this should be 'category' not 'scenario' - need to check...

safe/impact_functions/base.py
```diff
@@ -500,14 +500,29 @@ def hazard_notes(self):
...
+                    if 'single_event_notes' in hazard:
+                        notes += hazard['single_event_notes']
+                if self.hazard.keywords['scenario'] == 'multiple_event':
```
timlinux: as above...

Charlotte-Morgan:
Hi @timlinux - I discussed the need for greater granularity in the notes with the Minions last week.
I think we got to the following point (which is not in the standards):
for hazard data we need notes for single and multiple events
for exposure date we need notes for continuous vs classified data (regardless of whether its raster or vector)
We also discussed the idea from @samnawi that these 'notes' should be attached to the data ie metadata or a new keyword field.  

Looks to me like this discussion should remain open in #2920 and that we maybe have not resolved the best way to handle this yet. 

@easmetz - any thoughts on this one?

easmetz:
@Charlotte-Morgan your ideas sound good. I don't have any extra ideas too add at the moment except that we should probably add notes for aggregation IF as well if the process is not streamlined in our next release. It would be good to have an explanation for why aggregate vs non-aggregate analyses can get different results.

timlinux:
Hi @Charlotte-Morgan and @easmetz 

Ok so may I suggest we keep the granular separation of notes for hazard as already done in this PR (we can just not use the classified /  continuous notes for hazard until we need them) and then I just add the same type of implementation for exposure (only for classified / continuous)?

Charlotte-Morgan:
hi @timlinux - I think that sounds perfect thank you

timlinux:
Ok @Charlotte-Morgan that is done - we just need to actually add some continuous vs classified and single vs multi event  notes to definitions.py so we can see the results nicely. Did you or the minions have any specific ideas of what to add there?

timlinux:
@Charlotte-Morgan btw I am going to go ahead and merge this so that @ismailsunni can do the string updates for translations. I will follow up tweaks with subsequent PR's if needed.

PR #3113, opened 2016-08-05
Fix #2430 tick bullet

### LINKED ISSUE #2430: Minor Tweaks for Reporting
Problem
-----------
From @Charlotte-Morgan's comment in https://github.com/AIFDR/inasafe/pull/2415#issuecomment-145211543
Minor tweaks - if you still have enthusiasm for such things:
- [ ] Flood impact is the only one that quotes values in action list - if its not easy to get the correct value, would it be possible to remove them instead? 
- [ ] Population impact notes: I do not think we need the ".. rounded to nearest integer to avoid representing human lives as fractions"  can we drop this from all notes please. (we round everything and never quote decimals for anything)
- [ ] superscript note for population impact - again i do not think we need this as we do not use it elsewhere for other notes (unless you have just added it)
- [ ] now getting fancy - is it possible to use a tick bullet for action check list (really not essential)
- [x] Please change all references to "Perka 7/2008" to "BNPB Perka 7/2008"

ismailsunni:
Contents:
- Add custom bullet list style, you can add more of course
- Use tick bullet for action checklist.
Example:
screen shot 2016-08-05 at 10 07 58

### DIFF
--- a/resources/css/InaSAFE.css
+++ b/resources/css/InaSAFE.css
@@ -82,3 +82,18 @@ h5 {
 .caption {
   font-size: 8pt;
 }
+
+/* List styling */
+.customBulletList, .customBulletList ul {
+  margin-top: 0;
+  padding-left: 2em;
+  list-style-type: none;
+}
+.customBulletList li:before {
+  position: absolute;
+  margin-left: -1.3em;
+  font-weight: bold;
+}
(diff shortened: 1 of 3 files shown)


Gustry:
Nice, LGTM ;-)

timlinux:
LGTM

PR #3115, opened 2016-08-05
Fix #2847 no aggregation table

### LINKED ISSUE #2847: The aggregation result not appear in printed PDF.
### Problem
After I, @samnawi and @ivanbusthomi have been analysis the InaSAFE 3.4.0. between some of Hazard data and Exposure data and add with Aggregation data, there is some problem with the result. In the Print Table in PDF, there is no aggregation result. We have some problems when we have printed the PDF result in different hazard that we use it. 
This is one of example from printed PDF InaSAFE analysis result:
image

There is some problems that I have capture it :
image

### Environment
* InaSAFE : 3.4.0.
* QGIS : 2.14.1.
* OS : Windows 7 64-bit
* Data used :
  One of the problem result, i.e. Landslide Hazard (Generic)
  * Hazard : https://www.dropbox.com/s/97usybl1c5rkj7x/01%20Hazard.7z?dl=0
  * Exposure : https://www.dropbox.com/s/do20e70vzlis767/02%20Exposure.7z?dl=0
  * Aggregation : https://www.dropbox.com/s/tryg3qbo2bvi51u/03%20Aggregation.7z?dl=0

### Proposed Solution
Make it happy.

Cc: @Charlotte-Morgan @timlinux @ismailsunni @samnawi @ivanbusthomi

ismailsunni:
Problems:
- We use new keyword for aggregation to mark which field in exposure layer that should be aggregate. Meanwhile, post processor expects another keyword
- The checker is not really robust. Lowercasing one hand but not the other hand.

Fixes:
- Better setup for post processor keyword for building and road.
- Better checker.

### DIFF
--- a/safe/definitions.py
+++ b/safe/definitions.py
@@ -552,7 +552,8 @@ exposure_land_cover = {
         'The <b>land cover</b> exposure data describes features on '
         'the surface of the earth that might be exposed to a particular '
         ' hazard. This might include crops, forest and urban areas. '),
-    'notes': [  # these are additional generic notes for landcover - IF has more
+    'notes': [
+        # these are additional generic notes for landcover - IF has more
         tr('Areas reported for land cover have not been rounded.'),
     ],
     'actions': [  # these are additional generic actions - IF has more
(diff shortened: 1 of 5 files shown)


ismailsunni:
Just a note, we no longer need to use specific field name in the exposure layer to be aggregated. Previously, we need to use type or TYPE, otherwise the empty table will come to you.

felix-yew:
Hi Mas @ismailsunni 

I have tried your PR and use it to analysis the flood raster vs infrastructure vector, see the image below :
image

And use to analysis the flood vector vs infrastructure vector, see the image below :
image

The result is great, this PR can read other column or attribute, besides TYPE. This LGTM :+1:

timlinux:
LGTM

PR #3124, opened 2016-08-09
Translation 3.5

ismailsunni:
If there is no changes in InaSAFE's strings or in transifex, just go a head and merge this.

cc @Gustry @timlinux @lucernae


timlinux:
LGTM thanks @ismailsunni

PR #3131, opened 2016-08-10
Standards: map legends

Charlotte-Morgan:


Charlotte-Morgan:
@timlinux  - merging this without your comments. Standards are a work in progress, more is needed and the whole doc can be refined / reviewed as we implement things. The document can also be edited for consistency of style once all the bits are in.

PR #3159, opened 2016-08-15
Fix #3151 zero impact in IFCW

### LINKED ISSUE #3151: BUG: Volcanic ash on airports should be able to return no results
# problem
Volcanic ash on places (airports) returns an error if no airports are affected - when run from the IFCW.

# error message
image

# proposed solution
It should be acceptable for the impact function to return no results.
The user should get advice to say that no results were returned. I think this should not be an exception.
The same notes that are given to the user in the dock could be used in the wizard.

ismailsunni:


felix-yew:
Hi mas @ismailsunni 

I try your PR and the result is like this:
image

:+1:

Issue #2987, opened 2016-06-24
RFC: 7 Princes - Standardise all impact functions into 7 base IF's


timlinux:
**STATUS**: In Draft
# Background

Whenever we introduce support for a new hazard (lets say Tsunami), there is invariably code duplication and a maintenance overhead incurred as we clone existing IF's and modify them for a the new hazard. In addition, the new IF (let's say Raster Tsunami on Roads) is inherently characterised by the geometry type / data format of the hazard. So if we are actually generally interested in Tsunami impacts, we really want to cater for e.g.:
1. Polygon Tsunami impact on roads
2. Polygon Tsunami impact on polygon buildings
3. Polygon Tsunami impact on point buildings
4. Polygon Tsunami impact on polygon population
5. Polygon Tsunami impact on raster population
6. Polygon Tsunami impact on polygon landcover
7. Polygon Tsunami impact on raster landcover
8. Raster Tsunami impact on roads
9. Raster Tsunami impact on polygon buildings
10. Raster Tsunami impact on point buildings
11. Raster Tsunami impact on polygon population
12. Raster Tsunami impact on raster population
13. Raster Tsunami impact on polygon landcover
14. Raster Tsunami impact on raster landcover

And this is a more simple case where the hazard can only have e.g. continuous hazard data. Add support for classified data and we double the number of permutations. Add a new exposure type and we add more permutations. You can see why adding a new hazard becomes a high overhead activity with lots of code to maintain and a lot of work to do to keep everything consistent across all IF's and their reporting.
# Proposed solution

Our proposed solution looks something like this:

![inasafe4 0architecture](https://cloud.githubusercontent.com/assets/178003/16364847/a48b29f6-3bf0-11e6-993f-022b957f73f1.png)

In our solution there are some key points to note:
- When support for new hazards is added in InaSAFE, we do not create new impact functions we create **new metadata**.
- We do not generate reports, we generate datasets. The datasets are then used to create one or more reports.
- For a given family (e.g. impact on line features), all the reporting will be consistent as the IF methodology will be precisely the same regardless of Hazard type.
- We never have more than a handful of Impact Functions (5 illustrated here plus a few niche once like Earthquake damage curve based ones that do not fit this model).
- We assiduously avoid introducing special case logic into the impact functions.
- All hazards are first converted to classified data (barring the EQ exceptions), then polygonized. 
- We have two hazard polygonization strategies we can consider: 
[...]

PR #3161, opened 2016-08-16
WIP for #2987 - First pass implementation for standardised impact functions

ismailsunni:
## IF Implementation
Here is my plan (high level) of implementation.
if class revisited

1. Red box : Pre Processing
2. Black box : IF setup / setup_algorithm (based on hazard and exposure layer)
3. Orange box : Post processing
4. Green box : The algorithms (currently, they are separated class)
5. Blue box : The Impact Function class.

Need to be added:

1. Pre Processing
  1. Clipping
  2. Converting hazard layer to classified vector and add the value map

2. Post Processing
  1. All current post processing (gender, minimum needs, etc)
  2. Need to think where we put the action and notes since we are not using json anymore here. The only logical option is metadata.

Out side the blue box is the presentation layer. We need:

1. Template machine (jinja2)
2. Report format

### Dock interaction:
- Basically, we setup the IF based on layers chosen. Perhaps we can attach bind the IF object to the dock (?)
- IF name or the title (might by affected, flooded, etc) is generated on the fly

### Wizard Interaction
- Need to think about the flow

### Other system
- The other system should only call Impact Function class and give the required parameter

## Home work
- Hazard and exposure metadata which will be the driver of the IF

PS : 
Tim's drawing:

After heavy discussing with @timlinux @Gustry and @lucernae , we get this more details diagram for the new impact function architecture:
inasafe new impact function

You can see the realtimeboard of the diagram in here : https://realtimeboard.com/app/board/o9J_k1JfWJI=/

cc @timlinux @Gustry @lucernae 
please comment :D


safe/new_impact_function/test/test_impact_function.py
```diff
@@ -0,0 +1,152 @@
...
+from safe.test.utilities import get_qgis_app, standard_data_path
+QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
+
+from safe.new_impact_function.impact_function import ImpactFunction
+
+from qgis.core import QgsVectorLayer
+
+
+def read_json_flow(json_path):
+    """Helper method to read json file that contains a scenario
```
Gustry: Missing dot :)

safe/new_impact_function/impact_function.py
```diff
@@ -0,0 +1,645 @@
...
+        # Aggregation Preparation
+        if not self.aggregation:
+            impact_function_state['aggregation']['provided'] = False
+            if not self.actual_extent:
+                self._actual_extent = self.exposure.extent()
+            # Create aggregation layer from bbox = QgsVectorLayer()
+            impact_function_state['aggregation']['process'].append(
+                'Convert bbox aggregation to polygon layer with keywords')
+            aggregation_layer = QgsVectorLayer(
+                "Polygon?crs=epsg:4326", "aggregation", "memory")
```
Gustry: Should we use 4326 ? As we are going to reproject to exposure CRS I think.

safe/new_impact_function/test/test_impact_function.py
```diff
@@ -0,0 +1,152 @@
...
+
+        impact_function = ImpactFunction()
+        impact_function.exposure = exposure_layer
+        impact_function.hazard = hazard_layer
+        self.assertEqual(
+            impact_function.algorithm, impact_function.line_algorithm)
+        self.assertEqual(impact_function.name, 'Flood Polygon on Road Line')
+        self.assertEqual(impact_function.title, 'be affected')
+
+    @unittest.expectedFailure
```
Gustry: We should add a reason why it's an expected failure

safe/new_impact_function/test/test_impact_function.py
```diff
@@ -0,0 +1,152 @@
...
+            self.assertDictEqual(result, expected)
+
+        path = standard_data_path('scenario')
+        json_files = [os.path.join(path, f) for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+        for json_file in json_files:
+            test_scenario(json_file)
+
+
+
+
```
Gustry: Is-it PEP8 ?

safe/test/data/scenario/scenario_1.json
```diff
@@ -0,0 +1,46 @@
+{
```
Gustry: Can we give a more explicit name to the scenario ? haz + exp

ismailsunni: Do you mean change the hazard file name or how?

Gustry: The file name. "scenario_1.json" is not very explicit, if we have many scenarii, then it's gona be difficult to know.

safe/new_impact_function/impact_function.py
ismailsunni: Updated in the newer commit

safe/new_impact_function/test/test_impact_function.py
ismailsunni: Fixed in the newer commit.

safe/test/data/scenario/scenario_1.json
ismailsunni: Changed to polygon_hazard_point_exposure.json

safe/new_impact_function/test/test_impact_function.py
ismailsunni: Added in the comment

safe/new_impact_function/impact_function.py
```diff
@@ -0,0 +1,829 @@
...
+        formula = formula.replace(key, str(value))
+    return eval(formula)
+
+
+class ImpactFunction(object):
+    """Impact Function."""
+
+    def __init__(self):
+        self._hazard = None
+        self._hazard_keyword = {}
```
Gustry: We are going to use monkey patching to add  'keywords' to the 'QgsMapLayer'.
In the exposure, hazard and aggregation setters, you should monkey patch keywords.

The workflow will be to read keywords at the beginning of the IF and write them into the layer.keywords as a dictionary. Then in the impact function, we will only work with this monkey patching to insert/update/delete keywords. Only at the end of the impact function, we should use keyword_io to write them on disk.

So we should remove self._hazard_keyword etc

See PR https://github.com/inasafe/inasafe/pull/3219

safe/new_impact_function/impact_function.py
```diff
@@ -0,0 +1,766 @@
...
+    @hazard.setter
+    def hazard(self, layer):
+        """Setter for hazard layer property.
+
+        :param layer: Hazard layer to be used for the analysis.
+        :type layer: QgsMapLayer
+        """
+        self._hazard = layer
+        try:
+            self._hazard.keyword = KeywordIO().read_keywords(layer)
```
Gustry: I started to use `keywords`, not `keyword`. It seems more logical as we have many keyword ;-)

safe/new_impact_function/impact_function.py
```diff
@@ -0,0 +1,766 @@
...
+    @exposure.setter
+    def exposure(self, layer):
+        """Setter for exposure layer property.
+
+        :param layer: exposure layer to be used for the analysis.
+        :type layer: QgsMapLayer
+        """
+        self._exposure = layer
+        try:
+            self._exposure.keyword = KeywordIO().read_keywords(layer)
```
Gustry: Same for `keywords`

safe/new_impact_function/impact_function.py
```diff
@@ -0,0 +1,766 @@
...
+    @aggregation.setter
+    def aggregation(self, layer):
+        """Setter for aggregation layer property.
+
+        :param layer: aggregation layer to be used for the analysis.
+        :type layer: QgsMapLayer
+        """
+        self._aggregation = layer
+        try:
+            self._aggregation.keyword = KeywordIO().read_keywords(layer)
```
Gustry: Same for `keywords`

safe/new_impact_function/impact_function.py
```diff
@@ -0,0 +1,766 @@
...
+                post_processor)
+            self.set_state_process(
+                'post_processor',
+                'Post processor for %s %s.' % (
+                    post_processor['name'], post_processor_output))
+
+    def run(self):
+        self.preprocess()
+        self.run_algorithm()
+        self.post_process()
```
Gustry: Somewhere, at the end of the process, we need to write the keywords back using keyword_io.

Gustry:
Good work, do you think you will merge soon ?
We should try to keep our PR atomic and small.

safe/new_impact_function/impact_function.py
ismailsunni: Done.

safe/new_impact_function/impact_function.py
ismailsunni: Done also.

safe/new_impact_function/impact_function.py
ismailsunni: I have added it. But comment it out since it will make the test failed because there is no impact layer created.

ismailsunni:
@Gustry I have addressed your comment. We can merge this if you think it's good to be merged. We can add more code later. I agree with smaller PR.

safe/new_impact_function/impact_function.py
Gustry: Thanks

safe/new_impact_function/impact_function.py
Gustry: Ok, cool, it's fine like this.

Gustry:
LGTM.
For me you can merge, it's already a big PR, and I may need to work on definitions too.

Issue #3204, opened 2016-09-08
We need to refactor definitions.py


timlinux:
Big files are bad...lets split this up into a package with number of modules for easier maintenance.

Starting work on this in #3178

Issue #3178, opened 2016-09-01
We need to move hazard class mappings to definitions.py


timlinux:
### Problem

In the new architecture, hazard classes should be defined in definitions and not in the impact function metadata. For example definitions should tell you that if you have volcano we need three hazard class mappings for KRB1, KRB2 etc.

When the hazard is continuous data and thresholding is done, the class names from definitions.py should be applied to each threshold.

For categorical data we should have the hazard type from the original dataset replaced with the value map entry - which we take from definitions.py in the output.

![screen shot 2016-09-01 at 10 44 21 am](https://cloud.githubusercontent.com/assets/178003/18161218/558acb02-7031-11e6-9fac-13704d33847f.jpg)
### Proposed Solution

Here is the architecture we plan to implement for this:

![hazardclassification](https://cloud.githubusercontent.com/assets/178003/18348250/95ced480-75ca-11e6-81aa-9608e0c38531.png)

PR #3205, opened 2016-09-08
Implementation for assigning class names

timlinux:
See #3178 and #3204 

Not ready to merge yet


timlinux:
I am going to close this PR and redo it after @ismailsunni has finished his major changes to definitions.py since we are going to have a lot of conflicts. One thing we do want to retain are the definitions changes in d820a90 which add class values to hazard level names.

PR #3215, opened 2016-09-16
Minor fix (proposal)

kant:


timlinux:
Hi @kant : any reason why you want to take the user to that page rather than http://download.qgis.org? I prefer to send them to http://download.qgis.org since the that is a locale neutral URL. I'm going to close this without applying - if you have  good motivation for applying it please re-open this ticket and we can look at it!

Thanks!

PR #3219, opened 2016-09-20
add monkey patching for keywords on QgsMapLayer

Gustry:


safe/test/utilities.py
```diff
@@ -212,6 +213,60 @@ def standard_data_path(*args):
     return path
 
 
+def vector_test_layer(*args):
```
timlinux: load_test_vector_layer

safe/test/utilities.py
```diff
@@ -212,6 +213,60 @@ def standard_data_path(*args):
...
+    keyword_io = KeywordIO()
+    try:
+        layer.keywords = keyword_io.read_keywords(layer)
+    except NoKeywordsFoundError:
+        layer.keywords = {}
+
+    return layer
+
+
+def raster_test_layer(*args):
```
timlinux: load_test_raster_layer

PR #3220, opened 2016-09-21
Fix #3216 inasafe fields

### LINKED ISSUE #3216: Create List of Field Name for Impact Layer
### Problem

In the new impact function, we want to have more standard field name in the impact function. We need to list all the field name that will be put in the impact layer.

This will affect the post processor also.

cc @timlinux @Charlotte-Morgan @Gustry @lucernae

ismailsunni:
Contents:
- Introduce InaSAFE fields for version 4.

Please check @timlinux


safe/definitionsv4/fields.py
```diff
@@ -0,0 +1,353 @@
...
+    ]
+}
+# Feature Rate
+feature_rate_field = {
+    'key': 'feature_rate_field',
+    'name': tr('Feature Rate'),
+    'field_name': 'exp_rate',
+    'type': float,
+    'description': tr(
+        'Attribute where the rate value of the feature is located.'),
```
timlinux: Add: A rate value is the cost per unit of measure (m2 / m) for the feature.

safe/definitionsv4/fields.py
```diff
@@ -0,0 +1,353 @@
...
+            'text': None,
+            'link': None
+        }
+    ]
+}
+# Gender Ratio
+gender_ratio_field = {
+    'key': 'gender_ratio_field',
+    'name': tr('Gender Ratio'),
+    'field_name': 'gender_r',
```
timlinux: female_r

PR #3227, opened 2016-09-25
Fix import path problem.

ismailsunni:


timlinux:
LGTM thanks @ismailsunni

PR #3231, opened 2016-09-26
Add missing python package for tests discovering

Gustry:
This PR adds 58 missing tests in Travis.

The unit test discovering works only with package.


ismailsunni:
LGTM

PR #3232, opened 2016-09-26
fix reproject raster layer using gdal

Gustry:


safe/gisv4/raster/reproject.py
```diff
@@ -36,18 +45,91 @@ def reproject(layer, output_crs, callback=None):
...
     """
+    # I have a problem reprojecting with PyQGIS. I'm using GDAL.
+    # ET 26/09/16
+
+    # Fixme : Manage callback in these functions.
+    return _reproject_gdal(layer, output_crs, callback)
+
+
+def _reproject_gdal(layer, output_crs, callback=None):
+    """Reproject using GDAL command line.
```
ismailsunni: Add param and type

safe/gisv4/raster/reproject.py
```diff
@@ -36,18 +45,91 @@ def reproject(layer, output_crs, callback=None):
...
+    try:
+        reprojected.keywords = layer.keywords
+    except AttributeError:
+        raise KeywordNotFoundError
+
+    return reprojected
+
+
+def _reproject_pyqgis(layer, output_crs, callback=None):
+    """Not working properly for now. Need more tests.
```
ismailsunni: Add param and type.

safe/gisv4/vector/reproject.py
```diff
@@ -33,8 +35,8 @@ def reproject(layer, output_crs, callback=None):
     :type output_crs: QgsCoordinateReferenceSystem
```
ismailsunni: Complete param and type.

Gustry:
@ismailsunni to process a raster, we have 3 ways : 
- use the QGIS API only
- use the GDAL API
- use the GDAL command line

I tried to use to use the QGIS api, because it's better if we can use QgsRasterLayer everywhere. If I use GDAL API or GDAL command line, I need to work with a file on the file-system.
But it's not working properly. That's why I added this kind of wrapper to switch easily when I will find why it's not working with PyQGIS. The raster processing in pyqgis is not very well documented instead of GDAL.
So in the meantime, I added the GDAL command line to do the job.

safe/gisv4/vector/reproject.py
Gustry: Just expand the code, you will see it's there.

ismailsunni: Ah, my bad

Gustry:
We are not going to reproject raster. I'm going to close this PR and make a new one with only the code we need.

PR #3233, opened 2016-09-26
Updated diagrams

timlinux:


timlinux:
@ismailsunni @Gustry @Charlotte-Morgan note updated diagrams here...

Gustry:
LGTM. I think we should push to upstream, because we also would like to edit this file sometimes.

ismailsunni:
Agree with @Gustry

PR #3237, opened 2016-09-27
use definitions.py to know the step name and the output layer name

Gustry:
Removing also raster reproject
Small fixes


ismailsunni:
LGTM

PR #3241, opened 2016-09-28
Architecture diagram

timlinux:


Gustry:
I think we have fixed this issue in #3257 about the corrupted graph after the commit. i'm going to close this PR.

PR #3246, opened 2016-09-29
Update impact function flow

ismailsunni:
- Update IF flow based on the latest diagram per 29 September 2016


ismailsunni:
@timlinux yes I am aware of it. But I am not sure how to mark them as mutually exclusive for now. Basically, I just put all possible fields for each layer. 

If the layer's has only ratio field without population, the post processor won't be run.  Since there is not enough inputs for post processing.

If the user mark their field to have count field and ratio field, I think it's fine also. But, the post post processor won't be run (since there is already count filed).

So, user can choose which field is what, but the post processor needs to have proper inputs.

What do you think @timlinux ?

timlinux:
Yeah maybe just putting the logic in the postprocessor for now is fine.in a future iteration we can  add some front end logic to validate. In short term we can just put some text in the wizard to explain to the user what is allowed etc.

ismailsunni:
Btw @timlinux 
The exposure layer won't have women or age ratio. I recall that you said in the meeting that age or women ratio will be only in aggregation layer while age or women count can exist in exposure layer.

timlinux:
@ismailsunni yes correct

timlinux:
In the case where we have both ratio and count  for an exposure area, we can just give count precedence and ignore ration until someone gives us a compelling use case to do it differently...

PR #3254, opened 2016-09-30
#3178 work on hazard class mappings to definitions

Gustry:
This PR supersede : https://github.com/inasafe/inasafe/pull/3205 from @timlinux

### DIFF
--- a/safe/definitionsv4/hazard_classifications.py
+++ b/safe/definitionsv4/hazard_classifications.py
@@ -328,6 +328,7 @@ flood_raster_hazard_classes = {
             'key': 'wet',
             'name': tr('wet'),
             'description': tr('Water above ground height.'),
+            'class_value': 2,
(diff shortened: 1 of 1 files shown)


ismailsunni:
LGTM...

One question, what is the purpose of adding class_value? To be put in the hazard_class field?

Gustry:
Just a link for me : https://github.com/inasafe/inasafe/issues/2920#issuecomment-229874044

@timlinux Any reason why we didn't want to use a string for the hazard_class_field ?

timlinux:
Yeah I guess we can store just the class (string)

Gustry:
Ok, I will do.
@timlinux Can we also remove vector/hazard difference for hazard_classification ?

Gustry:
See the other PR : https://github.com/inasafe/inasafe/pull/3276

PR #3255, opened 2016-10-03
WIP - Keyword Wizard V4

ismailsunni:
Not ready yet.

Currently working on refactoring keyword wizard to use definitions.py


safe/definitionsv4/layer_modes.py
```diff
@@ -81,3 +81,8 @@
         }
     ]
 }
+
+layer_mode_all = [
+    layer_mode_classified,
+    layer_mode_continuous
+]
```
Gustry: make pep8 ?

safe/definitionsv4/test/test_utilities.py
```diff
@@ -0,0 +1,185 @@
...
+__revision__ = '$Format:%H$'
+
+
+class TestDefinitionsUtilities(unittest.TestCase):
+    """Test Utilities Class for Definitions."""
+
+    def test_definition(self):
+        """Test we can get definitions for keywords.
+
+        .. versionadded:: 3.2
```
Gustry: Copy/paste from 3.2 ?

safe/definitionsv4/utilities.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    result = []
+    for exposure in exposure_all:
+        if layer_geometry_key in exposure.get('layer_geometry'):
+            result.append(exposure)
+
+    return sorted(result, key=lambda k: k['key'])
+
+
+def hazard_categories_for_layer(layer_geometry_key, hazard_key=None):
```
Gustry: Useless parameter 'hazard_key' ?

safe/definitionsv4/utilities.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+
+    return sorted(hazard_category_all, key=lambda k: k['key'])
+
+
+def get_layer_modes():
+    """Return all sorted layer modes."""
+    return sorted(layer_mode_all, key=lambda k: k['key'])
+
+def hazard_units(hazard):
```
Gustry: pep8 ?

safe/gui/tools/wizard/step_kw45_extrakeywords.py
```diff
@@ -104,116 +103,19 @@ def get_next_step(self):
...
-        :type indx: int or str
-        """
-        if isinstance(indx, int) and indx > -1:
-            self.extra_keyword_changed(self.extra_keywords_widgets[7])
+        if (self.parent.get_layer_geometry_key() ==
+                layer_geometry_raster['key']):
+            return []
+        # Get hazard or exposure value
+        subcategory = self.parent.step_kw_subcategory.selected_subcategory()
+        return get_fields(
```
Gustry: Can you make a temporary var ?
It's difficult to debug when there is a `return` and a function at the same time.

safe/gui/tools/wizard/test/test_wizard_util.py
```diff
@@ -0,0 +1,14 @@
...
+
+"""Test for Wizard Util.
+"""
+
+import unittest
+
+
+class TestWizardUtil(unittest.TestCase):
+    pass
+
```
Gustry: Empty file ?

Gustry:
* Maybe later, we can improve the UI when we select the hazard ID
* Is-it possible to show the user that hazard ID is optional ?
* keyword_version is still 3.5 ?
* in the wizard, we can set `continuous` to a building exposure layer. It's not correct.

safe/gui/tools/wizard/step_kw45_extrakeywords.py
ismailsunni: Fixed in incoming commit.

safe/gui/tools/wizard/test/test_wizard_util.py
ismailsunni: So, there is already this `wizard_util.py`, but no test file for it. I added some methods in `wizard_util.py`, create `test_wizard_util.py` for my methods.

ismailsunni:
> Maybe later, we can improve the UI when we select the hazard ID
Yes, we can and should
> Is-it possible to show the user that hazard ID is optional ?
Simplest solution would be add description that it is optional.
> keyword_version is still 3.5 ? (at the end of the wizard)
We can change it later. For now, I want to minimalize failed test in this PR by not change it.
> at the end of the wizard, we can't see inasafe fields, such as exposure ID, hazard ID
Fixed in incoming commit.
in the wizard, we can set continuous to a building exposure layer. It's not correct.
> I haven't checked the exposure yet, so it's still pretty much not fixed. Will do it after this.
Inasafe fields should be InaSAFE fields
> Ah yes. The later was generated automatically. I will fix this issue.

safe/definitionsv4/utilities.py
ismailsunni: Fixed.

PR #3264, opened 2016-10-05
update graph with cleaning fields before the IF

Gustry:
I couldn't push to upstream because of the protected branch


ismailsunni:
I still can open it :D. LGTM.

It's not easy to see which part that has been updated.

PR #3268, opened 2016-10-05
Add Select Parameter for Parameter Package

ismailsunni:


safe_extras/parameters/numeric_parameter.py
```diff
@@ -9,9 +9,6 @@
 class NumericParameter(GenericParameter):
     """A subclass of generic parameter that accepts float or integer only.
 
-    Known issue : You should set the minimum and the maximum value explicitly.
-    See https://github.com/inasafe/inasafe/issues/2468#issuecomment-171124507
-
```
Gustry: Is-it no more a known issue ? Did you fix it ?

Gustry:
Looks good

safe_extras/parameters/numeric_parameter.py
ismailsunni: Hmm, I did nothing regarding this issue. I think I will need to check the status of parameter in inasafe and the parameter repo itself.
After looking deeper, there are some difference between the two which shouldn't happen.

Gustry: yes, I made some changes last year in inasafe/inasafe and i wasn't aware about inasafe/parameter repo.

Gustry: you don't want a GIT submodule ?

ismailsunni: There was a discussion about that, at the end we decided to not use git submodule or subtree. But, I can't remember what's the reason. Perhaps @timlinux remember about it?

Anyway, for short solution I will merge the update from InaSAFE repo to the parameter repository.

Gustry:
Looks good again

PR #3269, opened 2016-10-05
Fix for flood realtime reporting

lucernae:
Just smallfix for:
- Change hazard class style to mimic PetaJakarta.org scheme
- Separate folder for i18n


Gustry:
LGTM

PR #3271, opened 2016-10-06
Better at test skipping on docker QGIS.

ismailsunni:


safe/gui/tools/wizard/test/test_keyword_wizard.py
```diff
@@ -1031,14 +1024,14 @@ def test_integrated_line(self):
         # check if in step subcategory
         self.check_current_step(dialog.step_kw_subcategory)
 
-        expected_subcategories = ['Road']
+        expected_subcategories = ['Roads']
```
Gustry: Where does it come from ? Is it when we change in the definitions.py ? Can't you make the test dynamic ? To fetch the expected value from the definitons ?
Because we are not testing the definitions here, only the wizard, so our tests shoudn't be failing because of a change in definitions.

Gustry:
We should try to squash/rebase our commits. It's confusing to merge all these commits which are cancelled.

ismailsunni:
@Gustry yeah agree, I think I will squash and merge for this PR. The commits that I did were testing whether my changes are the problem for travis or not.

ismailsunni:
@Gustry btw the skipped test in the wizard actually fine in docker or local. Now, I am wondering wether we skip the test that only failed on docker or not.

safe/gui/tools/wizard/test/test_keyword_wizard.py
Gustry: It will be less maintenance if we change the value again in the definitions in the future. The main purpose of these tests is to check the wizard.

Gustry:
Ok thanks. Remove the `DOCKER` env in travis and you will use the squash button if you want in the github.

It's good if you could remove failing tests. I updated these tests a few days ago (last week) and I didn't look deeply, mea culpa.We shouldn't skip them !

ismailsunni:
Hi @Gustry 
After fixing some of the expected value to fetch from defintions, I realize I will need to update again the test later after I finished the keyword wizard, then I will work twice on it. So, I will leave it for now, and fix it afte I have finished the keyword wizard.

Gustry:
Ok, no problem

PR #3272, opened 2016-10-07
move back to an old docker image 2.14

Gustry:


Gustry:
See PR #3273, I forgot our own test_runner in this one

PR #3274, opened 2016-10-07
Add clip function

Gustry:


ismailsunni:
LGTM.
I am just wondering, should we put colors.py in definitionsv4/__init__.py also? I did that for other files, so that the `defintions` method will work properly.

PR #3287, opened 2016-10-11
Add hazard_classfications in hazard. PEP8.

ismailsunni:
Introduce hazard_classifications to hazard defintions

### DIFF
--- a/safe/definitionsv4/hazard.py
+++ b/safe/definitionsv4/hazard.py
@@ -1,6 +1,13 @@
 # coding=utf-8
 
 """Definitions relating to hazards."""
+from safe.definitionsv4.hazard_classifications import (
+    generic_hazard_classes,
+    volcano_hazard_classes,
+    flood_hazard_classes,
+    tsunami_hazard_classes,
+    ash_hazard_classes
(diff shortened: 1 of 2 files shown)


Gustry:
LGTM.
You are lucky, GIT is so clever that it can manage the pep8 conflict ;-) (we did the same fix)

ismailsunni:
@Gustry 
Yes, we can. For example a classification has 3 classes (low, medium, high) where the other has 4 classes (very low, low, medium, high) for the same hazard.

Gustry:
but we agree that we don't support this now ?
I mean I can see that all our lists have only one classification inside, like : 
```
'vector_hazard_classifications': [generic_vector_hazard_classes],
'raster_hazard_classifications': [generic_raster_hazard_classes],
'hazard_classifications': [generic_hazard_classes],
```

Gustry:
OK LGTM

ismailsunni:
@Gustry 
> but we agree that we don't support this now ?

Well, we just need to add more classification to the list. Since now we only have one per hazard, I only put one for each.

PR #3299, opened 2016-10-12
Use list of QVariant for all type in field definition.

ismailsunni:
Add better test for fields that only used in specific hazard or exposure


Gustry:
LGTM

ismailsunni:
Oops, many failed test. I will check again later. Low priority for now.

timlinux:
Thanks @ismailsunni LGTM

Issue #3289, opened 2016-10-11
We need an aggregate definitions document


timlinux:
### Problem

In our new v4.0 architecture we are collecting many terms and concepts in definitions. This is great! I think it would be very useful to collate all of these items into a single dynamically generated document which in essence forms a self documenting view of the InaSAFE conceptual space.
# CC

@Charlotte-Morgan

Gustry:
I think we can now close this one @timlinux , you did it !

PR #3300, opened 2016-10-12
[FEATURE] Added view for definitions help

timlinux:
See issue #3289 

screen shot 2016-10-12 at 9 41 24 am

Example rendering:

qgis

This is a first pass implementation. We can refine the presentation of this report as we go.


Gustry:
Great !

PR #3307, opened 2016-10-14
Migrate Metadata to v4

ismailsunni:


ismailsunni:
Closing this because we agree to not support keyword version 3.5 on InaSAFE 4.0.

We just need update all test to 4.0 like in #3438

PR #3308, opened 2016-10-14
Use list of QVariant for all type in field definition.

ismailsunni:
Not ready to merge


ismailsunni:
It's too complicated to fix it and the current one is working fine. I will close this based on monthly sprint meeting, 8 June 2017.

PR #3309, opened 2016-10-14
Wizard v4

ismailsunni:
Make keyword wizard read the current metadata of layer.

### DIFF
--- a/safe/test/data/hazard/volcano_krb.xml
+++ b/safe/test/data/hazard/volcano_krb.xml
@@ -185,19 +185,19 @@
             <gco:Tuple/>
           </prob_fatality_mag>
           <inasafe_fields>
-            <gco:Dictionary/>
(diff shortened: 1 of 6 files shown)


safe/test/data/hazard/volcano_krb.xml
```diff
@@ -232,6 +232,9 @@
           <layer_mode>
             <gco:CharacterString>classified</gco:CharacterString>
           </layer_mode>
+          <hazard_classification>
+            <gco:CharacterString>volcano_vector_hazard_classes</gco:CharacterString>
```
Gustry: Style using a vector hazard ?

Gustry:
LGTM after you read my comment above

safe/test/data/hazard/volcano_krb.xml
ismailsunni: Fixed in incoming commit. It was from bad copy and paste in volcano_hazard_classes definition.

PR #3333, opened 2016-10-21
Aggregate summary

Gustry:


safe/definitionsv4/fields.py
```diff
@@ -482,6 +525,7 @@
     adult_count_field,
     elderly_count_field,
     size_field,
+    total_field,
```
ismailsunni: Why does total go to exposure_fields? Should it go to aggregate_hazard layer?

safe/gisv4/vector/aggregate_summary.py
```diff
@@ -0,0 +1,132 @@
...
+        should accept params 'current' (int), 'maximum' (int) and 'step' (str).
+        Defaults to None.
+    :type callback: function
+
+    :return: The new aggregate hazard with summary.
+    :rtype: QgsVectorLayer
+
+    .. versionadded:: 4.0
+    """
+    impact_keywords = impact.keywords
```
ismailsunni: Would it be simpler if we just refer to `impact.keywords` directly? I saw there is only one reading for `impact_keywords` in line 48.

safe/definitionsv4/fields.py
Gustry: Done in the new commit

safe/gisv4/vector/aggregate_summary.py
Gustry: Yes of course ;-) I made some change many times and then I deleted some lines

Issue #3320, opened 2016-10-18
We need to make dock work with the new ImpactFunction class


timlinux:

Gustry:
What is this ticket @timlinux ?

timlinux:
I think we can just close it. Thanks!

PR #3368, opened 2016-10-29
Refactor dock to not use IF's

timlinux:
This is a work in progress. The goal is to not have dock do anything with listing impact functions etc. See issue #3320


timlinux:
@Charlotte-Morgan and others watching here note that since we do not have selectable impact functions any more, I am proposing that we instead formulate the question in the dock like this:

screen shot 2016-10-29 at 9 09 05 am

timlinux:
It might be nicer to change the word 'Aggregate' to 'Break down' as it is probably a bit less technical

Charlotte-Morgan:
Hi @tim
Thanks for this suggestion. I agree the dock does not need to use IF but would prefer to see the original question retained. 
eg. In the event of <hazard event>, how many <exposure > will be affected. 
This retains the InaSAFE "posing a question " approach, acknowledging the standards work on map titles and legends and preferred terminology of affected. 
If you want to drop the word aggregation this could become summarised by ...

timlinux:
Thanks for your comments @Charlotte-Morgan . How does this updated pic look?

screen shot 2016-10-29 at 12 19 07 pm

timlinux:
@Gustry or @ismailsunni can you review - I'll make follow up PR's as I go

safe/gui/widgets/dock.py
```diff
@@ -1015,11 +1018,38 @@ def draw_rubber_bands(self):
             self.extent.show_user_analysis_extent()  # blue
             try:
                 pass
-                # clip_parameters = self.impact_function.clip_parameters
-                # self.extent.show_last_analysis_extent(
+                #clip_parameters = self.impact_function.clip_parameters
+                #self.extent.show_last_analysis_extent(
```
Gustry: Did you run 'make pep8' ? The space is compulsory after a hash.

safe/gui/widgets/dock.py
```diff
@@ -1015,11 +1018,38 @@ def draw_rubber_bands(self):
...
                 #    clip_parameters['adjusted_geo_extent'])  # red
             except (AttributeError, TypeError):
-                pass
+                passinfo
```
Gustry: What is passinfo ?

safe/gui/widgets/dock.py
```diff
@@ -1034,17 +1064,15 @@ def accept(self):
             self.enable_busy_cursor()
             self.show_next_analysis_extent()
             self.impact_function = self.prepare_impact_function()
-            # clip_parameters = self.impact_function.clip_parameters
-            # self.extent.show_last_analysis_extent(
+            #clip_parameters = self.impact_function.clip_parameters
+            #self.extent.show_last_analysis_extent(
```
Gustry: Same here about the space

timlinux:
qgis

Here is an updated preview showing status updates as the analysis runs

Gustry:
Very nice ! Thanks Tim !

timlinux:
Thanks for the review @Gustry - PEP8 fixed with 78f9ad4

Gustry:
Thanks.
Can you fix travis and merge ? Already 34 files ;-)

PR #3371, opened 2016-10-30
Improve profiling

Gustry:
We can improve latter the display for the tree
screen shot 2016-10-30 at 22 51 56

### DIFF
--- a/safe/utilities/profiling.py
+++ b/safe/utilities/profiling.py
@@ -7,6 +7,7 @@ This code was taken from http://stackoverflow.com/a/3620972
 """
 
 import time
+import inspect
 from functools import wraps
 
 __copyright__ = "Vadim Shender (original poster in stack overflow), InaSAFE"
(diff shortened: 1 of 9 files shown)


safe/utilities/profiling.py
```diff
@@ -7,45 +7,103 @@
...
 def profile(fn):
     @wraps(fn)
     def with_profiling(*args, **kwargs):
-        global TOTAL_TIME
+        global ROOT
 
-        start_time = time.time()
-        PROF_DATA.append([fn.__name__, None])
-        index = len(PROF_DATA) - 1
+        curframe = inspect.currentframe()
```
timlinux: we should use nice naming conventions here too

timlinux: curframe -> current_frame

(I know that code is probably mine originally)

timlinux:
LGTM once small comments are taken into account

PR #3396, opened 2016-11-01
Fix 3380 kw wizard pop vector

### LINKED ISSUE #3380: Verify Wizard Workflow for Population Vector Layer
There have been many changes lately in wizard. Please check the workflow for population vector layer keyword assignment.

To do:
- [x]  Better unit test for population keyword creation
- [x]  Better unit test for existing population keyword

ismailsunni:
Fix #3380


safe/gui/tools/wizard/step_kw55_source.py
```diff
@@ -143,9 +140,12 @@ def set_widgets(self):
         source_date = self.parent.get_existing_keyword('date')
         if source_date:
             self.ckbSource_date.setChecked(True)
-            self.dtSource_date.setDateTime(
-                QDateTime.fromString(get_unicode(source_date),
-                                     'yyyy-MM-dd HH:mm:ss'))
+            try:
+                self.dtSource_date.setDateTime(source_date)
+            except TypeError:
+                raise TypeError(
```
Gustry: It's a bad practice to re-raise a **new** exception because we loose the traceback of the previous exception.
But it seems that `dtSource_date` is a QT object so we won't have a deeper traceback. So I'm fine for that one ;-)

ismailsunni: Ah, thanks. Fix it. The original TypeError give the same exception message, so I will delete it.

PR #3397, opened 2016-11-01
Fix 3383 kw wizard classified raster

### LINKED ISSUE #3383: Wizard for Classified Raster Layer
Currently, the wizard works classified vector layer. We will need to make it work also for classified raster layer.

To do:
- [x] Update how to get the class value from cell, not only from field
- [x] Update the class mapping step work with raster layer
- [x] Add unit test for classified raster data

ismailsunni:
Fix #3383

### DIFF
--- a/safe/gui/tools/wizard/test/test_keyword_wizard.py
+++ b/safe/gui/tools/wizard/test/test_keyword_wizard.py
@@ -13,7 +13,8 @@ from safe.test.utilities import (
     clone_shp_layer,
     get_qgis_app,
     standard_data_path,
-    load_test_vector_layer)
+    load_test_vector_layer,
+    load_test_raster_layer)
 # AG: get_qgis_app() should be called before importing modules from
 # safe.gui.tools.wizard
 QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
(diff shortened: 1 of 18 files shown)


safe/test/utilities.py
```diff
@@ -426,6 +426,7 @@ def load_test_raster_layer(*args, **kwargs):
     .. versionadded:: 4.0
     """
     path = standard_data_path(*args)
+    assert os.path.exists(path), 'Path %s is not exist' % path
```
Gustry: I'm not a pro in english, but `Path do not exist` or `Path is not existing` ;-)

Gustry: I didn't know we could assert outside of a test

ismailsunni: Ah yes, you are right about the English. Sometimes my English is off when there is '%s'. :D Thanks.

Well, we can. But, the main point of this to see in travis which path is not exist.

safe/gui/tools/wizard/test/test_keyword_wizard.py
```diff
@@ -1185,6 +1189,278 @@ def test_existing_exposure_population_polygon_keyword(self):
 
         self.assertDictEqual(real_keywords, expected_keyword)
 
+    # @unittest.skip('Path is not found')
```
Gustry: Do you need it ?

PR #3415, opened 2016-11-08
fix affected field and small refactoring in fields definition

Gustry:


lucernae:
LGTM.

Just a quick question, so: unaffected_count = total_field - total_affected_field?

safe/gisv4/vector/summary_3_analysis.py
```diff
@@ -132,16 +131,16 @@ def analysis_summary(aggregate_hazard, analysis, callback=None):
...
-    analysis.addAttribute(field)
-    analysis.keywords['inasafe_fields'][affected_count_field['key']] = (
-        affected_count_field['field_name'])
-
     field = create_field_from_definition(total_affected_field)
     analysis.addAttribute(field)
     analysis.keywords['inasafe_fields'][total_affected_field['key']] = (
         total_affected_field['field_name'])
 
+    field = create_field_from_definition(total_field)
```
lucernae: Ok, I think it is clear for me now, that Total field is the total count (from exposure), and total_affected_field is the count of exposure considered as affected.

Gustry: Yes !

Gustry:
> Just a quick question, so: unaffected_count = total_field - total_affected_field?

Yes, that's it, I didn't provide unaffected_count because it's very easy to compute it.

PR #3422, opened 2016-11-09
Fix allow resample step in kw wizard.

### LINKED ISSUE #3386: Verify Wizard Workflow for Population Raster Layer
There have been many changes lately in wizard. Please check the workflow for population raster layer keyword assignment.

To do:
- [ ] Check allow resampling step
- [ ] Better unit test for population keyword creation
- [ ] Better unit test for existing population keyword

ismailsunni:
Fix #3386

### DIFF
--- a/safe/gui/tools/wizard/step_kw20_unit.py
+++ b/safe/gui/tools/wizard/step_kw20_unit.py
@@ -24,6 +24,7 @@ from PyQt4.QtGui import QListWidgetItem
 
 from safe.definitionsv4.layer_purposes import (
     layer_purpose_exposure, layer_purpose_hazard)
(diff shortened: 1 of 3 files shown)


safe/gui/tools/wizard/step_kw20_unit.py
```diff
@@ -59,7 +60,10 @@ def get_next_step(self):
             new_step = self.parent.step_kw_classification
         else:  # No classifications
             if is_raster_layer(self.parent.layer):
-                new_step = self.parent.step_kw_source
+                if subcategory == exposure_population:
```
ismailsunni: No. Based on our discussion, we will put this kind of issue in this ticket: https://github.com/inasafe/inasafe/issues/3423

PR #3426, opened 2016-11-10
add function to fetch a layer according keywords

Gustry:


safe/datastore/datastore.py
```diff
@@ -187,6 +187,26 @@ def layer_uri(self, layer_name):
         """
         raise NotImplementedError
 
+    def layer_keyword(self, keyword, value):
+        """Get a layer according to a keyword and its value.
```
lucernae: This is fine for now, I only used hazard, exposure, impact and analysis layer for now.

PR #3427, opened 2016-11-10
add hazard class legend

Gustry:
It's not the best implementation. It's just to have a basic styling for the alpha release next week.


timlinux:
LGTM

PR #3430, opened 2016-11-12
check if layer is 4.0 in the IF setup

Gustry:
@ismailsunni I would like to merge this PR when the wizard will generate some 4.0 keywords.


ismailsunni:
Agree. I will try to use 4.0 this week.

PR #3453, opened 2016-11-17
Add PDF generations to dock

lucernae:


safe/gis/qgis_interface.py
```diff
@@ -104,7 +104,13 @@ def mock_getAlgorithm(name):
             :rtype: QgsAlgorithm  ?
             """
             Processing.initialize()
-            for provider in Processing.algs.values():
+            # FIXME: Had some weird bug in QGIS 2.18 MacOSX (KyngChaos)
```
lucernae: @Gustry @timlinux could you review this?
I encountered this error when I tried executing impact_function run with:
QGIS Version: 2.18 MacOSX (KyngChaos)
hazard: tsunami_wgs84.tif
exposure: pop_binary_raster_20_20.asc
in the test.

It is apparently tries to clip using processing algorithm, however it found that algs is a function, causing algs.values() to raise error.
When I check ```processing.__date__``` it was August 2012. Was it using outdated processing plugin?

Is it okay to do this?

timlinux:
Hi @lucernae. Check in your 
.qgis2 plugins folder that you have no processing folder. If you do remove it so that the built in one is used instead

PR #3468, opened 2016-11-22
Earthquake

Gustry:


safe/definitionsv4/hazard_classifications.py
```diff
@@ -88,6 +85,73 @@
     ]
 }
 
+earthquake_mmi_hazard_classes = {
+    'key': 'earthquake_mmi_hazard_classes',
+    'name': tr('Earthquake MMI classes'),
+    'description': tr(
+        'Three classes are supported for volcano vector hazard data: '
```
timlinux: Volcano -> earthquake.

Wouldn't it be better to have one class for each MMI level?

Gustry: It's just legacy. I took these 3 hazard levels in the previous EQ IF : https://github.com/inasafe/inasafe/blob/develop/safe/impact_functions/earthquake/earthquake_building/impact_function.py#L56

PR #3500, opened 2016-12-16
Make ITB Bayesian IF as default. Fix #3497.

### LINKED ISSUE #3497: Default IF: EQ raster on population raster
# problem
There are 3 EQ raster on raster population impact functions. The default is currently set as the ITB Fatality model. This is not the recommended model for the estimation of impact on people (ie fatalities and displacement).

# proposed solution
Following discussion with EQ experts at GA, we propose that the default IF be set as the ITB Bayesian Model

# cc
@RikkiWeber @dynaryu @griffij @ismailsunni @timlinux @Gustry @lucernae

ismailsunni:
Fix #3497 

Please check @Charlotte-Morgan


Gustry:
@timlinux has already release 3.5.4 without this fix

timlinux:
Let me but bump the version please. Also we should check with @Charlotte-Morgan what we should be using for real-time.

timlinux:
Sorry I meant LGTM not 'let me'.

PR #3501, opened 2016-12-16
Fix issue when multipart to singlepart results in none geometries.

timlinux:
No need to back port on this I think...

### DIFF
--- a/safe/storage/vector.py
+++ b/safe/storage/vector.py
@@ -484,8 +484,11 @@ class Vector(Layer):
             # Record coordinates ordered as Longitude, Latitude
             G = feature.GetGeometryRef()
             if G is None:
-                msg = ('Geometry was None in filename %s ' % filename)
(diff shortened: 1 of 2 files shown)


safe/storage/vector.py
```diff
@@ -484,8 +484,11 @@ def read_from_file(self, filename):
             # Record coordinates ordered as Longitude, Latitude
             G = feature.GetGeometryRef()
             if G is None:
-                msg = ('Geometry was None in filename %s ' % filename)
-                raise ReadLayerError(msg)
+                # Updated in 3.5.4 to skip invalid features TS
+                LOGGER.warning('Skipping feature with invalid geometry')
+                continue
+                #msg = ('Geometry was None in filename %s ' % filename)
+                #raise ReadLayerError(msg)
```
Gustry: Just for information, this is not PEP8 compliant. The space is compulsory after #.

PR #3502, opened 2016-12-16
extent and analysis analysis

Gustry:


safe/utilities/extent.py
```diff
@@ -31,53 +10,70 @@
...
 
 class Extent(object):
-    """Extent class to handle extent."""
+    """Extent class to handle analysis extent.
+
+    Rubber bands and extents for showing analysis extent etc.
+    Note that rubber bands are transient but their associated extents are
+    persistent for the session.
+
+    Every rubberbands are stored in the map canvas CRS.
```
timlinux: S/every//

timlinux:
Is this PR still valid?

Gustry:
yeah, I think so

PR #3503, opened 2016-12-16
Fix 3487 save as (need backport to develop)

Gustry:


timlinux:
Thanks for this @Gustry

PR #3507, opened 2016-12-20
Fix bug in debug mode and wizard

ismailsunni:


safe/impact_function_v4/impact_function.py
```diff
@@ -593,22 +593,19 @@ def _check_layer(layer, purpose):
             )
             return 1, message
 
+        # We should read it using KeywordIO for the very beginning. To avoid
+        # get the modified keywords in the patching.
```
Gustry: Why not updating the patching ?

Gustry:
Looks good to me, thanks !

PR #3508, opened 2016-12-20
Batch Runner v4

ismailsunni:
cc @ivanbusthomi


safe/gui/tools/batch/batch_dialog.py
```diff
@@ -520,59 +526,31 @@ def run_task(self, task_item, status_item, count=0, index=''):
...
-                        leg_impact_layer.providerType())
-                else:
-                    raise Exception('layer source is failed to be recognized')
-                # remove original layer created from dock, sometimes this
-                # code is failed leaving original file in layer panel
-                reg.instance().removeMapLayer(leg_impact_layer)
-                # turn off layer visibility just in case layer removing
-                # is not working so that it won't be shown in the next scenario
+                try:
+                    impact_layer = self.dock.impact_function._exposure_impacted
```
timlinux: I was rather hoping we could avoid using the dock and using the impact_function class directly....

safe/gui/tools/batch/scenario_runner.py
```diff
@@ -57,7 +57,8 @@ def completed(flag):
         LOGGER.debug("scenario done")
         dock.analysis_done.disconnect(completed)
 
-    dock.analysis_done.connect(completed)
+    # commenting this since the signal is removed from dock
```
timlinux: Maybe just clean it away completely. I dont think we should be worrying about the dock when using the batch runner...

safe/gui/tools/save_scenario.py
```diff
@@ -80,8 +80,8 @@ def validate_input(self):
         """
         self.exposure_layer = self.dock.get_exposure_layer()
         self.hazard_layer = self.dock.get_hazard_layer()
-        self.function_id = self.dock.get_function_id(
-            self.dock.cboFunction.currentIndex())
+        # self.function_id = self.dock.get_function_id(
```
timlinux: just delete this stuff....

safe/gui/tools/save_scenario.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, iface, dock):
         self.exposure_layer = None
         self.hazard_layer = None
         self.aggregation_layer = None
-        self.function_id = None
+        # self.function_id = None
```
timlinux: just delete ...

safe/gui/tools/save_scenario.py
```diff
@@ -98,10 +98,10 @@ def validate_input(self):
                 'hazard layer to do so.')
             is_valid = False
 
-        if self.function_id == '' or self.function_id is None:
-            warning_message = self.tr(
-                'The impact function is empty, can not save scenario')
-            is_valid = False
+        # if self.function_id == '' or self.function_id is None:
```
timlinux: just delete ...

safe/gui/tools/save_scenario.py
```diff
@@ -165,7 +165,7 @@ def save_scenario(self, scenario_file_path=None):
         parser.add_section(title)
         parser.set(title, 'exposure', relative_exposure_path)
         parser.set(title, 'hazard', relative_hazard_path)
-        parser.set(title, 'function', self.function_id)
+        # parser.set(title, 'function', self.function_id)
```
timlinux: just delete ...

Gustry:
Same for me, I would prefer the batch runner to be a new client for the IF and not using the dock.

ismailsunni:
@ivanbusthomi will address the comment and continue this PR.

+1 to make batch runner as a client for the IF.

ivanbusthomi:
@gustry and @timlinux 
OK, thanks to mas @ismailsunni I can continue working on it now. I will update the code with your suggestion thanks

ivanbusthomi:
@timlinux 
>If it is not too much work, can I suggest that we do not use the dock to get the impact function? Rather let the runner manage its own IF.

I've tried using a direct call to IF without using the dock, but I keep failing in making it works. Any tips maybe?

ivanbusthomi:
Hi @timlinux , @Gustry and @ismailsunni 
Apart from changing batchrunner to not use the dock to get the impact function, I think I've finished updating the code. Here are some samples to try, along with the scenario files. just extract and browse the directory when searching the scenario
[Ambon_clip.zip](https://github.com/inasafe/inasafe/files/663554/Ambon_clip.zip)

safe/gui/tools/batch/batch_dialog.py
```diff
@@ -520,59 +521,32 @@ def run_task(self, task_item, status_item, count=0, index=''):
...
-                else:
-                    raise Exception('layer source is failed to be recognized')
-                # remove original layer created from dock, sometimes this
-                # code is failed leaving original file in layer panel
-                reg.instance().removeMapLayer(leg_impact_layer)
-                # turn off layer visibility just in case layer removing
-                # is not working so that it won't be shown in the next scenario
+                impact_function = self.dock.impact_function
+                try:
+                    impact_layer = impact_function._exposure_impacted
```
Gustry: * You shouldn't use private members from the IF. You should always use public members.
* And you should use `impact_function.impact`, it will give the more detailed impacted layer (the exposure impacted or the aggregate_hazard_impacted, or at least the aggregation_impacted). So you can remove the try/except.

ivanbusthomi: ah, I was wondering about that too. I didn't know about it before. Thanks for the update @Gustry, TIL.
I will update the code

Gustry:
I tried the batch runner, but without success, it failed in the IF for now.

I'm adding the scenario.txt I'm using. (for mac/linux users)

[ambon_scenario.txt](https://github.com/inasafe/inasafe/files/663878/ambon_scenario.txt)

ivanbusthomi:
I've tried using the same data in Windows and Ubuntu, and it works after updating the scenario to correct directory.

PR #3509, opened 2016-12-20
compute the minimum extent in the IF

Gustry:


safe/impact_function_v4/test/test_impact_function.py
```diff
@@ -162,6 +163,73 @@ def test_impact_function_behaviour(self):
         self.assertEqual(impact_function.name, 'Flood Polygon On Road Line')
         self.assertEqual(impact_function.title, 'be affected')
 
+    def test_minimum_extent(self):
+        # Without aggregation layer
```
ismailsunni: Short descriptio for docstring here.

PR #3510, opened 2016-12-21
update graph about minimum extent for an analysis

Gustry:


timlinux:
LGTM merging for you...

PR #3515, opened 2016-12-22
Add minimum needs post processors.

lucernae:


timlinux:
Thats looking really good @lucernae ! Thanks! The english needs a bit of revision but I think we can merge (when your tests pass) and then revise definitions as a separate exercise. I think it would be good for @ismailsunni or @lucernae to review this too. Thanks!

safe/definitionsv4/minimum_needs.py
```diff
@@ -0,0 +1,140 @@
...
+def initializes_minimum_needs_fields():
+    """Initialize minimum needs fields.
+
+    Minimum needs definitions are taken from currently used profile.
+    """
+    needs_profile = NeedsProfile()
+    needs_profile.load()
+    fields = []
+
+    for n in needs_profile.get_needs_parameters():
```
ismailsunni: It's better to avoid one letter variable name. Perhaps it can be like this if you want to avoid breaking 80 chars in one line:
```
need_parameters = need.profile.get_needs_parameter
for need_parameter in need_parameters:
   pass
```

safe/definitionsv4/minimum_needs.py
```diff
@@ -0,0 +1,140 @@
...
+        need_parameter = n
+        """:type: ResourceParameter"""
```
ismailsunni: Better to use # for one line comment. Or is there special need for this comment?

safe/definitionsv4/minimum_needs.py
```diff
@@ -0,0 +1,140 @@
...
+def minimum_needs_field(field_key):
+    """Get minimum needs field from a given key.
+
+    :param field_key: Field key
+    :type field_key: str
+
+    :return: Field definition
+    :rtype: dict
+    """
+    for field in minimum_needs_fields:
```
ismailsunni: Where is this minimum_needs_fields defined? Global variable?

safe/metadata/exposure_layer_metadata.py
```diff
@@ -127,6 +127,13 @@ class ExposureLayerMetadata(GenericLayerMetadata):
             'inasafe/'
             'value_map/'
             'gco:Dictionary'),
+        'minimum_needs': (
```
ismailsunni: Where and when do we assign minimum_needs keyword in the exposure_layer_metadata? And why we put it here?

safe/definitionsv4/minimum_needs.py
lucernae: See comment below.

safe/definitionsv4/minimum_needs.py
lucernae: I'm using pycharm typehinting. AFAIK it doesn't work with # at the moment.
Also, that's why the variable is only n, because what I used is need_parameter.
Sadly, can't typehint for variable in pycharm directly for now.

safe/definitionsv4/minimum_needs.py
lucernae: Global variable in this file.

safe/metadata/exposure_layer_metadata.py
lucernae: Ah, I forgot to delete this. It's one of the things I want to ask in standup.
Should we let minimum needs expressed in exposure keywords too, or just taken from the profile as the only way to get it?

safe/reportv4/extractors/minimum_needs.py
```diff
@@ -28,49 +31,56 @@ def minimum_needs_extractor(impact_report, component_metadata):
...
     # minimum needs calculation only affect population type exposure
-    if not exposure_layer.keywords['exposure'] == exposure_population['key']:
+    # check if analysis keyword have minimum_needs keywords
+    have_minimum_needs_field = False
+    for field_key in analysis_keywords:
+        if field_key.startswith(minimum_needs_namespace):
+            have_minimum_needs_field = True
+            break
+
+    if not have_minimum_needs_field:
```
Gustry: Instead of having a boolean `found`, you can use a `for ... else` statement.

safe/impact_function_v4/postprocessors.py
```diff
@@ -218,13 +248,48 @@ def enough_input(layer, post_processor_input):
...
+                try:
+                    value = reduce(
+                        lambda d, k: d[k], input_value['value'], layer.keywords)
+                    found = True
+                    break
+                except KeyError:
+                    msg = 'Value %s is missing in keyword: %s' % (
+                        input_key, input_value['value'])
+
+        if not found:
```
Gustry: Instead of having a boolean `found`, you can use a `for ... else` statement.

safe/reportv4/extractors/minimum_needs.py
Gustry: http://psung.blogspot.fr/2007/12/for-else-in-python.html

safe/metadata/exposure_layer_metadata.py
ismailsunni: From the stand up, we agree to put it in the IF's product metadata provenance.

safe/impact_function_v4/test/test_impact_function.py
```diff
@@ -370,6 +372,113 @@ def test_post_processor(self):
                 self.assertIn(field_name, impact_fields)
         print_attribute_table(impact_layer, 1)
 
+    @expectedFailure
```
ismailsunni: Is there any reason why it's expected to be failed?

lucernae: Because it is failed at the moment :p. It doesn't return the correct results. Postprocessors calculation returned the wrong results for analysis layer. I hope @Gustry can help me on this after it is merged.

ismailsunni: I see. Perhaps you can add a notes that explain the reason in the code.

lucernae: I'm planning on writing it on issue rather than in the code.

ismailsunni: ok

PR #3517, opened 2016-12-23
Improve diagram and release script

timlinux:


scripts/release.sh
```diff
@@ -26,6 +26,7 @@ TARGZFILE="/tmp/${DIR}.tar.gz"
 mkdir -p ${WORKDIR}
 # Archive source code of the current branch to tar gz file.
 # Use git-archive-all since we use git submodule.
+brew install git-archive-all
```
ismailsunni: This is OSX specifix, should we put it here?
And `brew` is not available by default also.

timlinux: You will see there are already other brew commands in that script. I think its is better to be consistent and have all the reps installed rather than just some. Also on platforms without brew, those commands will just be silently skipped with a 'command not found' but should not prevent the script executing successfully (assuming the dependencies are there).

ismailsunni: Ah, I see. Ok, I agree.

PR #3518, opened 2016-12-26
Setting Page for InaSAFE default fields

ismailsunni:
I still get a funny bug, if I put value more than 100, it will revert to 100. Even if my value in QSetting is more than 100.
screen shot 2016-12-26 at 10 03 50
screen shot 2016-12-26 at 10 04 15


timlinux:
Isnt the funny bug just a validation constraint in the parameter widget?

ismailsunni:
@timlinux I thought that's the problem at first. But it's not. I set the max to a big number more than 100. Even if it's the case, it should make the user not able to input a value more than the max. But in this case, I can input a value, but not able to show it back after stored in the QSetting.

ismailsunni:
@timlinux ouch, it's indeed min and max, but from the default value not from the parameted widget since the min max are set after set the value.

safe/gui/tools/test/test_options_dialog.py
```diff
@@ -0,0 +1,30 @@
...
+LOGGER = logging.getLogger('InaSAFE')
+
+
+class TestOptionsDialog(unittest.TestCase):
+    """Test Options Dialog"""
+
+    def test_setup_dialog(self):
+        """Test Setup Options Dialog."""
+        dialog = OptionsDialog(PARENT, IFACE)
+        pass
```
Gustry: No `assert`, is-it normal ? Can we remove the test ? If this is a test, please explain it with a comment in the code.

safe/definitionsv4/constants.py
```diff
@@ -40,3 +40,6 @@
 ANALYSIS_SUCCESS = 0
 ANALYSIS_FAILED_BAD_INPUT = 3
 ANALYSIS_FAILED_BAD_CODE = 4
+
+GLOBAL = 'global'
```
Gustry: Can you explain with a comment what is RECENT and GLOBAL ?

ismailsunni:
@Gustry 
> Is-it possible to remove safe/defaults.py ? get_defaults ?

Well, it's still being used by EQ IF, so until we remove it, we can't remove the safe/defaults.py completely.

Gustry:
Looks good to me, thanks !

PR #3521, opened 2016-12-27
Fix 2nd step of the IFCW

ismailsunni:
Previously it shows all geometries, now it's only show allowed geometries based on the exposure and hazard.
 
ifcw_2nd_step

### DIFF
--- a/safe/gui/tools/wizard/step_fc05_functions2.py
+++ b/safe/gui/tools/wizard/step_fc05_functions2.py
@@ -1,35 +1,17 @@
 # coding=utf-8
-"""
-InaSAFE Disaster risk assessment tool by AusAid -**InaSAFE Wizard**
-
-This module provides: Function Centric Wizard Step: IF Constraint Selector 2
-
-Contact : ole.moller.nielsen@gmail.com
-
-.. note:: This program is free software; you can redistribute it and/or modify
-     it under the terms of the GNU General Public License as published by
-     the Free Software Foundation; either version 2 of the License, or
-     (at your option) any later version.
-
-"""
+"""InaSAFE Wizard Step for Choosing Layer Geometry"""
 
 import logging
 # noinspection PyPackageRequirements
 from PyQt4 import QtCore, QtGui
 from PyQt4.QtCore import pyqtSignature
 
-from safe.definitionsv4.layer_geometry import (
-    layer_geometry_point,
-    layer_geometry_line,
-    layer_geometry_polygon,
-    layer_geometry_raster
-)
 from safe.definitionsv4.utilities import get_allowed_geometries
 from safe.definitionsv4.layer_purposes import (
-    layer_purpose_exposure, layer_purpose_hazard
-)
-from safe.definitionsv4.hazard import hazard_all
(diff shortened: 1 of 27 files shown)


Gustry:
Is it OK if we hide the line if the line is grey ? We show the line if there is at least one green cell

timlinux:
Yeah that probably works @Gustry

ismailsunni:
@gustry which grey line?

Gustry:
screen shot 2016-12-27 at 09 42 47
Raster and line exposures are only grey.
I wonder if we should the same for hazard.

Gustry:
BTW, on the screenshot above, is-it normal to not have volcano raster ? It should be covered in our model.

ismailsunni:
> BTW, on the screenshot above, is-it normal to not have volcano raster ? It should be covered in our model.

Hmm, previously we didn't have raster volcano. But, I can add it since we can do analysis for volcano raster

safe/gui/tools/wizard/step_fc05_functions2.py
```diff
@@ -194,9 +185,17 @@ def set_widgets(self):
                 hc = hazard_layer_geometries[col]
                 ec = exposure_layer_geometries[row]
```
Gustry: Why did you use abbreviation such as hc, ec ?

safe/gui/tools/wizard/step_fc05_functions2.py
```diff
@@ -194,9 +185,17 @@ def set_widgets(self):
...
                 item = QtGui.QTableWidgetItem()
-                bgcolor = QtGui.QColor(120, 255, 120)
-                active_items += [item]
-                item.setBackground(QtGui.QBrush(bgcolor))
+
+                if (hc['key'] in h['allowed_geometries'] and ec['key'] in e[
+                        'allowed_geometries']):
+                    background_color = QtGui.QColor(120, 255, 120)
```
Gustry: Let's use the `definitions/colors` to store this

safe/gui/tools/wizard/step_fc05_functions2.py
```diff
@@ -194,9 +185,17 @@ def set_widgets(self):
...
+                    active_items += [item]
+                else:
+                    background_color = QtGui.QColor(220, 220, 220)
```
Gustry: Let's use the `definitions/colors` to store this

safe/gui/tools/wizard/step_fc05_functions2.py
ismailsunni: Legacy code. There are many like this in wizard. I will try to fix one by one.

safe/gui/tools/wizard/step_fc05_functions2.py
ismailsunni: Thanks, will do it.

ismailsunni:
GIS-order now :)
screen shot 2016-12-28 at 10 41 45

ismailsunni:
I will merge this since it's already addressed the main issue before it's getting bigger due to code refactoring and fixing coding standard.

PR #3524, opened 2016-12-28
Fix some issue reported by landscape and delete not used test.

ismailsunni:


Gustry:
Looks good to me, thanks !

safe/gui/tools/help/options_help.py
```diff
@@ -208,7 +208,7 @@ def content():
     bullets.add(m.Text(
         m.ImportantText(tr(
             'Contact email')),
-        tr(' - Use this option to specify the contact person\s email '
+        tr(' - Use this option to specify the contact person\'s email '
```
lucernae: In case like this, I think I would prefer for you to use double quote so you don't have to escape the quote inside.

safe/reportv4/impact_report.py
```diff
@@ -252,7 +252,7 @@ def __init__(
             analysis=None,
             exposure_breakdown=None,
             aggregation_impacted=None,
-            extra_layers=[],
+            extra_layers=None,
```
lucernae: Is this because quantified code warning? Did I use mutable on this?

safe/reportv4/impact_report.py
```diff
@@ -281,6 +281,8 @@ def __init__(
         self._aggregation_impacted = (
             aggregation_impacted or
             self._impact_function.aggregation_impacted)
+        if extra_layers is None:
```
lucernae: Why not:
```
if not extra_layers:
    extra_layers = []
```
Useful if you passed an empty dict, it still gets overwritten.

But still, I think we didn't have a chance to use this extra_layers for now.

safe/test/utilities.py
```diff
@@ -232,7 +232,7 @@ def standard_data_path(*args):
     :param args: List of path e.g. ['control', 'files',
         'test-error-message.txt'] or ['control', 'scenarios'] to get the path
         to scenarios dir.
-    :type args: list
+    :type args: str
```
lucernae: should be: list[str]
because args is a list of string from the looks of it in the code.

lucernae:
Other than that, LGTM. I will try to fix my quantified code too.

safe/reportv4/impact_report.py
ismailsunni: Ah yes, I think you are right.

safe/reportv4/impact_report.py
ismailsunni: Well, previously I didn't check it. 
But it seems we need to use immutable default argument after some google and read. One of the article: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

safe/reportv4/impact_report.py
ismailsunni: After some thought again, I think it's fine to do like this. The method expect a list if it gets an argument. If it doesn't get the argument `extra_layers`, it will use `None` (or any mutable value) and replace it with empty list in the code. I don't think we should handle the case where the argument is not a list, just like we did with the other arguments.

ismailsunni: The `None` here is only a way to make the default argument is a mutable object. We can change it with anything immutable.

safe/reportv4/impact_report.py
lucernae: Ah, ok. We shouldn't use mutable on args. Thanks

safe/gui/tools/help/options_help.py
ismailsunni: I think we should go with single quote, to keep our coding standard. https://github.com/inasafe/inasafe/wiki/Coding-Standards#strings-and-internationalisation

This string is simple enough (one escape only).

PR #3526, opened 2016-12-29
Fix various bugs in Option Dialog and Settings

ismailsunni:
- Add default InaSAFE settings
- Fix bug that some setting is never stored

### DIFF
--- a/safe/gui/tools/options_dialog.py
+++ b/safe/gui/tools/options_dialog.py
@@ -311,7 +311,9 @@ class OptionsDialog(QtGui.QDialog, FORM_CLASS):
     def accept(self):
         """Method invoked when OK button is clicked."""
         self.save_state()
-        self.dock.read_settings()
+        # FIXME: Option dialog should be independent from dock.
+        if self.dock:
+            self.dock.read_settings()
         self.close()
 
     # noinspection PyPep8Naming
--- a/safe/gui/tools/options_dialog.py
+++ b/safe/gui/tools/options_dialog.py
@@ -17,6 +17,7 @@ from safe_extras.parameters.integer_parameter import IntegerParameter
 
 from safe.definitionsv4.utilities import all_default_fields
 from safe.definitionsv4.constants import qvariant_whole_numbers, GLOBAL
+from safe.definitionsv4.default_settings import inasafe_default_settings
 from safe.common.utilities import temp_dir
 from safe.defaults import (
     disclaimer,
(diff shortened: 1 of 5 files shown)


safe/gui/tools/options_dialog.py
```diff
@@ -107,48 +106,89 @@ def __init__(self, iface, dock=None, parent=None, qsetting=''):
...
     def restore_state(self):
-        """Reinstate the options based on the user's stored session info.
-        """
-        # flag = settings.value(
-        #     'inasafe/useThreadingFlag', False)
-        # hack set use thread to false see #557
-        # TODO: Store default values somewhere else, e.g. defaults.py
+        """Reinstate the options based on the user's stored session info."""
         flag = False
         self.cbxUseThread.setChecked(flag)
```
lucernae: I don't understand this purpose?
Dos that mean we never support threading in the future? Why  not delete the options altogether?

ismailsunni: Historical:
There is a bug when the user enable the `use thread` option. So, we set it to alwasy False and hide the check box from the options dialog to prevent the user set it accidentally. 
Perhaps, there is still possibility to support it in the future.

safe/gui/tools/options_dialog.py
```diff
@@ -107,48 +106,89 @@ def __init__(self, iface, dock=None, parent=None, qsetting=''):
...
+        key = 'ISO19115_LICENSE'
+        value = self.settings.value(
+            'inasafe/%s' % key, inasafe_default_settings[key], type=str)
+        self.iso19115_license_le.setText(value)
+
+        # Keyword cache path
+        key = 'keywordCachePath'
+        path = self.settings.value(
+            'inasafe/%s' % key, inasafe_default_settings[key])
+        self.leKeywordCachePath.setText(path)
```
lucernae: All of the code above can be rewritten as for loops, something like this:

```
settings = [{
    'key': 'keywordCachePath',
    'control': self.leKeywordCachePath,
}]

for setting in settings:
    restoreControlSetting(setting['key'], setting['control'])
```

For me it improves readability, and reduce mistakes from copy pasting when you refactored it.
But it's up to you.

ismailsunni: @lucernae I am planning to do that also actually :), but may be after I worked on provenance.

safe/gui/tools/test/test_options_dialog.py
```diff
@@ -26,45 +26,106 @@
...
+            inasafe_default_settings['keywordCachePath'])
+        self.assertEqual(
+            dialog.template_warning_checkbox.isChecked(),
+            inasafe_default_settings['template_warning_verbose'])
+        self.assertEqual(
+            dialog.organisation_on_dock_checkbox.isChecked(),
+            inasafe_default_settings['showOrganisationLogoInDockFlag'])
+        self.assertEqual(
+            dialog.cbxDevMode.isChecked(),
+            inasafe_default_settings['developer_mode'])
```
lucernae: Same with this, I'd rather have for loops if its doing the same thing and have the assert displays the relevant key if it's not equal.

lucernae:
You can decide to go with the for loops or not :D.
Either of that, LGTM 👍

PR #3527, opened 2016-12-29
Use loop to set and restore option dialog.

ismailsunni:
I can't resist to make it sooner. Still need to fix here and there, but I think it's much better than 3-4 days ago. I will continue working on this option dialog after provenance.

cc @lucernae


lucernae:
LGTM 👍 as long as the test pass :D

PR #3528, opened 2016-12-30
Create Provenance in Metadata

ismailsunni:
- Remove unused code in IF


safe/impact_functions/bases/classified_rh_classified_ve.py
```diff
@@ -1,52 +0,0 @@
...
-    layer_geometry_point,
-    layer_geometry_line,
-    layer_geometry_polygon,
-    layer_geometry_raster)
-from safe.common.exceptions import MetadataLayerConstraintError
-from safe.impact_functions.base import ImpactFunction
-from safe.impact_functions.bases.layer_types.classified_raster_hazard import (
-    ClassifiedRasterHazardMixin)
-from safe.impact_functions.bases.layer_types.\
-    classified_vector_exposure import ClassifiedVectorExposureMixin
```
lucernae: If it is only for one import, what should we use? \ or ()?
Because, I also got confused sometimes. :p

ismailsunni: For me, I always avoid to use `\`. But, sometimes, the path to import is very long, and force me to use `\`.

It's unused legacy code anyway.

safe/impact_functions/bases/layer_types/continuous_vector_hazard.py
```diff
@@ -1,36 +0,0 @@
-# coding=utf-8
-from safe.common.exceptions import NoAttributeInLayerError
-from safe.impact_functions.bases.utilities import check_attribute_exist
-
-__author__ = 'Rizky Maulana Nugraha "lucernae" <lana.pcfre@gmail.com>'
-__date__ = '07/05/15'
```
lucernae: :'(
You deleted my codes from last year 😢 
hahaha.. that's ok.

safe/impact_functions/bases/test/test_base_classes.py
```diff
@@ -20,61 +11,21 @@
 from safe.common.exceptions import NoAttributeInLayerError
 from safe.impact_functions.bases.layer_types.classified_vector_exposure \
     import ClassifiedVectorExposureMixin
```
lucernae: Huh? Do we still use this class after refactoring?

ismailsunni: Yes, for earthquake IF. We will removed it for sure. Haha.

PR #3530, opened 2017-01-03
Provenance

ismailsunni:
@lucernae please list the things that you need for reporting or anything. I will add in the provenance. Or you can add it easily.

### DIFF
--- a/safe/impact_function_v4/impact_function.py
+++ b/safe/impact_function_v4/impact_function.py
@@ -750,9 +750,14 @@ class ImpactFunction(object):
             self._provenance['hazard_layer'] = self.hazard.source()
             self._provenance['hazard_keywords'] = deepcopy(
                 self.hazard.keywords)
-            self._provenance['aggregation_layer'] = (self.aggregation.source())
-            self._provenance['aggregation_keywords'] = deepcopy(
-                self.aggregation.keywords)
+            if self.aggregation:
+                aggregation_source = self.aggregation.source()
+                aggregation_keywords = self.aggregation.keywords
+            else:
+                aggregation_source = None
+                aggregation_keywords = None
+            self._provenance['aggregation_layer'] = aggregation_source
+            self._provenance['aggregation_keywords'] = aggregation_keywords
 
             return PREPARE_SUCCESS, None
 
@@ -1404,7 +1409,12 @@ class ImpactFunction(object):
         # InaSAFE
         self._provenance['impact_function_name'] = self.name
         self._provenance['impact_function_title'] = self.title
-        self._provenance['requested_extent'] = self.requested_extent
+        if self.requested_extent:
+            self._provenance['requested_extent'] = (
+                self.requested_extent.asWktCoordinates()
+            )
+        else:
+            self._provenance['requested_extent'] = None
         self._provenance['analysis_extent'] = (
             self.analysis_extent.exportToWkt()
         )
--- a/safe/impact_function_v4/impact_function.py
+++ b/safe/impact_function_v4/impact_function.py
@@ -1,12 +1,18 @@
 # coding=utf-8
 """Impact function
 """
+import getpass
+import platform
 from datetime import datetime
 from os.path import join, exists
 from os import makedirs
 from collections import OrderedDict
+from copy import deepcopy
+from socket import gethostname
 
-from PyQt4.QtCore import QSettings
+from PyQt4.QtCore import QT_VERSION_STR, QSettings
+from PyQt4.Qt import PYQT_VERSION_STR
+from osgeo import gdal
 from qgis.core import (
     QgsMapLayer,
     QgsGeometry,
@@ -21,6 +27,7 @@ from qgis.core import (
 import logging
 
 from safe.common.utilities import temp_dir
+from safe.common.version import get_version
 from safe.datastore.folder import Folder
 from safe.datastore.datastore import DataStore
 from safe.gisv4.vector.prepare_vector_layer import prepare_vector_layer
@@ -149,6 +156,7 @@ class ImpactFunction(object):
         self._performance_log = None
         self.reset_state()
         self._is_ready = False
+        self._provenance = {}
 
     @property
     def performance_log(self):
@@ -220,6 +228,7 @@ class ImpactFunction(object):
         """
         self._hazard = layer
         self._is_ready = False
+        self._provenance['hazard_layer'] = self._hazard.source()
 
     @property
     def exposure(self):
@@ -239,6 +248,7 @@ class ImpactFunction(object):
         """
         self._exposure = layer
         self._is_ready = False
+        self._provenance['exposure_layer'] = self._exposure.source()
 
     @property
     def aggregation(self):
@@ -258,6 +268,7 @@ class ImpactFunction(object):
         """
         self._aggregation = layer
         self._is_ready = False
+        self._provenance['exposure_layer'] = self._aggregation.source()
 
     @property
     def outputs(self):
@@ -648,6 +659,8 @@ class ImpactFunction(object):
                 return PREPARE_FAILED_BAD_INPUT, message
 
             status, message = self._check_layer(self.exposure, 'exposure')
+            self.provenance['exposure_keywords'] = deepcopy(
+                self.exposure.keywords)
             if status != PREPARE_SUCCESS:
                 return status, message
 
@@ -661,6 +674,8 @@ class ImpactFunction(object):
                 return PREPARE_FAILED_BAD_INPUT, message
 
             status, message = self._check_layer(self.hazard, 'hazard')
+            self.provenance['hazard_keywords'] = deepcopy(
+                self.hazard.keywords)
             if status != PREPARE_SUCCESS:
                 return status, message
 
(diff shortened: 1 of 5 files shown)


safe/impact_function_v4/impact_function.py
```diff
@@ -1337,3 +1356,121 @@ def style(self):
...
+        self._provenance['gdal_version'] = gdal.__version__
+        self._provenance['qt_version'] = QT_VERSION_STR
+        self._provenance['pyqt_version'] = PYQT_VERSION_STR
+        self._provenance['os'] = platform.version()
+        self._provenance['inasafe_version'] = get_version()
+
+        # InaSAFE
+        self._provenance['impact_function_name'] = self.name
+        self._provenance['impact_function_title'] = self.title
+        self._provenance['requested_extent'] = self.requested_extent
```
Gustry: exportToWkt too ?

safe/impact_function_v4/impact_function.py
```diff
@@ -1337,3 +1356,121 @@ def style(self):
...
+        """
+        # Environment
+        self._provenance['host_name'] = gethostname()
+        self._provenance['user'] = getpass.getuser()
+        self._provenance['qgis_version'] = QGis.QGIS_VERSION
+        self._provenance['gdal_version'] = gdal.__version__
+        self._provenance['qt_version'] = QT_VERSION_STR
+        self._provenance['pyqt_version'] = PYQT_VERSION_STR
+        self._provenance['os'] = platform.version()
+        self._provenance['inasafe_version'] = get_version()
```
Gustry: I would add the datastore's URI : https://github.com/inasafe/inasafe/blob/develop/safe/datastore/datastore.py#L71

Gustry: We need to check if there is a password in the URI (postgis connexion for instance)

safe/impact_function_v4/test/test_impact_function.py
```diff
@@ -381,6 +390,59 @@ def test_post_processor(self):
...
+
+        provenance = impact_function.provenance
+        # from pprint import pprint
+        # pprint(provenance)
+        self.maxDiff = None
+
+        expected_provenance.update({
+            'action_checklist': impact_function.action_checklist(),
+            'analysis_extent': impact_function.analysis_extent.exportToWkt(),
+            'gdal_version': gdal.__version__,
```
Gustry: Can you move host_name, inasafe_version, qt_version etc to dict creation ? And update the dict with only items related to the IF.

safe/impact_function_v4/test/test_impact_function.py
```diff
@@ -381,6 +390,59 @@ def test_post_processor(self):
...
+            'qgis_version': QGis.QGIS_VERSION,
+            'qt_version': QT_VERSION_STR,
+            'requested_extent': impact_function.requested_extent,
+            'user': getpass.getuser(),
+            'os': platform.version()
+        })
+
+        self.assertDictEqual(expected_provenance, provenance)
+
     @expectedFailure
```
Gustry: Any reason why this test is failing ?

safe/impact_function_v4/impact_function.py
```diff
@@ -1337,3 +1356,121 @@ def style(self):
...
+
+    @property
+    def provenance(self):
+        """Helper method to gather provenance for exposure_impacted layer.
+
+        :returns: Dictionary that contains all provenance.
+        :rtype: dict
+        """
+        # Environment
+        self._provenance['host_name'] = gethostname()
```
Gustry: I think we should make compulsory to run `prepare` before calling this property, no ?
Something like https://github.com/inasafe/inasafe/blob/develop/safe/impact_function_v4/impact_function.py#L816 ?

safe/impact_function_v4/impact_function.py
```diff
@@ -644,7 +652,10 @@ def prepare(self):
                 )
                 return PREPARE_FAILED_BAD_INPUT, message
 
+            self._provenance['exposure_layer'] = self.exposure.source()
```
Gustry: Hum, I'm confused. If you call self.exposure.source() in this function, you will get the source from the last exposure layer, not the original one. `performance_log` is called at the end. Same for hazard and aggregation

safe/impact_function_v4/impact_function.py
```diff
@@ -680,8 +694,13 @@ def prepare(self):
                     )
                     return PREPARE_FAILED_BAD_INPUT, message
 
+                self._provenance['aggregation_layer'] = (
```
Gustry: What happen if aggregation layer is not provided in the IF ?

safe/impact_function_v4/impact_function.py
Gustry: Be careful, requested_extent is empty if a aggregation layer is provided !
I think you should add boolean like `aggregation_provided`.

safe/impact_function_v4/test/test_impact_function.py
ismailsunni: It's from @lucernae PR. See https://github.com/inasafe/inasafe/pull/3515#pullrequestreview-14636589

Actually, he want to ask you about it :)

safe/impact_function_v4/impact_function.py
ismailsunni: Yes, I will just return empty dictionay if the IF is not ready. Thanks

safe/impact_function_v4/impact_function.py
ismailsunni: Really? I got the correct one actually.
I put here, in the prepare method to avoid getting the not-original source one when we call run (since there are some layer assignments inside the method).

safe/impact_function_v4/test/test_impact_function.py
ismailsunni: Done in the next commit.

safe/impact_function_v4/impact_function.py
ismailsunni: After checking that the provenance will be return on `self._is_ready`, I got this problem. The impact function will be set to not ready when it runs the `run` method since there are some exposure, hazard, or aggregation assignment in this method. Is it intended?

safe/impact_function_v4/impact_function.py
ismailsunni: Ah I understand. Because the layers are re-assigned, then it won't be able to run the IF again. So, it's not ready and need to call prepare method.
In that case, I will just add another status, which indicate that the IF has a proper output from the analysis. It will be set to True only after call run method finished.

safe/impact_function_v4/impact_function.py
ismailsunni: Done in the next commit.

safe/impact_function_v4/impact_function.py
ismailsunni: I will just use aggregation_layer = None if it doesn't have aggregation layer. And the requested_extent will set to None.

safe/impact_function_v4/impact_function.py
```diff
@@ -149,6 +156,18 @@ def __init__(self):
         self._performance_log = None
         self.reset_state()
         self._is_ready = False
+        self._provenance_ready = False
+        self._provenance = {
+        # Environment
```
Gustry: Indentation ?

safe/impact_function_v4/impact_function.py
```diff
@@ -721,6 +744,23 @@ def prepare(self):
...
+            self._provenance['exposure_keywords'] = deepcopy(
+                self.exposure.keywords)
+            self._provenance['hazard_layer'] = self.hazard.source()
+            self._provenance['hazard_keywords'] = deepcopy(
+                self.hazard.keywords)
+            if self.aggregation:
+                aggregation_source = self.aggregation.source()
+                aggregation_keywords = deepcopy(self.aggregation.keywords)
+                print aggregation_source
+                print aggregation_keywords.get('layer_purpose')
```
Gustry: Remove these 2 prints

safe/impact_function_v4/test/test_impact_function.py
Gustry: I think you forgot this comment. But no worries, it was a minor change. The same for the test just after without aggregation.

ismailsunni: hi @Gustry I already did it in https://github.com/inasafe/inasafe/pull/3530/files#diff-6d55c27a3cd8f7d2b3d4fac87ede6569R162.
I keep the test like that since I use it to compare. Or did I miss understand your comment?

safe/impact_function_v4/test/test_impact_function.py
lucernae: @Gustry yes, I think the number isn't right and @timlinux suggested to mark the test as expectedFailure first. I hoped you help me on this, actually. It is the thing I asked in the channel about population postprocessors calculations on analysis_impacted level.

safe/impact_function_v4/test/test_impact_function.py
```diff
@@ -383,8 +396,6 @@ def test_scenario_directory(self):
             scenario, expected_steps, expected_outputs = read_json_flow(
                 json_file)
             if scenario.get('enable', True):
-                print "Test JSON scenario : "
-                print json_file
```
Gustry: Sorry, but these two prints are useful.
* They are in a test, not in production environment.
* There is no other way to know  which scenario is running as the test will loop over every scenario. It's difficult to debug when we don't know which scenario is failing.

I wonder if the log produces the correct output.

ismailsunni:
Hmm, strange travis is failing.
test_provenance is failing if run in all test. But, if we skip test_profiling, it will pass.

ismailsunni:
Thanks @Gustry 
I will merge it.

PR #3532, opened 2017-01-04
Refactoring the extent selector

Gustry:


timlinux:
LGTM thanks @Gustry . 

Just one downside to mention with using enumerated constants is that if we are reading the settings file / registry, it will be more opaque as to what the actual extent mode is. However I think that is OK - you can just reading in definitions/constants to see what the values mean and the code is cleaner (e.g. less prone to bugs due to typos).

safe/definitionsv4/constants.py
```diff
@@ -33,9 +33,17 @@
     QVariant.ULongLong
 ]
 
+# Extent selector
+HAZARD_EXPOSURE = 'HazardExposure'
+HAZARD_EXPOSURE_VIEW = 'HazardExposureView'
+HAZARD_EXPOSURE_BOOKMARK = 'HazardExposureBookmark'
+HAZARD_EXPOSURE_BOUNDINGBOX = 'HazardExposureBoundingBox'
```
Gustry: @timlinux It's not an enumerated constant. In the settings, the string will be saved.

Gustry:
Ok, finally. I think I finished to refactor the dock. Sorry, I didn't provide unit test. I will try to do it in JK next week.

Can you review it again ? (Don't forget to expand the dock diff in github, it's hidden by default).

I think some bugs are there, for example some message are not displayed properly. But we can fix them step by step. Most of the refactoring is here.

@ismailsunni We need to fix the IFCW to use the extent too.

ismailsunni:
LGTM.

Yes, we need to do that. On my to do list after we have proper extent selector in dock.

safe/definitionsv4/constants.py
timlinux: Ah ok cool...sorry I  missed that...

PR #3551, opened 2017-01-09
remove the old impact_function package

Gustry:


Gustry:
@ismailsunni Can we merge this PR ?
I already fix what I could in Landscape and QuantifiedCode.

@lucernae I made some change in realtime to use inasafe v4 but it will not work.

ismailsunni:
@Gustry yes. I agree to merge this one. I think I will just disable quantifiedcode since I don't really give much attention to it.

Issue #3555, opened 2017-01-10
Skip Step in Wizards which Can be Skipped


ismailsunni:
Example:

![](https://files.gitter.im/inasafe/inasafe/QOj5/Screen-Shot-2017-01-10-at-09.47.20.png)

ismailsunni:
Closed in #3568

PR #3568, opened 2017-01-11
Skip inasafe-field step when there is no field found #3555

myarjunar:
issue #3555
- add inasafe_field_for_the_layer function
- add condition to check if there are any fields on layer-purpose that can be the inasafe-field. if nothing were found, skip the inasafe-field step and set the next step to source step.

### DIFF
--- a/safe/gui/tools/wizard/step_kw40_classify.py
+++ b/safe/gui/tools/wizard/step_kw40_classify.py
@@ -23,7 +23,8 @@ from osgeo.gdalconst import GA_ReadOnly
 
 from safe.definitionsv4.layer_purposes import layer_purpose_aggregation
 from safe.definitionsv4.layer_geometry import layer_geometry_raster
-from safe.definitionsv4.utilities import get_fields
+from safe.definitionsv4.utilities import get_fields, get_compulsory_fields
+from safe.definitionsv4.constants import no_field
 from safe.gui.tools.wizard.wizard_step import WizardStep
 from safe.gui.tools.wizard.wizard_step import get_wizard_step_ui_class
 from safe.gui.tools.wizard.wizard_strings import (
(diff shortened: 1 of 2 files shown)


Gustry:
@ismailsunni you are the wizard guru ;-)

safe/gui/tools/wizard/step_kw40_classify.py
```diff
@@ -77,6 +78,26 @@ def get_next_step(self):
         else:
             subcategory = {'key': None}
 
+        layer_data_provider = self.parent.layer.dataProvider()
+        # Iterate through all inasafe fields
+        for inasafe_field in self.inasafe_fields_for_the_layer():
```
ismailsunni: Hi @myarjunar 
I think you just need to check whether `self.inasafe_fields_for_the_layer()` is empty or not.

And in this code, it will go directly to `self.parent.step_kw_source` if it doesn't have `inasafe_fields`. What if it has inasafe_default_fields? It should go to inasafe_default_fields step first not go directly to `step_kw_source`.

safe/gui/tools/wizard/step_kw40_classify.py
```diff
@@ -92,6 +113,34 @@ def get_next_step(self):
         # Any other case
         return self.parent.step_kw_source
 
+    def inasafe_fields_for_the_layer(self):
```
ismailsunni: I suggest you to refactor this and put this method in wizard utilities, since there is similar method in the inasafe_field_step.

safe/gui/tools/wizard/test/test_keyword_wizard.py
```diff
@@ -320,6 +321,152 @@ def test_hazard_volcano_polygon_keyword(self):
         real_keywords = dialog.get_keywords()
         self.assertDictEqual(real_keywords, expected_keyword)
 
+    def test_hazard_earthquake_polygon_keyword(self):
+        """Test keyword wizard for earthquake hazard polygon."""
```
Gustry: I think you should rename the test to `test_layer_without_inasafe_fields` because this test is not specific to earthquake.

safe/gui/tools/wizard/step_kw40_classify.py
```diff
@@ -302,3 +307,24 @@ def populate_classified_values(
...
+        """
+        layer_data_provider = self.parent.layer.dataProvider()
+        # Iterate through all inasafe fields
+        for inasafe_field in inasafe_fields:
+            skip_flag = True  # set default value
+            for field in layer_data_provider.fields():
+                # Check the field type
+                if isinstance(inasafe_field['type'], list):
+                    if field.type() in inasafe_field['type']:
+                        skip_flag = False
```
ismailsunni: You can just return True if we found one inasafe_fields

ismailsunni:
LGTM @myarjunar 
Just one comment from me, and one from @Gustry and this PR is ready to be merged.
Way to go....

PR #3571, opened 2017-01-11
try to use the mocked QGIS interface

Gustry:
* We don't use the get_qgis_app, we use the iface provided by QGIS.
* We don't execute code before an import, it's gonna be good for landscape ;-)

### DIFF
--- a/safe/datastore/test/test_folder.py
+++ b/safe/datastore/test/test_folder.py
@@ -16,20 +16,13 @@ import unittest
 from tempfile import mkdtemp
 from os.path import join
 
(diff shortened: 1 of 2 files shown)


ismailsunni:
@Gustry the failed error that you got is the same as @lucernae got yesterday. Not sure why, restarting the travis will fix it.

ismailsunni:
Will it work on another OS with the same QGIS version?
cc @myarjunar who uses Ubuntu xenial.

Gustry:
This should work on any OS. It's only used for testing.

PR #3573, opened 2017-01-12
trying to check this random bug in the IF

Gustry:
We need to know why we have this bug.


Gustry:
Ok, I think I found it. It was because of the unique name which was unique !
The name is used to create the datastore.

PR #3576, opened 2017-01-12
Add multiple classifications in hazard layer metadata

ismailsunni:
Beware of many failing unit tests.


ismailsunni:
Included in #3703

PR #3577, opened 2017-01-12
Fix wizard not remove widget

ismailsunni:


safe/gui/tools/wizard/step_kw45_inasafe_fields.py
```diff
@@ -191,3 +191,11 @@ def get_inasafe_fields(self):
                 inasafe_fields[parameter.guid] = parameter.value
 
         return inasafe_fields
+
+    def clear(self):
```
Gustry: Not possible to refactor this in the parent class ?

ismailsunni: No, only 3 step from all steps that has similar behaviour.
We can, but the implementation is specific to each step. We can refactor it again later.

Gustry: Ok, no worries

Gustry:
lgtm

PR #3578, opened 2017-01-12
fix default values in the attribute table

Gustry:


ismailsunni:
can you check @samnawi

PR #3580, opened 2017-01-13
Set pop count to use float also.

ismailsunni:
Some data set use float / real number type for population count, so we need to set the population count to use float also.

cc @samnawi


samnawi:
LGTM

PR #3583, opened 2017-01-13
update crs saving to match changes in dock.extent

ivanbusthomi:
a small change to match changes in dock.extent in order to fix error in saving scenario for batchrunner.


Gustry:
Looks good to me, thanks, should we merge ?

ivanbusthomi:
Yep, thanks @Gustry

PR #3584, opened 2017-01-13
Skip step raster inasafe default values in wizard.

ismailsunni:


Gustry:
Comments look good to me !

PR #3585, opened 2017-01-13
Do not update any metadata after load the layers to QGIS.

ismailsunni:


Gustry:
Less code in the dock, Looks good to me, thanks !

PR #3591, opened 2017-01-14
More improvements to definitions and definitions help

timlinux:
Now we can set images for each definition and have them show up in the help. Also more improvements to the definitions text.

qgis


timlinux:
More screenies of things added:

screen shot 2017-01-15 at 3 53 50 am

screen shot 2017-01-15 at 3 54 48 am

safe/definitions/post_processors.py
```diff
@@ -106,44 +109,58 @@ def post_processor_affected_function(**kwargs):
...
+    field_input_type,
+    dynamic_field_input_type,
+    keyword_input_type,
+    needs_profile_input_type,
+    geometry_property_input_type,
+    layer_property_input_type
+]
+
+# Input values
+# TODO: @ismailsunni please give better explanation of these
```
Gustry: What do you need of us ?

timlinux:
@gustry wrote:

> You made some changes to safe_extras/parameters. Can you revert this change ?

Strange why it lists it since there are no actual changes:

screen shot 2017-01-15 at 1 43 35 pm


> Can you include the PDF generated in the docs folder ?

Hmm yeah maybe once I have the CSS rendering properly in the PDF and we have reached freeze I could do it. What is your thinking here? To be able to see changes to the metadata over time?

> Maybe we need to add a date and time when the PDF was generated ?

Yeah we could do - maybe an InaSAFE version too... something to add to the footer I guess...

I'd like to leave the PDF related things above for another PR...

PR #3593, opened 2017-01-15
Support for damage curve analysis

Gustry:
Support for damage curve analysis


safe/gisv4/raster/align.py
```diff
@@ -26,8 +26,7 @@
 
 
 @profile
-def align_rasters(
-        hazard_layer, exposure_layer, extent):
+def align_rasters(hazard_layer, exposure_layer, extent):
     """Align hazard and exposure raster layers so they fit perfectly and
```
timlinux: First line of doc string should be precis'd to 79 chars and terminated with a full stop and a new line.

Issue #3579, opened 2017-01-12
Checklist - suggestion for InaSAFE Wizard


ismailsunni:
From our meeting on Monday, 9 January 2016

- [x] Is the max included in the threshold range? Perhaps put < or <= sign,
- [x] Rename Setting as Global Default
- [ ] Fix combo box in field, to show all option at once. #4028 
- [x] Aggregation ID shown in exposure layer keywords. (Fixed in #3577)
- [ ] Add setting to choose currency
- [ ] Make the description lower case in Default Values Setting #4029
- [x] Rename Default values to Global Default
- [x] Group the default values, typo for Eldery in Default Values Setting
- [ ] Explain the value means in feature

From meeting with @Charlotte-Morgan 
- [ ] Rename female ratio —> gender attribute : female ratio
- [x] check to add up to 1 for age ratio
- [x] Set custom max = 1 for ratio
- [ ] Make the question uniform between wizard steps
- [x] Give more description from concepts.py

cc @timlinux it seems you were working on some parts of those items.

ismailsunni:
cc @samnawi and @felix-yew and @ivanbusthomi

Charlotte-Morgan:
Some thoughts on the process of assigning key words to continuous population vector count data with the following attributes: id, village, total_pop, female, youth, adult, elderly.

Stepping through the KW wizard: exposure: population: continuous: count: 
In the next step I think this should ask "Please select the attribute in this layer that represents the **total** number of people"
**Next step**: "Please select attributes in this layer for disaggregated age and gender counts"
- [ ] Age attribute: youth count
- [ ] Age attribute: adult count
- [ ] Age attribute: elderly count
- [ ] Gender attribute: female count

Notes for each to say something like: Attribute containing the number of females.
Drop down to be the definition: Females are wonderful ...

_I suggest that asking for the layer ID and layer name should come at the end of the wizard before the final screen with source data._

Please check drop downs have enough space to display options clearly.

**next screen** only show if one or more attribute in the previous screen was left blank.

Heading: "If your data do not have disaggregated age and gender details; please select age and gender ratios for your population or skip this step to apply the global defaults. "

Any field that has been populated (filled in) in the previous screen should not now ask people for a ratio.
The global defaults should be pre loaded as the default. Users should not be offered a field because the continuous population data can only have counts not ratios as  a unit. 
_the total for the age groups should equal 1 regardless of the number of classes used . ie if the user does not use elderly, youth and adult should equal 1._

[...]

PR #3595, opened 2017-01-16
Remove population ratio and count from exposure.

ismailsunni:
Removing population, female, age ratio and count field from exposure.

Question: should I remove also population count from exposure place?

cc @timlinux @Charlotte-Morgan

### DIFF
--- a/safe/report/test/test_impact_report.py
+++ b/safe/report/test/test_impact_report.py
@@ -260,21 +260,27 @@ class TestImpactReport(unittest.TestCase):
             'notes': u'Columns and rows containing only 0 or "No data" '
                      u'values are excluded from the tables.',
             'detail_table': {
-                'headers': [u'Structures type', u'High Hazard Zone',
-                            u'Medium Hazard Zone', u'Total Affected',
-                            u'Total Unaffected', u'Total'],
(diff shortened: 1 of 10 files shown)


Charlotte-Morgan:
hi @ismailsunni - thanks for making these changes. When I look at this PR I have a problem when i work with aggregation data - the screen freezes and I can not use the scroll bar or dropdowns.
For the population, can you please remove the option to add feature rate and feature value. From chat with @timlinux - i do not think this is relevant to population data #3579

safe/report/test/test_impact_report.py
```diff
@@ -258,21 +260,27 @@ def test_analysis_breakdown_detail(self):
...
+                    [u'health', 1, 0, 1, 0, 1]
+                ],
                 'footers': [u'Total', 4, 2, 6, 3, 9]
             }
         }
         actual_context = analysis_breakdown.context
-
         self.assertDictEqual(
-            expected_context, actual_context, different_context_message)
+            expected_context, actual_context)
```
lucernae: please add max_diff=NaN and delete all who uses different_context_message

ismailsunni: It should be maxDiff = None :D 
Added in new commit.

PR #3609, opened 2017-01-17
remove realtime directory

myarjunar:


Gustry:
Great, -26165 lines ! You will have a very bad ratio (removing VS adding code in inasafe) ;-)

Issue #3548, opened 2017-01-09
Editorial review needed for definitions


timlinux:
### Problem

We need to review definitions and make sure the language is all neat & tidy.

### Proposed Solution

Make it nice!

Charlotte-Morgan:
Hi @timlinux - its great to see all this in one place and at the same time a bit scary :)
For the user it may be a bit overwhelming to have all this info - I like the hyperlinked TOC - that helps.
There seems to be some overlap between this document and the help files available in the doc and tools; raising the question as to whether the content should only appear in one place or if additional help buttons should be added to some places (wizards) with relevant content available eg, hazard definitions, exposure definitions and defaults.
For people maintaining docs and polishing code - its a great resource, very quick and easy to see what is missing and to spot places for tweaks.
For people preparing templates and reports - great to see what content is available to incorporate.

Comments on code **arising:**
In the new v 4.0 analysis you add analysis layers and fields. Its not clear in the doc what the Type 2,3,4,5, and 6 mean.
![image](https://cloud.githubusercontent.com/assets/5743135/22012491/d53b5338-dcc6-11e6-9467-e0e4c3baa158.png)

In some data types it seems meaning less to have default min and max. here is one example - there are others 
![image](https://cloud.githubusercontent.com/assets/5743135/22012502/e479fd5e-dcc6-11e6-8c45-db679bf4f506.png)

9. InaSAFE defaults - does not show feature value
10.0.9 aggregation fields: descriptions of ratios refer to number of people. This should say "The proportion of young people for each feature." for youth etc
![image](https://cloud.githubusercontent.com/assets/5743135/22012870/c1d5db22-dcc8-11e6-8536-72701596426f.png)

13. There is some inconsistency with impact layer naming and descriptions 
I think there could be some benefit to show the table / attribute definitions with the layer descriptions. 
At the moment the fields are described before the laye ris defined. I would find it useful to have the concept, description and fields in that order. I still think the names are not intuitive - but do not have a good suggestion yet.

15. We need a concept description for a post processor so users know what ithis is.
Is it possible to add some styling and remove the extra "post processor" words in front of Input Types, Input Values, Process types?

Charlotte-Morgan:
point 9 above fixed in #3620

timlinux:
Thanks for the review @Charlotte-Morgan 

> For the user it may be a bit overwhelming to have all this info - I like the hyperlinked TOC - that helps.
[...]

PR #3614, opened 2017-01-17
Added missing batch runner help to definitions document

timlinux:
See #3548


timlinux:
@ismailsunni its due to the squash merge. Ill fix it by using @Gustry 's suggested workflow in the next PR....they won't show in the history because they will be squashed...

PR #3635, opened 2017-01-18
Optimize impact function for flooding on buildings

wonder-sk:
Low hanging fruits:
- cache exposure features instead of fetching them directly from layer by ID
- use prepared geometry for repeated intersection tests of hazard geometries
- in smart clip, only fetch layers within extent of mask layer

For a small analysis area, the time on my laptop went from 12.4s down to 6.1s


Gustry:
Thanks a lot Martin, it looks very nice !

PR #3642, opened 2017-01-19
Layer style according to standards

Gustry:


timlinux:
tandards = dentist in afrikaans

PR #3645, opened 2017-01-19
fix minimum needs calculator to use the new post-processor

myarjunar:
- use Qgis combo box object
- use the new post processor

### DIFF
--- a/safe/gis/vector/prepare_vector_layer.py
+++ b/safe/gis/vector/prepare_vector_layer.py
@@ -108,7 +108,7 @@ def prepare_vector_layer(layer, callback=None):
         raise NoFeaturesInExtentError
 
     _add_id_column(cleaned)
-    _rename_remove_inasafe_fields(cleaned)
+    rename_remove_inasafe_fields(cleaned)
 
     if _size_is_needed(cleaned):
         run_single_post_processor(cleaned, post_processor_size)
@@ -161,7 +161,7 @@ def _check_value_mapping(layer):
 
 
 @profile
-def _rename_remove_inasafe_fields(layer):
+def rename_remove_inasafe_fields(layer):
     """Loop over fields and rename fields which are used in InaSAFE.
 
     :param layer: The layer
@@ -186,7 +186,7 @@ def _rename_remove_inasafe_fields(layer):
     # for minimum needs calculator.
     # If there is no displaced_field keyword, then pass
     try:
-        if layer.keywords['inasafe_fields']['displaced_field']:
+        if layer.keywords['inasafe_fields'][displaced_field['key']]:
             fields.append(displaced_field)
     except KeyError:
         pass
--- a/safe/gis/vector/prepare_vector_layer.py
+++ b/safe/gis/vector/prepare_vector_layer.py
@@ -29,6 +29,7 @@ from safe.definitions.fields import (
     exposure_type_field,
     exposure_class_field,
     count_fields,
+    displaced_field
 )
 from safe.definitions.exposure import indivisible_exposure
 from safe.definitions.layer_purposes import (
@@ -166,7 +167,6 @@ def _rename_remove_inasafe_fields(layer):
     :param layer: The layer
     :type layer: QgsVectorLayer
     """
-
     # Exposure
     if layer.keywords['layer_purpose'] == layer_purpose_exposure['key']:
         fields = get_fields(
@@ -181,6 +181,15 @@ def _rename_remove_inasafe_fields(layer):
     elif layer.keywords['layer_purpose'] == layer_purpose_aggregation['key']:
         fields = get_fields(
             layer.keywords['layer_purpose'])
+
+    # Add displaced_field definition to expected_fields
+    # for minimum needs calculator.
+    # If there is no displaced_field keyword, then pass
+    try:
+        if layer.keywords['inasafe_fields']['displaced_field']:
+            fields.append(displaced_field)
+    except KeyError:
+        pass
 
     expected_fields = {field['key']: field['field_name'] for field in fields}
 
--- a/safe/gui/tools/minimum_needs/needs_calculator_dialog.py
+++ b/safe/gui/tools/minimum_needs/needs_calculator_dialog.py
@@ -30,10 +30,12 @@ from safe.messaging import styles
 from safe.gui.tools.help.needs_calculator_help import needs_calculator_help
 from safe.impact_function.postprocessors import run_single_post_processor
 from safe.definitions.post_processors import minimum_needs_post_processors
+from safe.definitions.fields import displaced_field, aggregation_name_field
+from safe.definitions.layer_purposes import layer_purpose_aggregation
 from safe.gis.vector.tools import (
     create_memory_layer, copy_layer)
 from safe.gis.vector.prepare_vector_layer import (
-    _rename_remove_inasafe_fields)
+    rename_remove_inasafe_fields)
 
 INFO_STYLE = styles.INFO_STYLE
 LOGGER = logging.getLogger('InaSAFE')
@@ -55,6 +57,8 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         self.setWindowTitle(self.tr(
             'InaSAFE %s Minimum Needs Calculator' % get_version()))
 
+        self.button_box.button(QtGui.QDialogButtonBox.Ok).setEnabled(False)
+
         # get qgis map layer combobox object
         self.layer = QgsMapLayerComboBox()
         self.layer.setFilters(QgsMapLayerProxyModel.VectorLayer)
@@ -71,10 +75,10 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         self.aggregation_id = QgsFieldComboBox()
 
         # add Qgis combo box to window
-        self.gridLayout_2.addWidget(self.layer, 0, 1)
-        self.gridLayout_2.addWidget(self.displaced, 1, 1)
-        self.gridLayout_2.addWidget(self.aggregation_name, 2, 1)
-        self.gridLayout_2.addWidget(self.aggregation_id, 3, 1)
+        self.gridLayout_2.addWidget(self.layer, 3, 1)
+        self.gridLayout_2.addWidget(self.displaced, 5, 1)
+        self.gridLayout_2.addWidget(self.aggregation_name, 7, 1)
+        self.gridLayout_2.addWidget(self.aggregation_id, 9, 1)
 
         # set field to the current selected layer
         self.displaced.setLayer(self.layer.currentLayer())
@@ -85,6 +89,10 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         self.layer.layerChanged.connect(self.displaced.setLayer)
         self.layer.layerChanged.connect(self.aggregation_name.setLayer)
         self.layer.layerChanged.connect(self.aggregation_id.setLayer)
+
+        # enable/disable ok button
+        self.update_button_status()
+        self.displaced.fieldChanged.connect(self.update_button_status)
 
         # Set up things for context help
         self.help_button = self.button_box.button(QtGui.QDialogButtonBox.Help)
@@ -100,6 +108,15 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         ok_button = self.button_box.button(QtGui.QDialogButtonBox.Ok)
         ok_button.clicked.connect(self.accept)
 
+    def update_button_status(self):
+        """
+        """
+        # enable/disable ok button
+        if len(self.displaced.currentField()) > 0:
+            self.button_box.button(QtGui.QDialogButtonBox.Ok).setEnabled(True)
+        else:
+            self.button_box.button(QtGui.QDialogButtonBox.Ok).setEnabled(False)
+
     def minimum_needs(self, input_layer):
         """Compute minimum needs given a layer and a column containing pop.
 
@@ -114,40 +131,63 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         for needs in minimum_needs_post_processors:
             run_single_post_processor(input_layer, needs)
 
+    def prepare_new_layer(self, input_layer):
+        """Prepare new layer for the output layer.
+
+        :param input_layer: Vector layer
+        :type input_layer: QgsVectorLayer
+
+        :return: New memory layer duplicated from input_layer
+        :rtype: QgsVectorLayer
+        """
+        # create memory layer
+        output_layer = (
+            '%s_minimum_needs' % input_layer.name())
+        output_layer = create_memory_layer(
+            output_layer,
+            input_layer.geometryType(),
+            input_layer.crs(),
+            input_layer.fields())
+
+        # monkey patching input layer to make it work with
+        # prepare vector layer function
+        temp_layer = input_layer
+        temp_layer.keywords = {'layer_purpose':
+                                   layer_purpose_aggregation['key']}
+
+        # add keywords to output layer
+        copy_layer(temp_layer, output_layer)
+
+        return output_layer
+
     def accept(self):
         """Process the layer and field and generate a new layer.
 
         .. note:: This is called on OK click.
 
         """
-        # create memory layer
-        output_layer = (
-            '%s_minimum_needs' % self.layer.currentLayer().name())
-        output_layer = create_memory_layer(
-            output_layer,
-            self.layer.currentLayer().geometryType(),
-            self.layer.currentLayer().crs(),
-            self.layer.currentLayer().fields())
+        # Create a new layer for output layer
+        output_layer = self.prepare_new_layer(self.layer.currentLayer())
 
-        # monkey patching input layer to make it work with
-        # prepare vector layer function
-        temp_layer = self.layer.currentLayer()
-        temp_layer.keywords = {'layer_purpose': 'aggregation'}
-
-        # add keywords to output layer
-        copy_layer(temp_layer, output_layer)
-        output_layer.keywords['layer_purpose'] = 'aggregation'
+        # Monkey patching output layer to make it work with
+        # minimum needs calculator
+        output_layer.keywords['layer_purpose'] = (
+            layer_purpose_aggregation['key'])
         output_layer.keywords['inasafe_fields'] = (
-            {'displaced_field': self.displaced.currentField(),
-             'aggregation_name_field': self.aggregation_name.currentField()})
+            {displaced_field['key']: self.displaced.currentField(),
+             aggregation_name_field['key']:
+                 self.aggregation_name.currentField()})
 
         # remove unnecessary fields & rename inasafe fields
-        _rename_remove_inasafe_fields(output_layer)
+        rename_remove_inasafe_fields(output_layer)
 
         try:
             self.minimum_needs(output_layer)
-        except:
+        except Exception as e:
+            LOGGER.debug(e)
             return
+
+        # write layer to geopackage file
 
         # noinspection PyArgumentList
         QgsMapLayerRegistry.instance().addMapLayers([output_layer])
--- a/safe/gui/tools/minimum_needs/needs_calculator_dialog.py
+++ b/safe/gui/tools/minimum_needs/needs_calculator_dialog.py
@@ -35,8 +35,9 @@ from safe.gis.vector.tools import (
 from safe.gui.tools.help.needs_calculator_help import needs_calculator_help
 from safe.impact_function.postprocessors import run_single_post_processor
 from safe.messaging import styles
-from safe.utilities.resources import html_footer, html_header, get_ui_class
 from safe.utilities.qgis_utilities import display_critical_message_box
+from safe.utilities.resources import html_footer, html_header, get_ui_class
+from safe.utilities.utilities import humanise_exception
 
 INFO_STYLE = styles.INFO_STYLE
 LOGGER = logging.getLogger('InaSAFE')
@@ -114,17 +115,28 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
             population counts.
         :type input_layer: QgsVectorLayer
 
-        :returns: Layer with attributes for minimum needs.
-        :rtype: QgsVectorLayer
+        :returns: A tuple containing True and the vector layer if
+                  post processor success. Or False and an error message
+                  if something went wrong.
+        :rtype: tuple(bool,QgsVectorLayer or basetring)
         """
         # Create a new layer for output layer
         output_layer = self.prepare_new_layer(input_layer)
 
         # count each minimum needs for every features
         for needs in minimum_needs_post_processors:
-            run_single_post_processor(output_layer, needs)
+            is_success, message = run_single_post_processor(
+                output_layer, needs)
+            # check if post processor not running successfully
+            if not is_success:
+                LOGGER.debug(message)
+                display_critical_message_box(
+                    title=self.tr('Error while running post processor'),
+                    message=message)
 
-        return output_layer
+                return False, None
+
+        return True, output_layer
 
     def prepare_new_layer(self, input_layer):
         """Prepare new layer for the output layer.
@@ -174,14 +186,20 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         .. note:: This is called on OK click.
 
         """
-
+        # run minimum needs calculator
         try:
-            self.result_layer = self.minimum_needs(self.layer.currentLayer())
+            success, self.result_layer = (
+                self.minimum_needs(self.layer.currentLayer()))
+            if not success:
+                return
         except Exception as e:
-            LOGGER.debug(e)
+            error_name, traceback = humanise_exception(e)
+            message = (
+                'Problem(s) occured. \n%s \nDiagnosis: \n%s'.format(
+                    error_name, traceback))
             display_critical_message_box(
-                title='Error while calculating minimum needs', message=e)
-            return
+                title=self.tr('Error while calculating minimum needs'),
+                message=message)
 
         # remove monkey patching keywords
         del self.result_layer.keywords
@@ -200,7 +218,10 @@ class NeedsCalculatorDialog(QtGui.QDialog, FORM_CLASS):
         else:
             data_store = Folder(temp_dir(sub_dir=self.result_layer.name()))
 
+        data_store.default_vector_format = 'geojson'
         data_store.add_layer(self.result_layer, self.result_layer.name())
+
+        self.result_layer = data_store.layer(self.result_layer.name())
 
         # noinspection PyArgumentList
         QgsMapLayerRegistry.instance().addMapLayers(
--- a/safe/gui/tools/minimum_needs/needs_calculator_dialog.py
+++ b/safe/gui/tools/minimum_needs/needs_calculator_dialog.py
@@ -36,6 +36,7 @@ from safe.gui.tools.help.needs_calculator_help import needs_calculator_help
 from safe.impact_function.postprocessors import run_single_post_processor
 from safe.messaging import styles
 from safe.utilities.resources import html_footer, html_header, get_ui_class
+from safe.utilities.qgis_utilities import display_critical_message_box
 
 INFO_STYLE = styles.INFO_STYLE
 LOGGER = logging.getLogger('InaSAFE')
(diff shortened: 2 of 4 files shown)


safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -14,17 +14,27 @@
 __copyright__ += 'Disaster Reduction'
 
 import logging
-from qgis.core import QgsMapLayerRegistry, QgsVectorLayer
-from PyQt4 import QtGui, QtCore
+from qgis.core import QgsMapLayerRegistry
+from qgis.gui import (
+    QgsFieldComboBox, QgsMapLayerComboBox, QgsMapLayerProxyModel,
```
Gustry: We use one line per import.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -14,17 +14,27 @@
...
-from safe.utilities.utilities import add_ordered_combo_item
 from safe.messaging import styles
 from safe.gui.tools.help.needs_calculator_help import needs_calculator_help
+from safe.impact_function.postprocessors import run_single_post_processor
+from safe.definitions.post_processors import minimum_needs_post_processors
+from safe.gis.vector.tools import (
+    create_memory_layer, copy_layer)
+from safe.test.utilities import monkey_patch_keywords
+from safe.utilities.keyword_io import KeywordIO
+from safe.gis.vector.prepare_vector_layer import \
```
Gustry: We use brackets to make it multi-line

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -45,7 +55,38 @@ def __init__(self, parent=None):
...
+        self.displaced_cb = QgsFieldComboBox()
+        self.displaced_cb.setFilters(QgsFieldProxyModel.Numeric)
+
+        # get field that represent aggregation name
+        self.aggregation_cb = QgsFieldComboBox()
+        self.aggregation_cb.setFilters(QgsFieldProxyModel.String)
+
+        # get field that represent aggregation id
+        self.aggregation_id_cb = QgsFieldComboBox()
+        self.aggregation_id_cb.setFilters(QgsFieldProxyModel.Numeric)
```
Gustry: I wonder if it's good to have a numeric only. We may have text ID

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
         ok_button = self.button_box.button(QtGui.QDialogButtonBox.Ok)
         ok_button.clicked.connect(self.accept)
 
-    def minimum_needs(self, input_layer, population_name):
+    def minimum_needs(self, input_layer):
         """Compute minimum needs given a layer and a column containing pop.
 
         :param input_layer: InaSAFE layer object assumed to contain
             population counts
         :type input_layer: read_layer
```
Gustry: QgsVectorLayer

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
         ok_button = self.button_box.button(QtGui.QDialogButtonBox.Ok)
         ok_button.clicked.connect(self.accept)
 
-    def minimum_needs(self, input_layer, population_name):
+    def minimum_needs(self, input_layer):
         """Compute minimum needs given a layer and a column containing pop.
 
         :param input_layer: InaSAFE layer object assumed to contain
```
Gustry: `InaSAFE layer object`, This is now a vector layer

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
-        # Now disable the run button if no suitable fields were found
-        # see #2206
-        ok_button = self.button_box.button(QtGui.QDialogButtonBox.Ok)
-        if not has_fields:
-            ok_button.setEnabled(False)
-        else:
-            ok_button.setEnabled(True)
+        # count each minimum needs for every features
+        for needs in minimum_needs_post_processors:
+            print needs
```
Gustry: remove print

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
-
-        index = self.cboPolygonLayers.currentIndex()
-        layer_id = self.cboPolygonLayers.itemData(
-            index, QtCore.Qt.UserRole)
-        # noinspection PyArgumentList
-        layer = QgsMapLayerRegistry.instance().mapLayer(layer_id)
-
-        file_name = layer.source()
+        # create memory layer
+        output_layer = "%s_minimum_needs" \
```
Gustry: `'%s_minimum_needs'`

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
+        output_layer = create_memory_layer(
+            output_layer,
+            self.map_layer_cb.currentLayer().geometryType(),
+            self.map_layer_cb.currentLayer().crs(),
+            self.map_layer_cb.currentLayer().fields())
+
+        # read keywords from input
+        temp_layer = self.map_layer_cb.currentLayer()
+        try:
+            keyword_io = KeywordIO()
```
Gustry: Remove the use of KeywordIO and use always the monkey patching

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
+        output_layer = "%s_minimum_needs" \
+                       % self.map_layer_cb.currentLayer().name()
+        output_layer = create_memory_layer(
+            output_layer,
+            self.map_layer_cb.currentLayer().geometryType(),
+            self.map_layer_cb.currentLayer().crs(),
+            self.map_layer_cb.currentLayer().fields())
+
+        # read keywords from input
+        temp_layer = self.map_layer_cb.currentLayer()
```
Gustry: Call it 'layer'. We don't use abbreviation in the code.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
+        try:
+            keyword_io = KeywordIO()
+            temp_layer.keywords = keyword_io.read_keywords(temp_layer)
+        except KeywordNotFoundError:
+            temp_layer.keywords['layer_purpose'] = 'aggregation'
+
+        # add keywords to output layer
+        copy_layer(temp_layer, output_layer)
+        monkey_patch_keywords(output_layer)
+        output_layer.keywords["layer_purpose"] = "aggregation"
```
Gustry: We use single quote 'abc' for strings

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
+        temp_layer = self.map_layer_cb.currentLayer()
+        try:
+            keyword_io = KeywordIO()
+            temp_layer.keywords = keyword_io.read_keywords(temp_layer)
+        except KeywordNotFoundError:
+            temp_layer.keywords['layer_purpose'] = 'aggregation'
+
+        # add keywords to output layer
+        copy_layer(temp_layer, output_layer)
+        monkey_patch_keywords(output_layer)
```
Gustry: We don't need to call this function

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
+        output_layer.keywords["layer_purpose"] = "aggregation"
+        output_layer.keywords["inasafe_fields"] = \
+            {"displaced_field": str(self.displaced_cb.currentField()),
```
Gustry: * Use single quote
* Do you need to use `str` ? We try to avoid it (a layer may have accents in the name and str will not like it)

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +102,62 @@ def __init__(self, parent=None):
...
+             "aggregation_name_field": str(self.aggregation_cb.currentField())}
+
+        # remove unnecessary fields & rename inasafe fields
+        _rename_remove_inasafe_fields(output_layer)
 
         try:
-            output_layer = self.minimum_needs(input_layer, field_name)
+            self.minimum_needs(output_layer)
         except ValueError:
```
Gustry: I think you can remove this ValueError.
Maybe it's better to catch `Exception` in case if there is an error and to display an error message.

myarjunar:
inasafe 4 0 0b0 minimum needs calculator_002
here is the screenshot of minimum needs calculator tool

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +100,57 @@ def __init__(self, parent=None):
...
+
+        # remove unnecessary fields & rename inasafe fields
+        _rename_remove_inasafe_fields(output_layer)
 
         try:
-            output_layer = self.minimum_needs(input_layer, field_name)
-        except ValueError:
+            self.minimum_needs(output_layer)
+        except:
             return
```
Gustry: We don't want to silently fail if there is an error. You should display an error message and log it using the LOGGER.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -45,7 +54,37 @@ def __init__(self, parent=None):
...
+
+        # get field that represent aggregation name
+        self.aggregation_name = QgsFieldComboBox()
+        self.aggregation_name.setFilters(QgsFieldProxyModel.String)
+
+        # get field that represent aggregation id
+        self.aggregation_id = QgsFieldComboBox()
+
+        # add Qgis combo box to window
+        self.gridLayout_2.addWidget(self.layer, 0, 1)
```
Gustry: You can add this combobox in the UI file and not in the code.
In QTDesigner, use a QComboxBox and use the `promote` function.

Gustry:
It would be very good if you can add a unit test : 
* Check if the button is not enabled if there is no layer loaded.
* Check if the button is enabled if there is at least one layer.
* Run the minimum need calculator nicely
* Check if the attribute table gets new fields.

Maybe you should also save the layer to a file at the end in the inasafe directory.
Think also to remove the monkey patching after the process (If we write the layer on the file system, we don't want to write these fake keywords).

safe/gis/vector/prepare_vector_layer.py
```diff
@@ -179,6 +179,15 @@ def _rename_remove_inasafe_fields(layer):
         fields = get_fields(
             layer.keywords['layer_purpose'])
 
+    # Add displaced_field definition to expected_fields
+    # for minimum needs calculator.
+    # If there is no displaced_field keyword, then pass
+    try:
+        if layer.keywords['inasafe_fields']['displaced_field']:
```
Gustry: You should use the `displaced_field` object : 
```
from safe.definitions.fields import displaced_field
...
if layer.keywords['inasafe_fields'][displaced_field['key']]
```

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +100,57 @@ def __init__(self, parent=None):
...
+        # monkey patching input layer to make it work with
+        # prepare vector layer function
+        temp_layer = self.layer.currentLayer()
+        temp_layer.keywords = {'layer_purpose': 'aggregation'}
+
+        # add keywords to output layer
+        copy_layer(temp_layer, output_layer)
+        output_layer.keywords['layer_purpose'] = 'aggregation'
+        output_layer.keywords['inasafe_fields'] = (
+            {'displaced_field': self.displaced.currentField(),
```
Gustry: The same for here for these two fields. Use the field object.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +100,57 @@ def __init__(self, parent=None):
...
+        output_layer = create_memory_layer(
+            output_layer,
+            self.layer.currentLayer().geometryType(),
+            self.layer.currentLayer().crs(),
+            self.layer.currentLayer().fields())
+
+        # monkey patching input layer to make it work with
+        # prepare vector layer function
+        temp_layer = self.layer.currentLayer()
+        temp_layer.keywords = {'layer_purpose': 'aggregation'}
```
Gustry: `from safe.definitions.layer_purpose import layer_purpose_aggregation` and then use `layer_purpose_aggregation['key']` instead of using the string 'aggregation'

safe/gui/ui/needs_calculator_dialog_base.ui
```diff
@@ -56,50 +47,208 @@
...
+     </property>
+    </widget>
+   </item>
   </layout>
  </widget>
  <customwidgets>
   <customwidget>
+   <class>QgsFieldComboBox</class>
+   <extends>QComboBox</extends>
+   <header>qgsfieldcombobox.h</header>
```
Gustry: I think the header is qgis.gui, as I showed you in my screenshot

safe/gui/ui/needs_calculator_dialog_base.ui
```diff
@@ -56,50 +47,208 @@
...
+  </customwidget>
+  <customwidget>
+   <class>QgsMapLayerComboBox</class>
+   <extends>QComboBox</extends>
+   <header>qgsmaplayercombobox.h</header>
```
Gustry: Same for here

safe/gui/ui/needs_calculator_dialog_base.ui
Gustry: I think the header is qgis.gui, as I showed you in my screenshot

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
         ok_button = self.button_box.button(QtGui.QDialogButtonBox.Ok)
         ok_button.clicked.connect(self.accept)
 
-    def minimum_needs(self, input_layer, population_name):
+    def update_button_status(self):
+        """Function to enable or disable the Ok button
```
Gustry: Add a full stop ad the end of the sentence.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
+            self.button_box.button(QtGui.QDialogButtonBox.Ok).setEnabled(True)
+        else:
+            self.button_box.button(QtGui.QDialogButtonBox.Ok).setEnabled(False)
+
+    def minimum_needs(self, input_layer):
         """Compute minimum needs given a layer and a column containing pop.
 
-        :param input_layer: InaSAFE layer object assumed to contain
+        :param input_layer: Vector layer assumed to contain
             population counts
```
Gustry: Add a full stop too at the end of the sentence.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
-        :type input_layer: read_layer
-
-        :param population_name: Attribute name that holds population count
-        :type population_name: str
+        :type input_layer: QgsVectorLayer
 
         :returns: Layer with attributes for minimum needs as per Perka 7
```
Gustry: Let's remove `as perka 7`. The user may not run the indonesian profile.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
-                add_ordered_combo_item(self.cboPolygonLayers, name, source)
-        # Now disable the run button if no suitable layers were found
-        # see #2206
-        ok_button = self.button_box.button(QtGui.QDialogButtonBox.Ok)
-        if found_flag:
-            self.cboPolygonLayers.setCurrentIndex(0)
-            ok_button.setEnabled(True)
-        else:
-            ok_button.setEnabled(False)
+        :param input_layer: Vector layer
```
Gustry: Full stop at the end of the docstring

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
 
-    # prevents actions being handled twice
-    # noinspection PyPep8Naming
-    @pyqtSignature('int')
-    def on_cboPolygonLayers_currentIndexChanged(self, index):
-        """Automatic slot executed when the layer is changed to update fields.
-
-        :param index: Passed by the signal that triggers this slot.
-        :type index: int
+        :return: New memory layer duplicated from input_layer
```
Gustry: Full stop at the end of the docstring

safe/gui/tools/minimum_needs/test/test_minimum_needs_tool.py
```diff
@@ -52,24 +56,60 @@ def tearDown(self):
...
+            'minimum_needs__rice': 2800,
+            'minimum_needs__drinking_water': 17500,
+            'minimum_needs__clean_water': 67000,
+            'minimum_needs__family_kits': 200,
+            'minimum_needs__toilets': 50}
+
+        self.assertDictEqual(expected_attributes, actual_attributes)
+
+    def test_ok_button(self):
+        """Test behaviour of Ok button
```
Gustry: Full stop and put `"""` on the same (because it fits with less than 80 chars)

safe/gui/tools/minimum_needs/test/test_minimum_needs_tool.py
```diff
@@ -52,24 +56,60 @@ def tearDown(self):
...
+        # Test Ok button without any input in the combo box
+        dialog = NeedsCalculatorDialog(PARENT)
+        ok_button = dialog.button_box.button(QtGui.QDialogButtonBox.Ok)
+        self.assertFalse(ok_button.isEnabled())
+
+        input_layer = QgsVectorLayer(
+            shapefile_path,
+            os.path.basename(shapefile_path),
+            'ogr')
+        QgsMapLayerRegistry.instance().addMapLayers([input_layer])
```
Gustry: You don't need to add the layer to the canvas while you are testing.

safe/gui/tools/minimum_needs/test/test_minimum_needs_tool.py
```diff
@@ -52,24 +56,60 @@ def tearDown(self):
...
+
+        # Test Ok button with layer and displaced field
+        # selected in the combo box
+        dialog.layer.setLayer(input_layer)
+        dialog.displaced.setField(u'displaced')
+        self.assertTrue(ok_button.isEnabled())
 
     # def Xtest_accept(self):
```
Gustry: Let's remove this old test if you can. We try to clean this legacy code.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
+            output_layer,
+            input_layer.geometryType(),
+            input_layer.crs(),
+            input_layer.fields())
+
+        # monkey patching input layer to make it work with
+        # prepare vector layer function
+        temp_layer = input_layer
+        temp_layer.keywords = {'layer_purpose':
+                                   layer_purpose_aggregation['key']}
```
Gustry: We usually write on the left as much as possible.
```
temp_layer.keywords = {
    'layer_purpose': layer_purpose_aggregation['key']}
```

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
+        else:
+            data_store = Folder(temp_dir(sub_dir=output_layer.name()))
 
-        output_layer.write_to_file(new_file)
+        data_store.add_layer(output_layer, output_layer.name())
 
-        new_layer = QgsVectorLayer(new_file, 'Minimum Needs', 'ogr')
         # noinspection PyArgumentList
-        QgsMapLayerRegistry.instance().addMapLayers([new_layer])
+        QgsMapLayerRegistry.instance().addMapLayers([data_store.layer
```
Gustry: Write also on the left.

safe/gui/tools/minimum_needs/test/test_minimum_needs_tool.py
myarjunar: I did this because I need to set the specific layer to the Qgis combo box (line 110). setLayer function will not set the layer if there is no layer on the canvas, I guess. I don't know if you have any other brilliant idea.

Gustry: Ah yes, you are right :)

myarjunar:
inasafe 4 0 0b0 minimum needs calculator_003

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
-        file_name = layer.source()
+        # remove unnecessary fields & rename inasafe fields
+        rename_remove_inasafe_fields(output_layer)
 
         try:
-            output_layer = self.minimum_needs(input_layer, field_name)
-        except ValueError:
+            self.minimum_needs(output_layer)
+        except Exception as e:
+            LOGGER.debug(e)
```
Gustry: Logging the error is not enough. We should also display a message box to the user.
The best would be to display an error based on the exception.
You can at least use `safe.utilities.qgis_utilities.display_critical_message_box' to say that something went wrong.

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +96,107 @@ def __init__(self, parent=None):
...
+        settings = QSettings()
+        default_user_directory = settings.value(
+            'inasafe/defaultUserDirectory', defaultValue='')
+
+        if default_user_directory:
+            path = os.path.join(default_user_directory, output_layer.name())
+            if not os.path.exists(path):
+                os.makedirs(path)
+            data_store = Folder(path)
+            data_store.add_layer(output_layer, output_layer.name())
```
Gustry: I think you need to remove this line. Because the layer will be added twice with the line 195

safe/gui/tools/minimum_needs/test/test_minimum_needs_tool.py
```diff
@@ -52,39 +56,58 @@ def tearDown(self):
...
-        """
+        """Test behaviour of the minimum needs function."""
         dialog = NeedsCalculatorDialog(PARENT)
-        layer = shapefile_path
+        layer = QgsVectorLayer(
+            shapefile_path,
+            os.path.basename(shapefile_path),
+            'ogr')
+        if layer.isValid():
+            layer = dialog.prepare_new_layer(layer)
```
Gustry: I think you shouldn't call this function from test.

safe/gui/tools/minimum_needs/test/test_minimum_needs_tool.py
```diff
@@ -52,39 +56,58 @@ def tearDown(self):
...
-    #         os.path.dirname(shapefile_path),
-    #         'ogr')
-    #     layer = None
-    #     dialog = MinimumNeeds(PARENT)
-    #     dialog.accept()
-
+        layer.keywords = {
+            'layer_purpose': layer_purpose_aggregation['key'],
+            'inasafe_fields': {displaced_field['key']: attribute}}
+        rename_remove_inasafe_fields(layer)
```
Gustry: The same here, I think you shouldn't call this function from the test.

The idea of writing test is to simulate the user, to write whet the user will do using the tool.
* Open the dialog, check that the run button is disabled (because there is no layer)
* Add a layer to the canvas, check that combobox are working well and the run button is enabled.
* Simulate the click on the run button
* Check the output layer.

In some other other words, we shouldn't call function which are inside the run such as `prepare_new_layer(layer)`, `minimum_needs(layer)` ...

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
```diff
@@ -61,148 +98,113 @@ def __init__(self, parent=None):
...
-        file_name = layer.source()
 
         try:
-            output_layer = self.minimum_needs(input_layer, field_name)
-        except ValueError:
+            self.result_layer = self.minimum_needs(self.layer.currentLayer())
+        except Exception as e:
+            LOGGER.debug(e)
+            display_critical_message_box(
+                title='Error while calculating minimum needs', message=e)
```
Gustry: For every strings, you should use the `tr` function from safe.utilities.i18n. This function will allow us to have a translated version of string. You can easily fix this one.

Gustry: But there is another issue.
`e` is a python exception object.
You are trying to give the exception as a message but the function expects a string : https://github.com/inasafe/inasafe/blob/develop/safe/utilities/qgis_utilities.py#L259

I just tried this in the QGIS python console : 
```
from safe.utilities.qgis_utilities import display_critical_message_box
import logging

LOGGER = logging.getLogger('InaSAFE')

try:
    raise TypeError
except Exception as e:
    LOGGER.exception(e)
    display_critical_message_box(title='Error while calculating minimum needs', message=e)
```
* The logger didn't log the exception correctly. I though it was going to be ok. Maybe you should just log an error.
* This code will generate another exception because you tried to to give an exception instead of a string. Let's only display that some thing went wrong.
```
display_critical_message_box(title='Error while calculating minimum needs', message=e)
  File "/Users/etienne/.qgis2/python/plugins/inasafe/safe/utilities/qgis_utilities.py", line 262, in display_critical_message_box
    QMessageBox.critical(parent, title, message)
TypeError: arguments did not match any overloaded call:
  QMessageBox.critical(QWidget, QString, QString, QMessageBox.StandardButtons buttons=QMessageBox.Ok, QMessageBox.StandardButton defaultButton=QMessageBox.NoButton): argument 3 has unexpected type 'exceptions.TypeError'
  QMessageBox.critical(QWidget, QString, QString, int, int, int button2=0): argument 3 has unexpected type 'exceptions.TypeError'
  QMessageBox.critical(QWidget, QString, QString, QString, QString button1Text=QString(), QString button2Text=QString(), int defaultButtonNumber=0, int escapeButtonNumber=-1): argument 3 has unexpected type 'exceptions.TypeError'
```
You can use `e.__class__.__name__` to get the name of the exception as a string. It might be enough. @ismailsunni ?

I just remember that the postprocessor function will return a boolean with a error message.
https://github.com/inasafe/inasafe/blob/develop/safe/impact_function/postprocessors.py#L62
I think you should check if the boolean if True. If not true, you can display this error message.

@ismailsunni Any ideas where to get the error message ?

timlinux:
@myarjunar 

inasafe 4 0 0b0 minimum needs calculator_003

Please take note of our HIG rules:

* Displaced people Layer -> Displaced people layer
* Same rule as above for all labels (Upper case first word, all lower case for rest)
* The top label "in specified aggregation area" -> "in the specified aggregation areas" (plural)

timlinux:
@myarjunar please try to get this merged tomorrow so that it can be tested and we can get some feedback on it.

Gustry:
> @myarjunar please try to get this merged tomorrow
Ahah, I keep adding new comments xD.
Sorry @myarjunar, at least your PR will be perfect now. I hope you learn our coding standards ;-)

safe/gui/tools/minimum_needs/needs_calculator_dialog.py
Gustry: Ok, I just merged an helper to get the name of the exception and traceback if we have an exception -> https://github.com/inasafe/inasafe/blob/develop/safe/utilities/utilities.py#L97

So now, you can use the first the return of the postprocessor function if the result is different than `True`.
Then you can also add all your code between a `try: .. except:...` and display the exception nicely with this new helper. You can ask @ismailsunni if you have any question.

timlinux:
LGTM

Gustry:
Thanks @myarjunar

Issue #3658, opened 2017-01-22
gis.raster test fails in windows


timlinux:
### Problem

Running the test I get a key error for ['HOME']

### Expected Behaviour

No error

### Environment
* InaSAFE : dev 
* QGIS : 2.14.11
* OS :  Windows 10

### Proposed Solution

Replace ``['HOME']`` with ``os.expanduser('~')`` - see https://github.com/botify-labs/python-simple-workflow/issues/41

PR #3659, opened 2017-01-22
Fix error expanding os ['HOME']

timlinux:
See #3658


Gustry:
You should use `QgsApplication.qgisSettingsDirPath()` I think

timlinux:
@gustry ah even better

PR #3662, opened 2017-01-23
Update the map legend

Gustry:
* The legend : 
screen shot 2017-01-23 at 15 20 41

* The report :
screen shot 2017-01-23 at 15 21 02

@lucernae : 
* We need to check why we don't get the same rounding. Numbers should match between the legend and the report.
* You need to add a new row I think for not exposed according to the new diagram : 
screen shot 2017-01-23 at 15 22 49

### DIFF
--- a/safe/utilities/rounding.py
+++ b/safe/utilities/rounding.py
@@ -6,6 +6,36 @@ __email__ = "info@inasafe.org"
 __revision__ = '$Format:%H$'
 
 
+def round_affected_number(
+        number,
+        enable_rounding=False,
+        use_population_rounding=False):
+    """Tries to convert and round the number.
+
+    Rounded using population rounding rule.
+
+    :param number: number represented as string or float
+    :type number: str, float
+
+    :param enable_rounding: flag to enable rounding
+    :type enable_rounding: bool
+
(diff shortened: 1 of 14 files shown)


lucernae:
@Gustry 
I remembered in the previous report, exposure type other than population, will not have rounding.
Like building count or roads length. We need to make sure that @timlinux and @Charlotte-Morgan are aware of this.

Gustry:
Yes, you are right.

I was just reading the markdown document about rounding and I did not see any conditions.
https://github.com/inasafe/inasafe/blob/develop/docs/reporting-standards.md#legend-label-rounding

@Charlotte-Morgan Can ew have your input ? Thanks.

Gustry:
Ok, I have disabled the rounding until we have some inputs.
I would like to merge this PR and work with @lucernae to include not exposed in the report.

safe/utilities/rounding.py
```diff
@@ -6,6 +6,36 @@
 __revision__ = '$Format:%H$'
 
 
+def round_affected_number(
```
lucernae: Thank's for moving this ;)

lucernae:
LGTM.

I will use units semantic too in the reporting and postprocessors.

PR #3663, opened 2017-01-23
Add displaced postprocessor

lucernae:
See issue: #3664


safe/definitions/post_processors.py
```diff
@@ -248,6 +313,8 @@ def post_processor_affected_function(**kwargs):
                 ],
             }]
     },
+    # output is described as ordered dict because the order is important
+    # and the postprocessor produce two fields.
```
Gustry: Do you mean that the postprocessor will produce two fields ? I need to check deeper how you could do that

lucernae: Yes, I added an ability for post processor to be able to produce more than one field.
In this case, for gender ratio, it will produce female and male count.
We don't use male count in reporting at the moment, I put it there just in case, since it uses the same input.
But now that I'm thinking, I might be protested since I assume male ratio is 1-female ratio :(.
Perhaps we should just delete the male count?

Gustry: I think it's ok. We are producing rich reporting and we are not limited in the number of fields.

Issue #3654, opened 2017-01-20
Earthquake analysis


Charlotte-Morgan:
# problem
We need to ensure that legacy details are retained in the migration of InaSAFE3.5.3 earthquake analysis to the new version 4 architecture.

In version 3.5.x we had 3 earthquake impact functions for impact on population: pager model, ITB fatality model, ITB bayesian fatality model  and one for impact on buildings: earthquake building

![image](https://cloud.githubusercontent.com/assets/5743135/22139884/9d601d02-df1b-11e6-9df2-e4ae25fd1af6.png)

The earthquake_building analysis supported Nexis building data with specific attributes and building valuations .... 
![image](https://cloud.githubusercontent.com/assets/5743135/22139979/430a9250-df1c-11e6-8c8c-b4ad2af2a18f.png)

In version 4.0b2 we have a single impact function which offers 3 earthquake on population functions; however the names of these have changed from version 3.5.x. This could lead to confusion as the names are quite similar to each other

![image](https://cloud.githubusercontent.com/assets/5743135/22140095/e5a3932c-df1c-11e6-9c44-8be445bc4d8e.png)

![image](https://cloud.githubusercontent.com/assets/5743135/22140083/d5f8111e-df1c-11e6-81b9-05a62349449f.png)

![image](https://cloud.githubusercontent.com/assets/5743135/22140108/f8d7cfb2-df1c-11e6-9033-14aae3f18bdf.png)

# proposed solution 1
 - [x] update the bayesian_fatality_rates to be called ITB_bayesian_fatality_rates
![image](https://cloud.githubusercontent.com/assets/5743135/22140108/f8d7cfb2-df1c-11e6-9033-14aae3f18bdf.png)
- [ ] make sure ITB_bayesian fatality function is the one used by default in both InaSAFE desktop and InaSAFE realtime @lucernae 

# proposed solution 2
Ensure the legacy analysis on NEXIS building data with valuations is retained in the code

# cc
@timlinux @hadighasemi @dynaryu @RikkiWeber

ismailsunni:
Hi @Charlotte-Morgan 

> Ensure the legacy analysis on NEXIS building data with valuations is retained in the code

Reading the code and recalling the history, this NEXIS is used to indicate whether the IF can run more calculation or not.

The work flow is:
[...]

PR #3665, opened 2017-01-23
change the EQ output raster to displaced people

Gustry:
* Default EQ set to ITB bayesian #3654 
* Rename function according to #3654 and #3497 

screen shot 2017-01-23 at 17 21 00

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -1203,7 +1203,10 @@ class ImpactFunction(object):
 
         self.set_state_process('exposure', 'Exposed people')
         exposed, self._exposure_impacted = exposed_people_stats(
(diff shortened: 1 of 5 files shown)


safe/impact_function/impact_function.py
```diff
@@ -517,6 +524,27 @@ def datetime(self):
...
+
+    @earthquake_function.setter
+    def earthquake_function(self, function):
+        """Set the earthquake function to use.
+
+        :param function: The earthquake function to use.
+        :type function: str
+        """
+        if function not in [model['key'] for model in EARTHQUAKE_FUNCTIONS]:
+            raise Exception(tr('Earthquake function not recognised.'))
```
ismailsunni: What about using custom exception here?

PR #3668, opened 2017-01-24
use rounding in the map legend

Gustry:


timlinux:
LGTM - might be nice to include a little before vs after screenie for these PR's

PR #3670, opened 2017-01-24
fix union algo from https://github.com/qgis/QGIS/pull/4037

Gustry:
This is the backport from @wonder-sk in QGIS.
This is not my code. I think we can just merge it.

### DIFF
--- a/safe/gis/vector/clip.py
+++ b/safe/gis/vector/clip.py
@@ -68,6 +68,8 @@ def clip(layer_to_clip, mask_layer, callback=None):
 
     # Begin copy/paste from Processing plugin.
     # Please follow their code as their code is optimized.
+    # The code below is not following our coding standards because we want to
+    # be able to track any diffs from QGIS easily.
 
     # first build up a list of clip geometries
     clip_geometries = []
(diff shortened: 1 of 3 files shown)


timlinux:
Since there is not much code, do you think it is worth updating it to follow our naming conventions? It is not very pythonic like this :-P I guess you might be motivated to keep it verbatim so that we can diff it against upstream. I'll leave the call to you, but at least let's add a comment indicating why we did not follow standards in this module. LGTM once you have made that call....

Gustry:
Yes you are right.
I tried to stay close from the original code in case if we have to do this kind of things. I just did pep8 and some other cleaning.

I made a comment at the beginning : https://github.com/inasafe/inasafe/blob/develop/safe/gis/vector/union.py#L97
and at the end : https://github.com/inasafe/inasafe/blob/develop/safe/gis/vector/union.py#L278

timlinux:
> I made a comment at the beginning :

Ah sorry I missed that...ok

Gustry:
Ok I added some new comments about coding standards.

PR #3671, opened 2017-01-24
Exception

Gustry:
@myarjunar It might be good to use this `humanise_exception` function. It will give the name and the traceback of the exception.

### DIFF
--- a/safe/utilities/utilities.py
+++ b/safe/utilities/utilities.py
@@ -146,70 +146,6 @@ def humanise_seconds(seconds):
             days, hours, minutes))
 
 
-def reorder_dictionary(unordered_dictionary, expected_key_order):
-    """Reorder a dictionary according to a list of keys.
-
-    .. versionadded: 3.4
-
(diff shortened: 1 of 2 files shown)


Gustry:
@ismailsunni Feel free to merge. It could be nice if @myarjunar can use it in the minimum need calculator.

timlinux:
Funny we used to have something like this in utilities or somewhere that made a nice message out of an exception. I took a quick look but can't remember where it is - might be worth scratching around for some more.... LGTM for now though....

safe/utilities/utilities.py
```diff
@@ -49,7 +49,6 @@
 def get_error_message(exception, context=None, suggestion=None):
```
Gustry: Yes, @timlinux I think you are talking of that one. But this one is doing too much as I don't want the the Messaging system.

timlinux: Ok no problem thanks.

PR #3672, opened 2017-01-24
Fix #3667 - add petaJakarta flood thresholds

### LINKED ISSUE #3667: We need to add PetaJakarta flood thresholds
### Problem

We discussed with @Charlotte-Morgan having more sensible hazard thresholds for floods. Currently You can choose between generic (High, Medium, Low) classes and Flood (wet, dry) classes. It would be much better to also provide classes similar to those used in PetaJakarta:

screen shot 2017-01-24 at 10 04 49 am

### Expected Behaviour

User should be able to choose standard flood thresholds for continuous data.

### Proposed Solution

@timlinux to add a new set of flood thresholds based on PetaJakarta.

### CC

@Charlotte-Morgan @ismailsunni

timlinux:
@charlotte-morgan PetaJakarta doesnt give class names for the different thresholds - I have used high/medium/low for now - feel free to suggest better names and also any tweaks to the descriptions you may like.

capture

capture

### DIFF
--- a/safe/definitions/hazard_classifications.py
+++ b/safe/definitions/hazard_classifications.py
@@ -294,7 +294,7 @@ volcano_hazard_classes = {
 
 flood_hazard_classes = {
     'key': 'flood_hazard_classes',
-    'name': tr('Flood classes'),
+    'name': tr('Flood wet/dry classes'),
     'description': tr(
         'This is a binary classification for an area. The area is either '
         '<b>wet</b> (affected by flood water) or <b>dry</b> (not affected '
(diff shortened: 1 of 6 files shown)


timlinux:
See #3667

safe/definitions/hazard_classifications.py
```diff
@@ -220,9 +220,9 @@
     ]
 }
 
-flood_hazard_classes = {
-    'key': 'flood_hazard_classes',
-    'name': tr('Flood classes'),
+flood_wet_dry_hazard_classes = {
+    'key': 'flood_wet_dry_hazard_classes',
```
Gustry: AFAIK, changing an existing key is going to break every existing datasets using the old `flood_hazard_classes` hazard classification.
I expect some python exception to be raised as classes are not the same. Maybe we should check first.

timlinux:
Thanks for your comments @Gustry 

I changed the keywords in the test data now. Given your comments, I am actually surprised that all the tests were passing...lets see if changing the keyword in the test data breaks anything...

timlinux:
heh now it fails :-P

timlinux:
@Gustry @Charlotte-Morgan OK for me to merge?

Charlotte-Morgan:
looking

Gustry:
Looks good to me, thanks !

Charlotte-Morgan:
sorry - we do not seem to have any petajakarta data to test it on - is it possible you can add some to the "add test data". If not, I'm happy in principle - the help file classes are still empty for all hazards

timlinux:
> sorry - we do not seem to have any petajakarta data to test it on - is it possible you can add some to the "add test data". If not, I'm happy in principle - the help file classes are still empty for all hazards

I tested with continuous raster data - I should test with continuous vector data too. We should also update the PetaJakarta downloader to apply these thresholds. I will make separate tickets for these things. In the mean time I will see if I can do a query against an old flood date to get some test data for you.

safe/definitions/hazard_classifications.py
timlinux: Just a note that this does not break anything....I updated all references to it.

lucernae:
Some test data
[2016111123-6-rw.zip](https://github.com/inasafe/inasafe/files/744068/2016111123-6-rw.zip)

timlinux:
@lucrnae thanks!

timlinux:
@lucernae that dataset has no records in it (same as what you get from the PetaJakarta tool). Can you please get archive data from 1st and 2nd of April 2016 8am for testing (the time during the flood hack).

safe/gui/tools/wizard/test/test_keyword_wizard.py
```diff
@@ -1689,7 +1689,11 @@ def test_continuous_raster(self):
             'license': source_license,
             'scale': source_scale,
             'source': source,
-            'thresholds': {'dry': [0.0, 1.0], 'wet': [1.0, 999999.0]},
+            'thresholds': {
+                'high': [1.5, 999999.0],
```
ismailsunni: I know it's a dictionary, but may be it's better to write it ascending or descending :)

timlinux:
screencapture

Here you can see it working with flood on roads (all of JK)

timlinux:
2017-02-20 11_14_44-qgis 2 14 11-essen - dki_jakarta

Gustry:
Finally good !
Looks good to me, thanks !

timlinux:
2017-02-20 11_25_02-desktop

Here it is with flood on buildings.

PR #3673, opened 2017-01-24
check if the analysis extent is valid from the aggregation layer

Gustry:
@timlinux From your problem 'No feature in extent' with DKI jakarta project. Your subdistrict layer was invalid.


timlinux:
Nice thanks! LGTM

PR #3677, opened 2017-01-25
revised population exposure concepts

### LINKED ISSUE #3649: Check the definitions again
We need to check definitions and update concepts.py again before implementing new exposure concepts. 
- [ ] exposed
- [ ] affected
- [ ] displaced
- [ ] evacuated
- [ ] impacted
- [ ] killed

### LINKED ISSUE #3657: We need a note about the possibility of errors of omissions
### Problem

@charlotte-morgan raised the issue that in some cases it may be possible for a building to extend beyond an aggregation (where it is tagged as not flooded) into a hazard pixel / polygon that is not in an aggregation area - and in so doing be incorrectly counted as not flooded.

### Expected Behaviour

Users should understand the limitations of the analysis environment.

### Proposed Solution

Add a note in limitations indicating that in some cases exposure features may be incorrectly classified if hazards are in an adjacent area.

### LINKED ISSUE #3675: Notes: check rounding notes are on all reports
In Version 4.0 we apply the standard rounding to all exposure types.
We need to check that this is documented in the notes for maps and reports where it is applied.

Charlotte-Morgan:
fix #3649 fix #3657 fix #3675

### DIFF
--- a/safe/definitions/concepts.py
+++ b/safe/definitions/concepts.py
@@ -127,8 +127,8 @@ concepts = {
         'key': 'exposed_people',
         'description': tr(
             'People who are present in hazard zones and are thereby subject '
-            'to potential losses. In InaSAFE, people who are exposed are those '
-            'people who are within the extent of the hazard.'),
+            'to potential losses. In InaSAFE, people who are exposed are '
+            'those people who are within the extent of the hazard.'),
         'citations': [
             {
                 'text': tr(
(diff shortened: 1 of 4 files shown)


safe/definitions/concepts.py
```diff
@@ -167,7 +219,25 @@
...
+                'link': 'http://www.preventionweb.net/files/'
+                        '45462_backgoundpaperonterminologyaugust20.pdf'
+            }
+        ],
+    },
+    'relocated_people': {
+        'key': 'relocated_people',
+        'description': tr(
+            'Relocated people are people who, for different reasons or '
+            'circumstances because of risk or disaster, have movedm '
```
Gustry: Movedm?

Charlotte-Morgan:
thanks - i've fixed it

Charlotte-Morgan:
i think my attempts to make the test pass are making things worse :(

Gustry:
I'm going to have a look to your PR

PR #3680, opened 2017-01-25
Refine infographics

lucernae:
- Add svg to png converter (workaround for rendering svg)
- Convert svg icons to png
- Move hazard class label below donut chart
- Add multiple layers to show in impact map pdf
- Add tests for infographic
- Add, rounding now applied to all exposure types
  to infographics and impact report table
- Better documentation for donut svg generations
- Add thousand separator for numbers


lucernae:
Need to wait for displaced postprocessor to correct the numbers.

timlinux:
@lucernae can you make it a habit to share a screenshot or two in your PR's please? It makes it much easier to understand what is going on / changed.

timlinux:
Also I see you added lots of PNG's - I thought we were going to convert SVG's on the fly to PNG  if needed.

lucernae:
Well, no one commented after 6 days, so I get lonely :(.
Ok, I will share screenshot if needed.

For the PNG's, the files is only for static file, since I need to show something for now.
I thought the converter is only for dynamic svgs (like the chart in this case).

PR #3681, opened 2017-01-25
update map legend examples

### LINKED ISSUE #3653: Standards: map legends - syntax and rounding requirements
# problem

The map legend standards are a bit confusing.
In https://github.com/inasafe/inasafe/blob/develop/docs/reporting-standards.md it states 

> Map legend classes will have labels. The label should complement the legend title and should use the following syntax:
> 
> <label> <threshold> <count> <unit>
> 
> Label, refers to the class in the table above, i.e. affected, not affected, high hazard zone etc. threshold is the dynamic value which defines the class boundary i.e. the value of 0.7m for people affected by tsunami. count is the number of exposure elements in each class. Counts will be included in the legend label in brackets. These values will be rounded according to the legend label rounding principles. In some cases count will include a unit. The table below lists the units for each exposure type.

However the examples do not show this. The examples do not have rounded values and they do not include the threshold.

# proposed solution
- [ ] confirm that legends will alway have threshold or only if hazard is continuous raster data
- [ ] Update standards to show examples of legends

@felix-yew  - can you please review the issue about including thresholds and prepare some new examples
I will then update the standards document to reflect requirements.

# cc 
@Gustry @timlinux @samnawi

Charlotte-Morgan:
fix #3653


Charlotte-Morgan:
oh yay - I fixed this one by myself :)

Gustry:
So we can merge it ?
Thanks @Charlotte-Morgan

PR #3684, opened 2017-01-25
Fix string error on minimum needs calculator

myarjunar:


Gustry:
Looks good to me, thanks !

PR #3687, opened 2017-01-26
add not_exposed class

Gustry:


safe/definitions/hazard_classifications.py
```diff
@@ -9,7 +9,8 @@
 from safe.utilities.i18n import tr
 from safe.definitions.units import unit_centimetres, unit_miles_per_hour, \
```
lucernae: beautify this, perhaps?

lucernae:
Other than that, LGTM!

PR #3702, opened 2017-01-26
disable not_exposed in the map for landcover and population exposures

Gustry:
Same hazard, but different exposure : 
screen shot 2017-01-26 at 11 00 35

CC : @Charlotte-Morgan @felix-yew

### DIFF
--- a/safe/definitions/exposure.py
+++ b/safe/definitions/exposure.py
@@ -148,6 +148,7 @@ exposure_population = {
         elderly_count_field
     ],
     'layer_modes': [layer_mode_continuous],
+    'display_not_exposed': False,
     'layer_legend_title': tr('Number of people'),
     'measure_question': tr('how many')
 }
@@ -206,6 +207,7 @@ exposure_road = {
         feature_rate_field
     ],
     'layer_modes': [layer_mode_classified],
+    'display_not_exposed': True,
     'layer_legend_title': tr('Length of roads'),
     'measure_question': tr('what length of')
 }
@@ -264,6 +266,7 @@ exposure_structure = {
         feature_rate_field
     ],
     'layer_modes': [layer_mode_classified],
+    'display_not_exposed': True,
     'layer_legend_title': tr('Number of buildings'),
     'measure_question': tr('how many')
 }
@@ -302,6 +305,7 @@ exposure_place = {
         exposure_name_field
     ],
     'layer_modes': [layer_mode_classified],
+    'display_not_exposed': True,
     'layer_legend_title': tr('Number of places'),
     'measure_question': tr('how many')
 }
@@ -361,6 +365,7 @@ exposure_land_cover = {
         feature_rate_field
     ],
     'layer_modes': [layer_mode_classified],
+    'display_not_exposed': False,
     'layer_legend_title': tr('Area of landcover'),
     'measure_question': tr('what area of')
 }
--- a/safe/definitions/hazard_classifications.py
+++ b/safe/definitions/hazard_classifications.py
@@ -32,6 +32,8 @@ __revision__ = '$Format:%H$'
 # We do not include it in the classes below because we do not want the user
 # to be presented with not exposed in the keywords when setting up
 # their classes.
+# This class is not displayed if it's a polygon exposure (landcover and
+# population) in the legend.
 not_exposed_class = {
     'key': 'not exposed',
     'name': tr('Not exposed'),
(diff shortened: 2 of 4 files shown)


Gustry:
@lucernae @ismailsunni Can you review this PR ?
I'm little bit confused about thousand separator. It's was not working well on my French OS.

@lucernae You should be able to use `format_number` for the reporting. It will add the thousand separator and apply the rounding if needed.

safe/impact_function/provenance_utilities.py
```diff
@@ -28,17 +28,17 @@ def get_map_title(hazard, exposure, hazard_category):
     :rtype: str
     """
     if hazard == hazard_generic:
-        map_title = tr('{exposure_name} affected'.format(
+        map_title = tr(u'{exposure_name} affected'.format(
```
ismailsunni: Perhaps I was wrong. But should we do tr first before doing the formating? I know I am the one who write the code, but I just realised that probably I was wrong.

felix-yew:
Hi, @Gustry - I think for the Dry zone in the roads impacted should be in Green colour. Because we use Green colour in the Dry zone when the exposures are buildings and/or roads ~ cmiiw.

safe/utilities/test/test_rounding.py
```diff
@@ -41,6 +42,29 @@ def test_population_rounding(self):
...
+        """Test formatting integer"""
+        number = 10000000
+        lang = os.getenv('LANG')
+        formatted_int = add_separators(number)
+        if lang == 'id':
+            expected_str = '10.000.000'
+        else:
+            expected_str = '10,000,000'
+        message = 'Format integer is not valid'
+        assert (formatted_int == expected_str or
```
ismailsunni: Any reason not to use self.assert?

ismailsunni:
Btw, what's the problem in your French OS?

safe/utilities/test/test_rounding.py
Gustry: Copy/paste from the old test !

Gustry:
> Btw, what's the problem in your French OS?

The thousand separator is not used by default with the `format` function from python. I had to use `monetary=True`.

```
>>> import locale
>>> locale.setlocale(locale.LC_ALL, 'en_US')
'en_US'
>>> locale.format("%d", 1255000, grouping=True)
'1,255,000'
```
If I swtich to french, it doesn't work, except if I had `monetary=True`

safe/impact_function/provenance_utilities.py
Gustry: Ok, I will work on that in another PR. Let's keep this one for reporting standards

Gustry:
@felix-yew I though I was following the standards : 
https://cloud.githubusercontent.com/assets/5743135/17542495/6b968a16-5ef3-11e6-8622-af1cab2e0a5f.png
We can check that together to fix it.

Gustry:
@ismailsunni Can i merge it ?
I will work in another PR for unicode and thousand separator.

Charlotte-Morgan:
hi @Gustry - I am happy with this. I discussed with @felix-yew and @ivanbusthomi and we agree that you have applied the standard. :)
Some confusion around the use of dry (when there are 3 hazard classes) and the fact that we only display the wet class in binary data so it looks like the dry class is actually not exposed (but its within the extent of the hazard data).
FYI - if you change the key words you can make the binary data not consider 0 dry and get the not exposed colour in more places.

Gustry:
> I discussed with @felix-yew and @ivanbusthomi and we agree that you have applied the standard. 

It's very nice to work on that, when we understand what is exactly the standard that we have to follow.

> Some confusion around the use of dry (when there are 3 hazard classes) and the fact that we only display the wet class in binary data so it looks like the dry class is actually not exposed (but its within the extent of the hazard data).

I agree it brings some confusion. But we are following the hazard classification which have been set in the wizard : wet and dry. If the user sets a value mapping with the dry class, they will get a impact layer with the dry class.

> FYI - if you change the key words you can make the binary data not consider 0 dry and get the not exposed colour in more places.

Correct. Is-it ok like this ?

@felix-yew Maybe you can check your python skills by updating the color ? 😀 
Everything is driven by definitions, so it should be easy for your team to do this kind of change.
It's this line : https://github.com/inasafe/inasafe/blob/develop/safe/definitions/hazard_classifications.py#L289 if you want to do it. You can check with @ivanbusthomi to help you and make a pull request.
Whem you have done your change, you can check in the InaSAFE help window from Tim that your color is correct : 
screen shot 2017-01-27 at 09 43 46

Charlotte-Morgan:
i'm happy the way it is

PR #3710, opened 2017-01-26
enable a test about minimum needs with a vector layer

Gustry:
This test is green on my mac @lucernae 
But travis is not happy

I need to check that


Gustry:
I need to wait for the displacement rate.

And it seems that we have some difference between QGIS 2.14 and QGIS 2.18

lucernae:
@Gustry Do you need my help on this? I'm not sure what's the problem?
Perhaps we can change the test file to use Flood Vector? Flood hazard have displacement rate.

timlinux:
Current status: @Gustry needs too validate the numbers produced by Travis to see if they are correct, Then we can merge,

PR #3718, opened 2017-01-27
add thousand separator

Gustry:
* FR
screen shot 2017-01-30 at 14 16 31

* EN
screen shot 2017-01-30 at 14 17 35

* ID
screen shot 2017-01-30 at 14 18 28

@lucernae You should be able to use `format_number` in the report. It adds the thousand separator and apply the rounding if needed.

### DIFF
--- a/safe/utilities/rounding.py
+++ b/safe/utilities/rounding.py
@@ -1,9 +1,98 @@
+# coding=utf-8
 from math import ceil
+
+from safe.utilities.i18n import locale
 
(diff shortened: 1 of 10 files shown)


timlinux:
LGTM thanks @Gustry

safe/utilities/rounding.py
```diff
@@ -1,11 +1,100 @@
+# coding=utf-8
 from math import ceil
```
ismailsunni: Docstring?

Issue #3553, opened 2017-01-09
Percentage of evacuated people is missing in V4


Gustry:
### Problem

In InaSAFE 3, we had the number of people who need evacuation based on a percentage.
This feature is missing in the V4.

Gustry:
@lucernae added `displaced_rates` in each hazard classification.
The thing is that the user can't update this value for now as it's hard coded in definitions.

timlinux:
For me it is ok to hard code the displacement rates for v4.0. I think we must just make sure they are clearly documented in definitions help so that if users want to understand what displacement rates were used they can easily access that information.

Gustry:
I think we can close this issue.
 Displacement rate has been implemented.

PR #3722, opened 2017-01-27
Added displacement rates for all hazard classes. See #3553

timlinux:


timlinux:
CC @lucernae and @Gustry

Charlotte-Morgan:
Hi @timlinux @Gustry @lucernae 
its great to see displacement rates included in the hazard classes now.
From what I see, and ignoring earthquake which is special, its only flood classes that have a displacement rate calculated and that is back to the 1%. Does this mean that no other hazard classes / impact functions had a % applied to the 'affected' ?
Ie everyone in a hazard class with affected = true was 'affected' and had minimum needs calculated for them?

safe/definitions/hazard_classifications.py
```diff
@@ -58,10 +58,10 @@
             'affected': True,
             'description': tr('The locations having the highest impact.'),
             'string_defaults': ['high'],
+            'displace_rate': 0.0,
```
timlinux: displacement

ismailsunni: Did you review your own PR @timlinux ? :)

lucernae:
To reply @Charlotte-Morgan :
Minimum needs now calculated from displacement number. So if it is zero (for other hazard classification other than floods on population), all minimum needs number (also vulnerabilities and age) will be zero. Then I think we should hide minimum needs section if that's the case and have some explanatory text that no minimum needs section, since displacement number is zero.

timlinux:
Great thanks @lucernae and yes lets put in placeholder text when there are minimum needs - something like:

> There is no minimum needs report displayed here because either
> a) there are no displaced people expected within the analysis area; or
> b) the hazard classes used for this analysis do not have displacement rates defined.

> If you believe this is an error, please submit an issue on the InaSAFE issue tracker at http://github.com/inasafe/inasafe/issues

How does that sound for you @Charlotte-Morgan

timlinux:
@Charlotte-Morgan 

> From what I see, and ignoring earthquake which is special, its only flood classes that have a displacement rate calculated and that is back to the 1%. Does this mean that no other hazard classes / impact functions had a % applied to the 'affected' ?
Ie everyone in a hazard class with affected = true was 'affected' and had minimum needs calculated for them?

I am going to need to go and poke around in the old IF's to see how displacement rates were computed. I suspect a ratherblanket policy was applied but will need to review the old code before I can verify that.

safe/definitions/hazard_classifications.py
timlinux: I did indeed :-)

Issue #3732, opened 2017-01-30
EQ Raster on Population raster : raster symbology


Gustry:
### Problem
With an EQ raster on population raster, InaSAFE V4 produces a raster of displaced people.

This raster doesn't have any symbology for now.

We agreed to display the raster with a binary style.
Can you provide the style (colors) that we should use ?

Charlotte-Morgan:
hi @Gustry - I think @ivanbusthomi has already supplied this. Maybe to @lucernae 
If not; I'm passing this to @ivanbusthomi to follow up

Charlotte-Morgan:
@ivanbusthomi - if this is finished - please close the ticket

Gustry:
I didn't get any feedback on this one.
How should we style the exposed raster (binary style) and the vector aggregation ?

Charlotte-Morgan:
maybe we are all stuck with key error :(
@ivanbusthomi  - please supply the symbology we developed for Pidie.

ivanbusthomi:
This is the symbology that we did for the Pidie Earthquake raster on raster population.
![image](https://cloud.githubusercontent.com/assets/7351977/23004402/18cdc166-f429-11e6-8f4e-7bdc859ca6c8.png)

Zip containing the .qml files [Aceh_EarthquakeOnPopulation_Subdistrict_WGS84.zip](https://github.com/inasafe/inasafe/files/778899/Aceh_EarthquakeOnPopulation_Subdistrict_WGS84.zip)

Gustry:
The raster styling has been done in the PR above.

Issue #3719, opened 2017-01-27
Assertion Error - earthquake raster on population raster


Charlotte-Morgan:
### Problem
The InaSAFE analysis stops with an assertion error and writes a DEBUG folder (empty) when not in debug mode
### Expected Behaviour
InaSAFE analysis will run and be happy with no error

### Environment
* InaSAFE : 4.0.0b0
* QGIS : 2.14.0
* OS : windows
* Data used :
  * Exposure : Aceh population raster
  * Hazard : Pidie Earthquake
  * Aggregation : none

### Error Message
Problem

The following problem(s) were encountered whilst running the analysis.

AssertionError
Suggestion

You can try the following to resolve the issue:

Check that you have the latest version of InaSAFE installed - you may have encountered a bug that is fixed in a subsequent release.
Check the InaSAFE documentation to see if you are trying to do something unsupported.
Report the problem using the issue tracker at https://github.com/inasafe/inasafe/issues. Reporting an issue requires that you first create a free account at http://github.com. When you report the issue, please copy and paste the complete contents of this panel into the issue to ensure the best possible chance of getting your issue resolved.
Try contacting one of the InaSAFE development team by sending an email to info@inasafe.org. Please ensure that you copy and paste the complete contents of this panel into the email.
Diagnostics
```
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 977, in run self._run()
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1124, in _run self.damage_curve_analysis()
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1226, in damage_curve_analysis self.earthquake_raster_population_raster()
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
[...]

Issue #3666, opened 2017-01-24
Check Reporting behaviour is correct for Earthquake


lucernae:
Reporting logic is generic, so we need to check that it is produced correctly for Earthquake.

Gustry:
# Current problem : 
![screen shot 2017-01-30 at 15 01 31](https://cloud.githubusercontent.com/assets/1609292/22425785/082b102c-e6fd-11e6-9617-43cf952a5050.png)

> Is it MMI levels instead of hazard zones?

No, we don't display the breakdown I think.
I think in 3.5, we were using low, medium and high. I need to check.
We can have it with https://github.com/inasafe/inasafe/blob/develop/safe/definitions/hazard_classifications.py#L114 so in that case, I need to work on it.

Gustry:
@Charlotte-Morgan Did you get some inputs in this issue ?

Charlotte-Morgan:
@Gustry - we did not discuss this one. I still think we need to add a new classification for the full MMI range. I am not sure where the existing classification comes from.  I would need to check it against BNPB requirements and published maps

lucernae:
I included displaced and fatalities to Analysis result, in case we want to show it.

<img width="404" alt="screen shot 2017-02-23 at 23 34 07" src="https://cloud.githubusercontent.com/assets/831865/23268525/9bf0d1e4-fa20-11e6-9c55-327f6ae5a4ef.png">

Also I prepared another section in case we want to show it:
For this one, I need mmi-related fields in analysis layer (for each mmi levels and each exposed/displaced/fatalities combinations).
<img width="402" alt="screen shot 2017-02-23 at 23 34 47" src="https://cloud.githubusercontent.com/assets/831865/23268561/b5edd632-fa20-11e6-8bc7-fe29f1501f41.png">

Charlotte-Morgan:
Fatalities should always be shown as a range. Not an exact number or even rounded.
Let me get more advice from @hadighasemi on this on Thursday. 
FYI - @RikkiWeber

lucernae:
Yes, @hadighasemi clarified some of my question just today.
CMIIW, @hadighasemi .
So, also a note for @Gustry and @timlinux 
In Earthquake, fatalities is a range of number and can be **with** a probability for that MMI range.
Hadi showed me this:

![cpebpnvwsaagbpe](https://cloud.githubusercontent.com/assets/831865/23401066/f2231016-fdd8-11e6-9b51-f7b1714c6ac7.png)

[...]

PR #3734, opened 2017-01-30
add earthquake scenario

Gustry:
### What does it fix ?
* Tickets: #3666 #3719 #3803 #3732
* Funded by : DFAT
* Description : Fix EQ IF

Known issue : breakdown exposed, not exposed, affected in the report


safe/impact_function/test/test_impact_function.py
```diff
@@ -374,6 +374,7 @@ def test_scenarii(self):
         Let's keep booleans to False by default.
         """
         scenarii = {
```
timlinux: Would still like to see this called 'scenarios' rather than 'scenarii'.

timlinux:
LGTM when tests pass....

Gustry:
@ivanbusthomi Can you try this PR about EQ raster on population raster ? 
The PR is not finished, but I will really appreciate if I can have some feedbacks already. Thanks a lot.

safe/definitions/hazard_classifications.py
```diff
@@ -183,11 +182,10 @@
             'affected': True,
             'description': tr('The medium hazard class.'),
             'string_defaults': ['medium'],
-            # Not used because EQ algs take care of this
-            # 'displacement_rate': 0.0,
+            'displacement_rate': 1.0,
             # Not used because EQ algs take care of this
             # 'fatality_rate': 0.0,
-            'numeric_default_min': 7,
+            'numeric_default_min': 6,
```
Gustry: @Charlotte-Morgan You changed these values yesterday. But I need to rollback, the minimum is not included. Please confirm this is still correct.

PR #3738, opened 2017-01-31
Move monkey patch for keyword fix #3737

### LINKED ISSUE #3737: Make Sure Release Package is Working
### Problem
Currently, we put
`from safe.test.utilities import monkey_patch_keywords` in `datastore.py`. But, we remove test directory when do the packaging since we don’t put test related code to the package.
There is `rm -rf ${WORKDIR}/${DIR}/safe/test` in the `release.sh`

### Expected Behaviour
User can install from release package.

### Environment
* InaSAFE : 
* QGIS : 
* OS : 
* Data used :
  * Exposure : 
  * Hazard : 
  * Aggregation :

### Proposed Solution
Avoid import from test.

ismailsunni:


Gustry:
Thanks and sorry for that. I will merge as I'm going to use it just now.

PR #3739, opened 2017-01-31
move check_inasafe_fields outside of tests

Gustry:


Gustry:
So easy to do this mistake when you do a copy/paste. thanks

PR #3757, opened 2017-02-01
add ±1 tolerance in the summary calculation

Gustry:


Gustry:
I cherry picked this commit in PR #3758

Issue #3741, opened 2017-01-31
Failure to run flood on buildings using IF Wizard vers 4


NyakudyaA:
### Problem
When running flood on buildings using wizard there is an error that comes.

![out](https://cloud.githubusercontent.com/assets/2510900/22456585/10a4a4a8-e79c-11e6-908f-b01e821b189b.gif)

```
An error has occurred while executing Python code: 

AttributeError: 'StepFcAnalysis' object has no attribute 'extent' 
Traceback (most recent call last):
  File "/home/admire/.qgis2/python/plugins/inasafe/safe/gui/tools/wizard/wizard_dialog.py", line 656, in on_pbnNext_released
    self.go_to_step(new_step)
  File "/home/admire/.qgis2/python/plugins/inasafe/safe/gui/tools/wizard/wizard_dialog.py", line 592, in go_to_step
    self.step_fc_analysis.setup_and_run_analysis()
  File "/home/admire/.qgis2/python/plugins/inasafe/safe/gui/tools/wizard/step_fc90_analysis.py", line 114, in setup_and_run_analysis
    self.impact_function = self.prepare_impact_function()
  File "/home/admire/.qgis2/python/plugins/inasafe/safe/gui/tools/wizard/step_fc90_analysis.py", line 188, in prepare_impact_function
    if self.extent.user_extent:
AttributeError: 'StepFcAnalysis' object has no attribute 'extent'


Python version: 2.7.12 (default, Nov 19 2016, 06:48:10) [GCC 5.4.0 20160609] 
QGIS version: 2.18.2 Las Palmas, 102ee87 

```

### Expected Behaviour
Run analysis without errors

### Environment
* InaSAFE :  develop
* QGIS : 2.18.3 and 2.14.3
* OS :  Ubuntu 16.04
* Data used : `safe/test/data`
  * Exposure :  `exposure/buildings.shp`
  * Hazard : `hazard/floods.shp`
  * Aggregation : None

### Proposed Solution
Run analysis

Gustry:
Thanks, I got this one too

ismailsunni:
I got this one also today. Strange, may be I break the code somewhere, it was working before. Ouch, we really need a better unit test.

NyakudyaA:
@Gustry , @ismailsunni  It seems most of the IF wizard is broken. Have just tested with Ash raster on building centroids and have the same error as above.

[...]

PR #3758, opened 2017-02-01
fix default analysis extent behaviour

Gustry:


safe/gui/widgets/dock.py
```diff
@@ -268,7 +269,8 @@ def read_settings(self):
             if not crs.isValid():
                 crs = None
 
-        if crs and extent:
+        mode = setting('analysis_extents_mode', HAZARD_EXPOSURE_VIEW)
```
ismailsunni: Note for me: cross reference for #3741

Issue #3106, opened 2016-08-04
Minidump error on cancelling an OSM download request


Gustry:
### Problem
QGIS will close unexpectedly if you cancel a download in the OSM Downloader

### Expected Behaviour
The request should be cancelled without closing QGIS and loosing your project.

### Environment
* InaSAFE : Develop
* QGIS : 2.14
* OS : Windows, Mac, Linux
* Data used : OSM Downloader

samnawi:
Hi @Gustry I also got the same issue when canceling the OSM download request
![image](https://cloud.githubusercontent.com/assets/8871958/18075084/031ffffe-6e9c-11e6-81b6-c7f62c449964.png)

### Environment
* InaSAFE : Develop
* QGIS : 2.14.3
* OS : Windows7 64 bit

Gustry:
@myarjunar I tried to fix this one, but I couldn't really fix this issue.

I made a PR to show you what I started. If you have an idea how to fix it, please feel free to make a PR.

myarjunar:
@Gustry cool! I'm having a look

Gustry:
I think we will drop this issue soon.
In QGIS 2.18, the next LTR in a few months, there is a `QgsFileDownloader`. It will solve this issue.

PR #3760, opened 2017-02-02
WIP : trying to fix OSM downloader minidump

Gustry:
Ticket : #3106


timlinux:
Can we merge this so we can get it out the queue?

Gustry:
This is not fixed. It was just to share my branch with @myarjunar if he could work on the issue. I couldn't fix it.

Gustry:
In 2.18, we can use the QgsFileDownloader : https://qgis.org/api/classQgsFileDownloader.html, then we won't have this bug anymore

Issue #3463, opened 2016-11-21
Setting for Keyword Assignment


ismailsunni:
### Problem
Currently, we have default value for some field (e.g. youth_ratio_field) and we have default value for it. We need a setting windows to set this value.

![inasafe_default_values_setting](https://cloud.githubusercontent.com/assets/1421861/20515377/eff5bed6-b0c3-11e6-96cf-37fdd9fbe72e.png)

This window will be put in the setting tab.

Gustry:
Add a button `Restore to defaults` to restore these defaults user values by hard-coded values.

Gustry:
And also, why not using QgsExpressionContextScope ?

ismailsunni:
> Please, add a button Restore to defaults to restore these defaults user values by hard-coded values.

Added to the wireframe. Will do it.

> And also, why not using QgsExpressionContextScope ? Just thinking about that, I will have a look and come back.

Never use that before. What's the benefit of it compared to store it in QSettings for this use case?

ismailsunni:
@myarjunar the only thing that we need to do is adding "Restore to default" button

Gustry:
For the button, think about the QButtonBox Reset button

myarjunar:
@ismailsunni @Gustry Reset or Restore Defaults button?? :smiley:

Gustry:
Ahah, sorry I meant check the QButtonBox class, there is one role for this kind of purpose. I didn't want to mean use **this** one.

Let's use a RestoreDefaults

ismailsunni:
Done in #3768

PR #3768, opened 2017-02-03
add button for restore default ratio

myarjunar:
inasafe 4 0 0b0 options_017

### DIFF
--- a/safe/gui/tools/options_dialog.py
+++ b/safe/gui/tools/options_dialog.py
@@ -128,6 +128,13 @@ class OptionsDialog(QtGui.QDialog, FORM_CLASS):
         self.custom_org_disclaimer_checkbox.toggled.connect(
             self.set_org_disclaimer)
 
+        # Set up listener for restore defaults button
+        self.restore_defaults = self.button_box_restore_defaults.button(
+            QtGui.QDialogButtonBox.RestoreDefaults)
+        self.restore_defaults.setCheckable(True)
+        self.restore_defaults.clicked.connect(
+            self.restore_defaults_ratio)
+
     def save_boolean_setting(self, key, check_box):
         """Save boolean setting according to check_box state.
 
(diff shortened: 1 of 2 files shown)


Gustry:
I would add the button in the toolbar in the bottom, no ?

myarjunar:
So, my consideration of putting the button there is because it will only restore the global default tab. That's what I'm getting from the issue. Am I right? Or it should restore the default value for the whole setting? If so, you're right, we're gonna put it at the bottom.

Gustry:
Ah yes correct, thanks !

safe/gui/tools/options_dialog.py
```diff
@@ -545,3 +557,14 @@ def save_default_values(self):
                 parameter.guid,
                 parameter.value
             )
+
+    def restore_defaults_ratio(self):
+        '''Restore InaSAFE default ratio.'''
```
ismailsunni: Triple double-quote

Gustry:
You can have a look to landscape, it's a an easy fix ;-)

ismailsunni:
Hi @myarjunar 
I tried your PR, but it seems it doesn't do what it should do. The button should reset the value when clicked with the default one. You can get the default value from https://github.com/inasafe/inasafe/blob/develop/safe/definitions/default_values.py#L16

myarjunar:
So the previous one will only restore to defaults on the moment the user edit the value then directly click the button, it is not working when the user edit the value, then click ok, then reopen it, then click the the restore defaults button. Hope current commit will fix.

ismailsunni:
Hi @myarjunar 
I checked and it looks good and has proper behaviour. But I saw you recreate the widget in your code. Can you make it only to update the value in the combobox ?

timlinux:
See #3463

ismailsunni:
I think my we can merge this since it has fixed the problem. LGTM. We can visit again later for better implementation.

PR #3769, opened 2017-02-03
Fix #3135. Only check safe package and always run update.

### LINKED ISSUE #3135: Check if we need to update make update-translation-strings for safe package only
### Problem

Currently the script will check to all python file. Meanwhile, we should only care / translate to the strings under safe directory only.
### Expected Behaviour

Only get strings from safe directory.
### Proposed Solution

Check `scripts/update-strings.sh`

Notes: this is just my note. I will check tomorrow again.

ismailsunni:


Gustry:
MBGTM : Might Be Good To Me

PR #3773, opened 2017-02-06
Switch help description

ismailsunni:
Fix missing commits.


ismailsunni:
I will just merge this since I only fix needed error (wrong hash for submodule)

Issue #3770, opened 2017-02-03
Better definitions Function


ismailsunni:
### Problem
Currently we have this `definition` function to get the full definition from a key. So far so good, until we have two definition which have the same key. It was unit `generic` in units.py and `generic` in hazards.py.

It's not good since we can get the wrong definition.

### Expected Behaviour
We should have a better mechanism to get the full definitions

### Proposed Solution
One of the solution is updating the method to have new parameter, that wil give the function a file to check the definitions. For example
```
def definition(definition_file, keyword):
    pass
```
Then we will need to list all of the file, and put them as a variable, for example:
```
UNITS_FILE = 'units.py'
HAZARDS_FILE = 'hazards.py'
```
By doing this, we can do the refactoring easily in the future

cc @Gustry @timlinux @lucernae @myarjunar

Gustry:
Thanks, it was scary for me this `definition` function ;-)

timlinux:
Yeah this is why I suggested before that we should use a naming scheme that makes all keys unique. Also that we should have a unit tests that fails if we find two keys the same. I prefer this approach to having a second parameters as we will still have hell if we move definitions between files.

I can help to come up with a better naming scheme for keys....

ismailsunni:
> Yeah this is why I suggested before that we should use a naming scheme that makes all keys unique. Also that we should have a unit tests that fails if we find two keys the same. I prefer this approach to having a second parameters as we will still have hell if we move definitions between files.

Ah yes, I didn't thought that there is a case to move definitions....

I prefer to use a naming scheme now

ismailsunni:
In the meeting we decided to review all the keys and make sure to avoid conflict. We will improve it later in the next version.

PR #3774, opened 2017-02-06
Fix 3619 missing constant

### LINKED ISSUE #3619: Problem in KW wizard for flood hazard definitions:  Missing constant
### Problem

I get this when defining the keywords for a flood layer in my postgres database:

screen shot 2017-01-17 at 3 27 37 pm

I don't think the issue is specific to the fact that the layer is in postgres.

### Expected Behaviour

A more sensible message should be shown to the user in the wizard step description

### Environment
* InaSAFE : 4.beta
* QGIS : 2.14.10
* OS : MacOS
* Data used : 
  * Exposure : n/a
  * Hazard : sri lanka flood data
  * Aggregation : n/a

### Proposed Solution

Please check where this message is going wrong and fix

ismailsunni:
Fix #3619 
Tester for #3770

### DIFF
--- a/safe/gui/tools/wizard/wizard_strings.py
+++ b/safe/gui/tools/wizard/wizard_strings.py
@@ -74,38 +74,31 @@ allow_resampling_question = tr(
     'raster to the hazard layer resolution during analyses. Please select the '
     'check box below if you want to set the <i>allow_resampling</i> '
     'keyword to <i>False</i>.</p>')  # (subcategory, layer purpose, layer_mode)
-flood_metres_depth_question = tr(
-    'flood depth in metres')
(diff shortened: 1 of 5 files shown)


safe/gui/tools/wizard/wizard_strings.py
```diff
@@ -74,38 +74,31 @@
     'raster to the hazard layer resolution during analyses. Please select the '
```
Gustry: I think this file would be better in definitions. But you don't have to change it.

ismailsunni: Agree

PR #3775, opened 2017-02-06
update strings

Gustry:


timlinux:
LGTM - a few small corrections if you don't mind applying them...

Gustry:
You forgot to send your comments

safe/utilities/keyword_io.py
```diff
@@ -231,16 +232,17 @@ def copy_keywords(
...
-        except Exception, e:
+        except Exception as e:
             message = self.tr(
                 'Failed to copy keywords file from : \n%s\nto\n%s: %s' % (
                     source_layer.source(), new_destination, str(e)))
             raise Exception(message)
         return
 
-# methods below here should be considered private
+    # methods below here should be considered private
```
ismailsunni: +1

Gustry: We are already doing that. I think this one is legacy and I changed the indent only

Gustry:
We need to be careful with this submodule. Thanks, fixed

Issue #3779, opened 2017-02-07
Tsunami on Population: displacement rates


Charlotte-Morgan:
Following discussion with Tsunami hazard experts; please assign the following affected status and displacement rates for tsunami hazard classes in #2907

The same rates apply regardless of the 3 or 5 class classification

![image](https://cloud.githubusercontent.com/assets/5743135/22678127/df395912-ed4b-11e6-932f-1c3565fa83db.png)

Gustry:
I did it in the PR above @Charlotte-Morgan

PR #3782, opened 2017-02-07
update displacement rates for tsunami classification

Gustry:
See ticket #3779 from @Charlotte-Morgan

### DIFF
--- a/safe/definitions/hazard_classifications.py
+++ b/safe/definitions/hazard_classifications.py
@@ -489,7 +489,7 @@ tsunami_hazard_classes = {
             'affected': True,
             'description': tr('Water above 8.0m.'),
             'string_defaults': [],
-            'displacement_rate': 0.0,
+            'displacement_rate': 0.01,
             'fatality_rate': 0.0,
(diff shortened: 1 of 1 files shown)


timlinux:
@charlotte-morgan it would be great if we could add a citation for this even if it is something like 'Personal communication: GA Tsunami Expert, February 2017. '

Charlotte-Morgan:
hi @Gustry - apologies I missed this review request.  As you see I have also made a PR that covers the tsunami displacement rates.  In my discussion with @RikkiWeber we decided to apply a rate of 1.0 to all people in the medium, high and very high classes. ie these are people in experiencing a tsunami wave of 0.7 m or more. Certainly we should have 1.0 or 100% displaced for people in high >3m or very high >8m.

I also see you have tweaked the thresholds; this is only relevant to exposure = population.
Do you have a neater way to handle default classes for different exposures?

Gustry:
Ok, so I will close this PR.

> Do you have a neater way to handle default classes for different exposures?

For now, make a new classification specific to population.
We can come back to refactor this later.

Issue #3783, opened 2017-02-07
TypeError when running volcanic hazard on land cover


ivanbusthomi:
### Problem
When I tried running volcanic hazard on land cover, I got following error
```
Problem
-------------
The following problem(s) were encountered whilst running the analysis.
TypeError

Details
-------------
These additional details were reported when the problem occurred.
QgsFeature.setGeometry(QgsGeometry): argument 1 has unexpected type 'NoneType'

Diagnostics
-------------
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 981, in run self._run()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1130, in _run self.gis_overlay_analysis()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1224, in gis_overlay_analysis self.intersect_exposure_and_aggregate_hazard()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1526, in intersect_exposure_and_aggregate_hazard self._exposure = clean_layer(self.exposure)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\gis\vector\clean_geometry.py", line 38, in clean_layer feature.setGeometry(geometry_checker(geom))
```
### Expected Behaviour
It should run without error

### Environment
* InaSAFE : 4.0 develop version
* QGIS : 2.14.10
* OS : Windows
* Data used : Nagekeo dataset
  * Exposure : Landuse (NGK_Landuse_250K_WGS84.shp)
[...]

PR #3784, opened 2017-02-07
remove features if the geometry is None

Gustry:
Issue #3783 from @ivanbusthomi


ivanbusthomi:
@Gustry LGTM, the error is gone,thanks

Issue #3780, opened 2017-02-07
InaSAFE help enhancement


Charlotte-Morgan:
It would be great to be able to minimise the help window or click outside it to interact with the main window without having to close the help each time

Gustry:
For now @Charlotte-Morgan, you can right click on the page and choose `Open in webbrowser`.

@myarjunar Can you have a look to this issue ?

ivanbusthomi:
I also noticed that everytime I open help, it creates new temporary file instead of calling already created one. Is it intended?

![image](https://cloud.githubusercontent.com/assets/7351977/22686603/4a27363a-ed57-11e6-8b09-73db6cd1fc79.png)

Gustry:
Yes, I guess this is not a problem. And it's easier for us to create a new one.

Gustry:
We can also on the right click menu save it as a PDF.

timlinux:
It only creates that file if you use the open in browser context menu. Its not a problem if it does that, your OS should clean that away again....

PR #3796, opened 2017-02-07
Make help non-modal and minimazable / maximizable

timlinux:
Note that the non-modal behavior was already there for me. Added minimize/maximize buttons. See #3780

screencapture

### DIFF
--- a/safe/gui/tools/help_dialog.py
+++ b/safe/gui/tools/help_dialog.py
@@ -5,12 +5,12 @@
 # noinspection PyUnresolvedReferences
 import qgis  # pylint: disable=unused-import
 from PyQt4 import QtGui
+from PyQt4.QtCore import Qt
 
 from safe.common.version import get_version
 from safe.utilities.resources import (
     get_ui_class, html_footer, html_help_header)
 from safe.gui.tools.help.dock_help import dock_help
-from safe.gui.widgets import message_viewer
 
 FORM_CLASS = get_ui_class('help_dialog_base.ui')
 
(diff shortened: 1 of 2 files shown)


timlinux:
@myarjunar No need to do anything on #3780 from your side ok?

safe/gui/tools/help_dialog.py
```diff
@@ -33,7 +33,9 @@ def __init__(self, parent=None, message=None):
         :type parent: QWidget
         """
 
-        QtGui.QDialog.__init__(self, parent)
+        QtGui.QDialog.__init__(
+            self, parent,
+            flags=Qt.WindowMinimizeButtonHint | Qt.WindowMaximizeButtonHint)
         self.setupUi(self)
         self.setWindowTitle(self.tr('InaSAFE %s Help' % get_version()))
```
Gustry: Your title is overwrite here @timlinux I think you want to change it.

timlinux: @Gustry yeah I saw that too - I'm happy to leave the nicer title in the ui file anyway....better than 'Dialog'

Gustry: Yes better than `Dialog` but I though you wanted to rename your dialog by removing the inasafe version, like we did in the minimum needs calculator

Gustry:
😱You are using windows :D

safe/gui/tools/help_dialog.py
timlinux: Ah - yes maybe that will be a good idea....

timlinux:
Call me @timwindows :-) I think it is good to test a bit with what the masses have....

safe/gui/tools/help_dialog.py
ismailsunni: I think we should have convention on how we set title or text in the dialog. I prefer to do anything beside the static layout / widget in python code. Set text is in python code for me.

timlinux: @ismailsunni I could not really understand your comment :-P I'm going to go with defining it in the UI for now ok? It should still show up in translation list.

Gustry:
Looks good to me, thanks !

Issue #3789, opened 2017-02-07
Wizard : multi hazard classification


Gustry:
### Problem

* [x] 1 : What does `No classification` mean ? We should not be able to launch an IF according to the exposure in the dock and wizard if the user didn't set this exposure in the hazard keywords ?
For now, it will raise an exception if you launch an IF with an exposure which has not been defined.
![screen shot 2017-02-07 at 10 43 09](https://cloud.githubusercontent.com/assets/1609292/22685861/3f87855c-ed22-11e6-843f-58bdf51adaff.png)

* [x] 2 : Should we be able to click `Next` if there is no hazard classification ? (screenshot above). We will get an empty value_map : 
![screen shot 2017-02-07 at 10 20 51](https://cloud.githubusercontent.com/assets/1609292/22685931/8ce5865a-ed22-11e6-8055-3402aa2bb004.png)

* [x] 3 : Same question when we start editing a value map : 
![screen shot 2017-02-07 at 10 18 01](https://cloud.githubusercontent.com/assets/1609292/22686035/ef98b894-ed22-11e6-9c57-e240a7942ac8.png)

* [x] 3 : Bug if we press `save` and `edit` again. It will not display the value map that we did just before.
![value_map](https://cloud.githubusercontent.com/assets/1609292/22685958/a9a0ff9a-ed22-11e6-98bf-6bff868e1a19.gif)

* [x] 4 : What is this small `1` ?
![screen shot 2017-02-07 at 10 15 43 copy](https://cloud.githubusercontent.com/assets/1609292/22686252/a653f800-ed23-11e6-87c0-c6fae0b2c874.png)

* [x] 4 : What is the button `Load default`, I tried, nothing happened on my laptop.
![screen shot 2017-02-07 at 10 15 43](https://cloud.githubusercontent.com/assets/1609292/22686134/3c440464-ed23-11e6-9299-692e1bdac229.png)

* [x] 5 : I don't think it's a good idea to display definitions. The user is thinking where does this come from. I think it's confusing to display Kawasan Rawan Bencana, especially when I use InaSAFE outside of Indonesia. And these values are not in my attribute table. 
![screen shot 2017-02-07 at 10 56 19](https://cloud.githubusercontent.com/assets/1609292/22686377/3fa63e8c-ed24-11e6-9a2a-62e9f4ffebcb.png)

[...]

PR #3800, opened 2017-02-08
check if the hazard can run on the exposure

Gustry:
Ticket #3789 

dock

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -117,6 +117,7 @@ from safe.utilities.gis import is_vector_layer, is_raster_layer
 from safe.utilities.i18n import tr
 from safe.utilities.unicode import get_string
 from safe.utilities.keyword_io import KeywordIO
+from safe.utilities.metadata import active_thresholds_value_maps
 from safe.utilities.utilities import (
     replace_accentuated_characters, get_error_message)
 from safe.utilities.profiling import (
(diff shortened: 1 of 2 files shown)


safe/impact_function/impact_function.py
```diff
@@ -800,6 +801,30 @@ def prepare(self):
...
                     return PREPARE_FAILED_BAD_INPUT, message
 
+            # We need to check if the hazard is OK to run on the exposure.
+            hazard_keywords = self.hazard.keywords
+            exposure_key = self.exposure.keywords['exposure']
+            if not active_thresholds_value_maps(hazard_keywords, exposure_key):
+                warning_heading = m.Heading(
+                    tr('Incompatible exposure/hazard'), **WARNING_STYLE)
+                warning_message = tr(
+                    'The hazard layer is not set up for this kind of exposure.'
```
timlinux: Need space after the full stop.

safe/impact_function/impact_function.py
```diff
@@ -800,6 +801,30 @@ def prepare(self):
...
+                    'In InaSAFE, you need to enable the set up the hazard '
```
timlinux: In InaSAFE, you need to define keywords in the hazard layer for each exposure type that you want to use with the hazard.

safe/impact_function/impact_function.py
```diff
@@ -800,6 +801,30 @@ def prepare(self):
...
+                    'classification for each kind of exposure you want to '
+                    'use.')
+                suggestion_heading = m.Heading(
+                    tr('Suggestion'), **SUGGESTION_STYLE)
+                suggestion = tr(
+                    'Please run the keyword wizard on the hazard layer to add '
```
timlinux: "Please run the select the hazard layer in the legend and then run the keyword wizard to define the needed keywords for this exposure."

It would be nice to also replace 'this exposure' with a string interpolator so that it reads e.g. "for **population** exposure".

Issue #3791, opened 2017-02-07
Hazard thresholds information in dock is not properly styled.


ivanbusthomi:
### Problem
I noticed that after we implemented the new thresholds, the dock shows it without a style. I think it is a little bit hard for the user to understand.
![image](https://cloud.githubusercontent.com/assets/7351977/22687045/13576efc-ed59-11e6-8767-fdae59896f95.png)

### Expected Behaviour
The dock should show the thresholds nicely :smile: 

### Environment
* InaSAFE : 4.0 develop
* QGIS : 2.14.10
* OS : Windows
* Data used :
  * Exposure : 
  * Hazard : 
  * Aggregation :

### Proposed Solution
The one in the keywords creation wizard is great. I think it would be great if it also implemented in the dock
![image](https://cloud.githubusercontent.com/assets/7351977/22687204/a3bdd846-ed59-11e6-80d4-a58c8860f261.png)
___
CC: @ismailsunni @Gustry  @timlinux @samnawi @felix-yew @Charlotte-Morgan @fredychandra

Gustry:
Thanks, I forgot to put this one in my big ticket.
@ismailsunni is already aware about this issue.

timlinux:
Eeesh yeah that does not look nice. @Gustry can you repurpose my logic for that from definitions_help.py?

timlinux:
Ah sorry @ismailsunni above comment should apply to you since you are assigned.

ismailsunni:
See also #3373

ivanbusthomi:
I think we can close this since the dock is looking really nice when showing the hazard classifications.

ismailsunni:
Thanks @ivanbusthomi

Issue #3373, opened 2016-10-31
We need to fix dock keywords display in v4.0


timlinux:
![screen shot 2016-10-31 at 9 42 40 am](https://cloud.githubusercontent.com/assets/178003/19842747/7239db16-9f4e-11e6-82ec-da41b1c3e438.png)

Gustry:
I should also clean keywords along the workflow.
Some keywords are missing and some keywords are not relevant at the end of the process.

ismailsunni:
I will review this while fixing my new keywords.

Gustry:
If you want to have a look to this ticket : https://github.com/inasafe/inasafe/issues/2800

PR #3801, opened 2017-02-08
Fix 3791 keyword in dock

ismailsunni:
Including #3791 #3373 
Sample
screen shot 2017-02-08 at 21 07 33
screen shot 2017-02-08 at 21 07 42

### DIFF
--- a/safe/utilities/keyword_io.py
+++ b/safe/utilities/keyword_io.py
@@ -107,7 +107,8 @@ class KeywordIO(QObject):
 
         return dictionary[keyword]
 
-    def read_keywords(self, layer, keyword=None):
+    @staticmethod
+    def read_keywords(layer, keyword=None):
         """Read keywords for a datasource and return them as a dictionary.
 
         This is a wrapper method that will 'do the right thing' to fetch
@@ -140,7 +141,8 @@ class KeywordIO(QObject):
         # Try to read from ISO metadata first.
         return read_iso19115_metadata(source, keyword)
 
-    def write_keywords(self, layer, keywords):
+    @staticmethod
+    def write_keywords(layer, keywords):
         """Write keywords for a datasource.
 
         This is a wrapper method that will 'do the right thing' to store
@@ -309,6 +311,7 @@ class KeywordIO(QObject):
             'continuous_hazard_unit',
             'value_map',  # attribute values
             'thresholds',  # attribute values
+            'value_maps',  # attribute values
             'inasafe_fields',
             'inasafe_default_values',
             'resample',
@@ -420,6 +423,9 @@ class KeywordIO(QObject):
                 'inasafe_fields',
                 'inasafe_default_values']:
             value = self._dict_to_row(value)
+        elif keyword == 'value_maps':
+            LOGGER.debug('Showing value maps')
+            value = self._value_maps_row(value)
         elif keyword == 'thresholds':
             value = self._threshold_to_row(value)
         # In these KEYWORD cases we show the DESCRIPTION for
@@ -457,7 +463,7 @@ class KeywordIO(QObject):
         return row
 
     @staticmethod
-    def _threshold_to_row(thresholds_keywords):
+    def _threshold_to_row(thresholds_keyword):
         """Helper to make a message row from a threshold
 
         We are expecting something like this:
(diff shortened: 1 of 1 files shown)


Gustry:
Why is-it always `Class name : Values` in your screenshot ? Is-it on purpose ?

ismailsunni:
@gustry, yes it does. Class name and values is the header table.

Gustry:
Ahah sorry. It's not obvious that it's a header. It's mixed between two properties : the name of the classification and the the value mapping, so we think it's also a property

safe/utilities/keyword_io.py
```diff
@@ -457,35 +463,127 @@ def _keyword_to_row(self, keyword, value, wrap_slash=False):
...
         :rtype: safe.messaging.items.table
         """
-        if isinstance(keyword_value, basestring):
-            keyword_value = literal_eval(keyword_value)
+        if isinstance(thresholds_keyword, basestring):
+            thresholds_keyword = literal_eval(thresholds_keyword)
+
+        table = m.Table(style_class='table table-condensed table-striped')
+
+        i = 0
```
Gustry: Ahah, it's a long time in python I didn't see that. I usually use `enumerate` ;-)
No worries. We can keep it, I guess it will not fit one your line.

safe/utilities/keyword_io.py
```diff
@@ -457,35 +463,127 @@ def _keyword_to_row(self, keyword, value, wrap_slash=False):
...
+        for exposure_key, classifications in thresholds_keyword.items():
+            i += 1
+            exposure = definition(exposure_key)
+            exposure_row = m.Row()
+            exposure_row.add(m.Cell(m.ImportantText('Exposure')))
+            exposure_row.add(m.Cell(m.Text(exposure['name'])))
+            exposure_row.add(m.Cell(''))
+            table.add(exposure_row)
+
+            active_classification = None
```
Gustry: Sorry I didn't catch where you can set a classification is active or not ?

safe/utilities/keyword_io.py
```diff
@@ -457,35 +463,127 @@ def _keyword_to_row(self, keyword, value, wrap_slash=False):
...
+                if value.get('active'):
+                    active_classification = definition(classification)
+                    classification_row.add(
+                        m.Cell(active_classification['name']))
+                    classification_row.add(m.Cell(tr('')))
+                    break
+
+            if not active_classification:
+                classification_row.add(m.Cell(tr('No classifications set.')))
+                classification_row.add(m.Cell(tr('')))
```
Gustry: Why do you want to translate an empty string ? One line less in transifex ;-)

timlinux: I think @ismailsunni is cut and pasting my code probably - I think I used tr('') when I had some unicode issues with the generated html....

safe/utilities/keyword_io.py
timlinux: The i is used further down to keep track of the row number....

safe/utilities/keyword_io.py
```diff
@@ -535,3 +633,115 @@ def _dict_to_row(self, keyword_value):
 
             table.add(row)
         return table
+
+    @staticmethod
+    def _value_maps_row(value_maps_keyword):
+        """Helper to make a message row from a value maps.
```
timlinux: maps -> map

safe/utilities/keyword_io.py
ismailsunni: Yes, like Tim said.

safe/utilities/keyword_io.py
ismailsunni: In the keyword wizard. With the combo box.

safe/utilities/keyword_io.py
ismailsunni: Ahaha, no @timlinux 
I think I copied and pasted the code above it and only deleted the string. I will remove the `tr`.
Thanks

safe/utilities/keyword_io.py
ismailsunni: Well, there are several value map, so I use plural for it.

ismailsunni:
> Just a question, can we still display the old value_map and thresholds ? In keywords at the end of the IF, we don't have this complex dictionary. It will be useful if we run the IF in debug mode to be able to see which classification has been used in the dock.

@Gustry I have added handler in the new commit for "old" thresholds format. The value_map should work since it has different key (value_map vs value_maps)

PR #3807, opened 2017-02-09
cyclone fixes

mbernasocchi:
added metres per second unit

### DIFF
--- a/safe/definitions/units.py
+++ b/safe/definitions/units.py
@@ -91,7 +91,7 @@ unit_metres_per_second = {
     'plural_name': tr('m/s'),
     'abbreviation': tr('m/s'),
     'description': tr(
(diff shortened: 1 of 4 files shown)


mbernasocchi:
@wcarthur can you have a look at this please?

safe/definitions/units.py
```diff
@@ -85,6 +85,21 @@
         }
     ]
 }
+unit_metres_per_second = {
+    'key': 'metres_per_second',
+    'name': tr('m/s'),
+    'plural_name': tr('m/s'),
+    'abbreviation': tr('m/s'),
+    'description': tr(
+        '<b>The metre per second</b> is a unit of speed, expressing the '
```
timlinux: <b>Metres per second</b>

Issue #3214, opened 2016-09-16
We should move the metadata.db and minimum needs into .qgis2/inasafe


Gustry:
### Problem

The metadata sqlite DB is stored in `~/.inasafe/metadata.db`
The minimum needs are stored in `~/.qgis2/minimum_needs/`
### Expected Behaviour

InaSAFE should store all these files under : 
`.qgis2/inasafe/`
using
`QgsApplication.qgisSettingsDirPath()`

@ismailsunni Do you think that 4.0 will break (or partially break) all previous keywords (3.5) ?

ismailsunni:
@Gustry I am still not sure. I will introduce some new keywords. The old keywords will not work in some of the new post processor architecture since we run the post processor based on this new introduced keywords.
I will try my best to have a migration seamless. With some extra functions

Gustry:
For the metadata file, see PR https://github.com/inasafe/inasafe/pull/3431
For the minimum needs, we will wait for the minimum needs refactoring

myarjunar:
closed in #3810

PR #3810, opened 2017-02-10
change min needs profile path to ~/.qgis2/inasafe/minimum_needs

myarjunar:
see #3214

### DIFF
--- a/safe/gui/tools/minimum_needs/needs_profile.py
+++ b/safe/gui/tools/minimum_needs/needs_profile.py
@@ -288,15 +288,19 @@ class NeedsProfile(MinimumNeeds):
         :rtype: QString
         """
         if self._root_directory is None or self._root_directory == '':
+            # Map the root directory to ~/.qgis2/inasafe so the minimum needs
+            # profile will be placed there (~/.qgis2/inasafe/minimum_needs).
             try:
                 # noinspection PyArgumentList
-                self._root_directory = QgsApplication.qgisSettingsDirPath()
+                self._root_directory = os.path.join(
+                    QgsApplication.qgisSettingsDirPath(),
+                    'inasafe')
             except NameError:
                 # This only happens when running only one test on its own
                 self._root_directory = None
             if self._root_directory is None or self._root_directory == '':
                 self._root_directory = os.path.join(
-                    os.path.expanduser('~'), '.qgis2')
+                    os.path.expanduser('~'), '.qgis2', 'inasafe')
         return self._root_directory
 
     @staticmethod
--- a/safe/gui/tools/minimum_needs/needs_profile.py
+++ b/safe/gui/tools/minimum_needs/needs_profile.py
@@ -198,6 +198,7 @@ class NeedsProfile(MinimumNeeds):
                 locale_minimum_needs_dir, file_name)
             if not os.path.exists(destination_file) or overwrite:
                 copy(source_file, destination_file)
+        self.load_old_profile(locale_minimum_needs_dir)
         profiles = [
             profile[:-5] for profile in
             os.listdir(locale_minimum_needs_dir) if
@@ -338,3 +339,20 @@ class NeedsProfile(MinimumNeeds):
             os.path.join(
                 str(self.root_directory), 'minimum_needs', profile + '.json')
         )
+
+    def load_old_profile(self, locale_minimum_needs_dir):
+        """Load old minimum needs profile under ~/.qgis2/minimum_needs.
+
+        :param locale_minimum_needs_dir: User local minimum needs profile path.
+        :type locale_minimum_needs_dir: str
+        """
+        old_profile_path = os.path.join(
+            QgsApplication.qgisSettingsDirPath(), 'minimum_needs')
+
+        if os.path.exists(old_profile_path):
+            for filename in os.listdir(old_profile_path):
+                source_file = os.path.join(old_profile_path, filename)
+                destination_file = os.path.join(
+                    locale_minimum_needs_dir, filename)
+                if not os.path.exists(destination_file):
+                    copy(source_file, destination_file)
--- a/safe/gui/tools/minimum_needs/needs_profile.py
+++ b/safe/gui/tools/minimum_needs/needs_profile.py
@@ -283,25 +283,17 @@ class NeedsProfile(MinimumNeeds):
 
     @property
     def root_directory(self):
-        """Get the home root directory
+        """Map the root directory to ~/.qgis2/inasafe so the minimum needs
+           profile will be placed there (~/.qgis2/inasafe/minimum_needs).
 
         :returns: root directory
         :rtype: QString
         """
         if self._root_directory is None or self._root_directory == '':
-            # Map the root directory to ~/.qgis2/inasafe so the minimum needs
-            # profile will be placed there (~/.qgis2/inasafe/minimum_needs).
-            try:
-                # noinspection PyArgumentList
-                self._root_directory = os.path.join(
-                    QgsApplication.qgisSettingsDirPath(),
-                    'inasafe')
-            except NameError:
-                # This only happens when running only one test on its own
-                self._root_directory = None
-            if self._root_directory is None or self._root_directory == '':
-                self._root_directory = os.path.join(
-                    os.path.expanduser('~'), '.qgis2', 'inasafe')
+            # noinspection PyArgumentList
+            self._root_directory = os.path.join(
+                QgsApplication.qgisSettingsDirPath(),
+                'inasafe')
         return self._root_directory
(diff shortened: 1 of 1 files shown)


safe/gui/tools/minimum_needs/needs_profile.py
```diff
@@ -288,15 +288,19 @@ def root_directory(self):
...
+                self._root_directory = os.path.join(
+                    QgsApplication.qgisSettingsDirPath(),
+                    'inasafe')
             except NameError:
                 # This only happens when running only one test on its own
                 self._root_directory = None
             if self._root_directory is None or self._root_directory == '':
                 self._root_directory = os.path.join(
-                    os.path.expanduser('~'), '.qgis2')
+                    os.path.expanduser('~'), '.qgis2', 'inasafe')
```
Gustry: Please update this code to use `QgsApplication.qgisSettingsDirPath()`. QGIS 3 will be release soon and `.qgis2` will be outdated.

ismailsunni: He has already used it. This code is error handling when root directory is None or empty.

ismailsunni:
@Gustry yes, it will be a problem. I don't know if many people has their custom minimum needs. It will be still there. May be we just need to ask them to copy to the new folder.

Gustry:
We can put a little warning or we do it for them.

safe/gui/tools/minimum_needs/needs_profile.py
ismailsunni: ah sorry. What I meant with root directory is `self._root_directory`. It set from `QgsApplication.qgisSettingsDirPath()` first. If `self._root_directory` is empty, it will manually set to `.qgis2/inasafe`.

ismailsunni:
Either way is fine for me. Can you add it @myarjunar?

myarjunar:
@Gustry @ismailsunni ok cool, so which option will we use? We put a little warning or we do it for them?

safe/gui/tools/minimum_needs/needs_profile.py
Gustry: Well if QgsApplication.qgisSettingsDirPath(), it's a QGIS bug ! And I guess it will not happen and we should trust this function, no ?

Gustry:
@myarjunar I think it would be good to check if the old folder exists. If yes, move it first in the new directory.

And then you can always return the new directory.

Does-it work for you ?

safe/gui/tools/minimum_needs/needs_profile.py
ismailsunni: Yes, it will be a QGIS bug.
I don't mind to put the code there. Perhaps, someone from the histroy got the bug then he put the code.

+0 from me for remove or not to remove it :)

Gustry: I guess it's mix of legacy code. I'm pretty sure `QgsApplication.qgisSettingsDirPath()` was not there before, a few months ago in this function.
+1 to remove it 😋

safe/gui/tools/minimum_needs/needs_profile.py
```diff
@@ -334,3 +331,20 @@ def remove_profile(self, profile):
             os.path.join(
                 str(self.root_directory), 'minimum_needs', profile + '.json')
         )
+
+    def load_old_profile(self, locale_minimum_needs_dir):
+        """Load old minimum needs profile under ~/.qgis2/minimum_needs.
+
```
Gustry: Can you add a comment that this is for legacy things for InaSAFE < 4.0.

safe/gui/tools/minimum_needs/needs_profile.py
```diff
@@ -334,3 +331,20 @@ def remove_profile(self, profile):
...
+        old_profile_path = os.path.join(
+            QgsApplication.qgisSettingsDirPath(), 'minimum_needs')
+
+        if os.path.exists(old_profile_path):
+            for filename in os.listdir(old_profile_path):
+                source_file = os.path.join(old_profile_path, filename)
+                destination_file = os.path.join(
+                    locale_minimum_needs_dir, filename)
+                if not os.path.exists(destination_file):
+                    copy(source_file, destination_file)
```
Gustry: Can you add a check if the destination exists, then we should remove the file and the folder ?
The idea is to get ride off this old folder and have everything centralized in .qgis2/inasafe.

safe/gui/tools/minimum_needs/needs_profile.py
Gustry: Can you add a comment that this is for legacy things for InaSAFE < 4.0.

safe/gui/tools/minimum_needs/needs_profile.py
```diff
@@ -334,3 +331,20 @@ def remove_profile(self, profile):
             os.path.join(
                 str(self.root_directory), 'minimum_needs', profile + '.json')
         )
+
+    def load_old_profile(self, locale_minimum_needs_dir):
```
Gustry: Can you rename it to 'move_old_profile'.
The idea is not to load it, but to move the folder.

safe/gui/tools/minimum_needs/needs_profile.py
```diff
@@ -282,21 +283,17 @@ def provenance(self):
...
     @property
     def root_directory(self):
-        """Get the home root directory
+        """Map the root directory to ~/.qgis2/inasafe so the minimum needs
+           profile will be placed there (~/.qgis2/inasafe/minimum_needs).
 
         :returns: root directory
         :rtype: QString
         """
         if self._root_directory is None or self._root_directory == '':
```
Gustry: Is-it not possible to put load_old_profile here ?

Gustry:
Thanks @myarjunar 
The idea is to move the folder from the old place to the new one only and only if the old one exists.

We don't to manage two folders, we want user to migrate.

safe/gui/tools/minimum_needs/needs_profile.py
myarjunar: yes it is, but I don't know, I prefer to put it in the group of code which has similar functionality haha.

safe/gui/tools/minimum_needs/needs_profile.py
myarjunar: did you mean the `source_file` (the old profile)?

Gustry: No, if the file has been copied (ie if the file destination exists), we can safely remove it.
When the folder is empty, when you can remove it at the end.
So the user is migrating from .qgis2/minimums_needs to .qgis2/inasafe. In a few months, we will remove this code because we will assume that our users  doesn't use this old folder anymore

myarjunar: Oh for that problem, that's why I put this function here, https://github.com/inasafe/inasafe/pull/3810/files#diff-1a84cd819178b6f6317dd094fdf4448dR201, where the destination_file already created before this function is being called.

Gustry: Sorry I didn't understand what you mean ;-) 

Something we could do was something like : 
'''
def minimum_needs_path():
   """Return the path where minimum needs are stored"""
    path = 'new folder path'
    old_folder = 'old folder path'
    if old_folder exists:
        We copy from old to the new folder
        If the copy is fine, we remove old folder
    
   return path #We always return the new folder, as they should have been migrated
'''

myarjunar: ah.. that's what I want to make sure in my first comment. 
> Can you add a check if the destination exists, then we should remove the file and the folder?

I thought you were telling me that we should remove the destination.
a lot of misunderstanding here haha.

Gustry: Yes check if the destination exists and then remove the old folder

Gustry:
Looks good to me, thanks !

safe/gui/tools/minimum_needs/needs_profile.py
ismailsunni: Btw, if you run unit test, e.g. test_keyword_wizard, QgsApplication.qgisSettingsDirPath() will give you '' (empty string). Then, it will copy old minimum needs to the current directory of the test.

myarjunar: I also found it today :smile:  maybe that's why someone wrote this code. Should we put this code back?

ismailsunni: I think, just skip copying it if QgsApplication.qgisSettingsDirPath() is empty string.

Issue #3809, opened 2017-02-09
We should disable radio buttons if a field is selected.


Gustry:
### Problem
![screen shot 2017-02-09 at 17 08 51](https://cloud.githubusercontent.com/assets/1609292/22791710/9c1de90e-eeea-11e6-9969-81670d252927.png)

It's confusing to have radio buttons enabled if we set a field.

### Expected
Correct me if I am wrong : 
* If a field is selected (different than 'No Field') -> Disable every radio buttons, remove the checked radio button, even if it's disabled.
* If 'No field' is selected, enable every radios buttons, set 'Do not use' by default

myarjunar:
sorry, which dialog is this? @Gustry @ismailsunni

ismailsunni:
@myarjunar it is keyword wizard, in InaSAFE Default Step.

@Gustry yes, I agree with this behaviour, same one with I discussed with @Charlotte-Morgan and I put in the #3579 

Do you want to work on it @myarjunar ?

Gustry:
We can close this one. Thanks @myarjunar !

myarjunar:
closed in #3813

PR #3813, opened 2017-02-10
fix radio button behaviour on inasafe default step

myarjunar:
see #3809

### DIFF
--- a/safe/common/parameters/default_select_parameter_widget.py
+++ b/safe/common/parameters/default_select_parameter_widget.py
@@ -144,8 +144,6 @@ class DefaultSelectParameterWidget(SelectParameterWidget):
             self.custom_value.setValue(default)
 
         self.toggle_custom_value()
-        # Set selected radio button to 'Do not use'
-        self.set_default_radio_button()
 
     def toggle_custom_value(self):
         radio_button_checked_id = self.default_input_button_group.checkedId()
@@ -172,10 +170,10 @@ class DefaultSelectParameterWidget(SelectParameterWidget):
         else:
             for button in self.default_input_button_group.buttons():
                 button.setEnabled(True)
-            self.set_default_radio_button()
+            self.set_selected_radio_button()
             self.custom_value.setEnabled(True)
 
-    def set_default_radio_button(self):
+    def set_selected_radio_button(self):
         """Set selected radio button to 'Do not use'."""
         dont_use_button = self.default_input_button_group.button(
             len(self._parameter.default_values) - 2)
--- a/safe/common/parameters/default_select_parameter_widget.py
+++ b/safe/common/parameters/default_select_parameter_widget.py
@@ -79,6 +79,7 @@ class DefaultSelectParameterWidget(SelectParameterWidget):
 
         # Connect
         # noinspection PyUnresolvedReferences
+        self.input.currentIndexChanged.connect(self.toggle_input)
         self.default_input_button_group.buttonClicked.connect(
             self.toggle_custom_value)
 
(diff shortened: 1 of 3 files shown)


ismailsunni:
Hi @myarjunar 
Can you add unit test for default_select_parameter_widget.py for this behaviour?
Thanks

safe/common/parameters/default_select_parameter_widget.py
```diff
@@ -151,3 +152,29 @@ def toggle_custom_value(self):
             self.custom_value.setDisabled(False)
         else:
             self.custom_value.setDisabled(True)
+
+    def toggle_input(self):
+        """Change behaviour of radio button based on input."""
+        # If input is a field, then disable all radio button.
+        if self.input.currentText() != 'No Field':
```
Gustry: Be careful 'No Field' will be translated in the UI, so you won't get this string. You should translate it here too. But I would prefer if you have 'data' (http://doc.qt.io/qt-4.8/qcombobox.html#itemData) in the combobox. I don't like string comparaison from the UI (we got some issues in the past between french, bahasa, ...).

myarjunar: thanks @Gustry I will change it

myarjunar:
ok sure :+1:  @ismailsunni

safe/common/parameters/default_select_parameter_widget.py
myarjunar: can I use `no_field = tr('No Field')` definition instead? https://github.com/inasafe/inasafe/blob/develop/safe/definitions/constants.py#L15

Gustry: I don't know how @ismailsunni build the QComboxBox. If there is some data in the comboxbox, I would really prefer this way.
But as a quick fox for now, if you don't have `data`, ok to use this definition.

ismailsunni: We have data, but the content is the same as the text.
```
            self.input.addItem(opt)
            self.input.setItemData(index, opt, Qt.UserRole)
```
I think it's fine to use that @myarjunar.

ismailsunni: Btw, `No field` is very specific. You can't put that in the default_select_parameter_widget. May be, you should add something like, a special attribute to the parameter. This attribute will store the string that will triger the behaviour. Then, you can set the string in the creation.

example:
```
if self.input.currentText() != self.special_content:
```

safe/common/parameters/default_select_parameter_widget.py
```diff
@@ -151,3 +152,29 @@ def toggle_custom_value(self):
             self.custom_value.setDisabled(False)
         else:
             self.custom_value.setDisabled(True)
+
+    def toggle_input(self):
+        """Change behaviour of radio button based on input."""
+        # If input is a field, then disable all radio button.
+        if self.input.currentText() != self.input.itemData(0, Qt.UserRole):
```
ismailsunni: Hi @myarjunar I am not sure this is a good way to compare currentText and itemData.
Actually, I suggest to make an attribute in the DefaultSelectParameterWidget class to indicate which string that will trigger the action (hide/show the radio button). So, there will be a set of this attribute in the wizard.

myarjunar: Yes, I'm sorry. My intention of doing this back then is I just want to update my remote repo with my local repo. But I also wondering about this. I think by creating an attribute, it will still checking the current combo box text or the itemData anyway. Isn't it?

PR #3817, opened 2017-02-10
Fix issue when wizard not skip empty inasafe fields.

ismailsunni:


myarjunar:
cool. maybe there will be conflicts since I'm working on the same file. So I will update inasafe-field-step behaviour test on this after this commit has been merged.

Issue #3792, opened 2017-02-07
ValueError while generating the report


Gustry:
### Problem
![screen shot 2017-02-07 at 12 02 14](https://cloud.githubusercontent.com/assets/1609292/22688608/53ba5972-ed2d-11e6-9ed4-eb3838da118f.png)

Using this dataset in debug mode.
[Dummy Data.zip](https://github.com/inasafe/inasafe/files/757351/Dummy.Data.zip)

If we don't use debug mode, we can see the exception quickly : 
![exception](https://cloud.githubusercontent.com/assets/1609292/22688714/c468e3be-ed2d-11e6-95b6-d076c59c00f7.gif)

### Expected Behaviour

* We fix this exception
* And we shouldn't hide the exception, it's more difficult to debug.

lucernae:
Fix is on the way. Apparently char d is typed while I'm trying to Ctrl+S

lucernae:
Should not occured anymore.

Issue #3787, opened 2017-02-07
Zero rows in detailed agregate result on Popoulation exposure


ivanbusthomi:
### Problem
When I tried running volcano and tsunami hazard on population exposure using aggregation, I got zero rows for each village that doesn't have affected people
![image](https://cloud.githubusercontent.com/assets/7351977/22685935/dc96748e-ed54-11e6-9532-420bb5c85baf.png)

### Expected Behaviour
Columns and rows containing only 0 or "No data" values are excluded from the tables.

### Environment
* InaSAFE : 4.0 develop
* QGIS : 2.14.10
* OS : Windows
* Data used : Nagekeo and Ambon
  * Exposure : Population
  * Hazard : Volcano and Tsunami
  * Aggregation : Villages

### Proposed Solution
make it happy :)

---
cc: @Gustry @ismailsunni @timlinux @Charlotte-Morgan @samnawi @felix-yew

lucernae:
Fix is on the way

lucernae:
Try it again @ivanbusthomi and close this ticket if it is solved.

ivanbusthomi:
Hi @lucernae , sorry for late response.
I tried using volcano hazard on landcover, population and buildings with aggregation. In the report for buildings and landcover, there are no zero rows or columns appear in aggregation result. But in population, I got no aggregation result section. Only the estimated summary, action checklist, notes and assumption, and analysis details sections.

lucernae:
Hi @ivanbusthomi, It will not show up if all the aggregation result is empty. Perhaps you could use a bigger dataset? All population calculation is using displaced with ratio 1%, so you have to at least have 100 people affected?
If this is a problem, maybe we should keep the section but tell the user that it is empty.

Gustry:
And the displacement rate is 0 for some hazards.

> If this is a problem, maybe we should keep the section but tell the user that it is empty.

I though we said we should display a message @lucernae ? At least to show them that's not a bug.

ivanbusthomi:
Hi mas @lucernae , and @Gustry 
So I tried using bigger population dataset from Worldpop, and this is how the data look like,
![image](https://cloud.githubusercontent.com/assets/7351977/23245358/658c063a-f9be-11e6-9760-29fac625b305.png)

---
 and this is the result table that I got. there is still no aggregated result.
![image](https://cloud.githubusercontent.com/assets/7351977/23245366/776f7404-f9be-11e6-82d6-ee9bfe83395e.png)

[...]

PR #3818, opened 2017-02-10
fix issue #3787: zero rows should not be displayed

lucernae:
- Zero rows should not be displayed in aggregation table
- fix issue #3792: Value Error when generating infographic
- make analysis exited if report generation encountered an error

### DIFF
--- a/safe/definitions/post_processors.py
+++ b/safe/definitions/post_processors.py
@@ -677,6 +677,8 @@ age_postprocessors = [
     post_processor_elderly,
 ]
 
+vulnerability_postprocessors = [post_processor_gender] + age_postprocessors
+
 post_processors = [
     post_processor_size,
     post_processor_size_rate,
--- a/safe/definitions/report.py
+++ b/safe/definitions/report.py
@@ -237,7 +237,7 @@ population_infographic_component = {
             },
             'vulnerability': {
                 'header': tr('Vulnerability'),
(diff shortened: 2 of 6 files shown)


lucernae:
Ah, strange error in impact_function. It seems, no 'inasafe_fields' in the keywords.
@Gustry could you look at this too?

Gustry:
I will try during the week-end

Gustry:
Can you update the conflict ?

I tried your PR on my mac, the test is passing. I already relaunched travis, no luck.
If the test is passing on your mac too, can you make many commits ? To see which commit is failing on travis. Do you want me to try @lucernae ?

lucernae:
Yes, it passes in my mac too. I will update the PR then.

lucernae:
I described the issue in #3821 . It is in current develop, at first glance I thought it is something you're more familiar with @Gustry 😄 . The error is about inasafe_fields not found in the aggregation_preparation keywords.
It rise up in this PR because I removed some unit test with expectedFailure decorator.

Gustry:
So can you please keep the expectedFailure decorator and merge your PR (so as to work on this test on my side). I'm going to fix the other ticket you created

lucernae:
Ah, all right if that's ok with you.

Gustry:
Yes please. I can work on that this afternoon if it's merged.

It was this test : 'test_minimum_needs'

Issue #3821, opened 2017-02-13
KeyError: Flood Polygon on Population Failed if used without aggregation


lucernae:
# Problem

Using current develop

1. Load Test Layer
2. Pick Flood Polygon hazard
3. Pick Population
4. Make sure Aggregation set to Entire Area
5. Run Analysis.

Error:

```
Failure
Traceback (most recent call last):
  File "/Users/lucernae/Documents/WorkingDir/Projects/InaSAFE/inasafe/safe/report/test/test_impact_report.py", line 389, in test_minimum_needs
    self.assertEqual(return_code, ANALYSIS_SUCCESS, message)
AssertionError: ****Problem

    The following problem(s) were encountered whilst running the analysis.

 - KeyError

****Suggestion

    You can try the following to resolve the issue:

 - Check that you have the latest version of InaSAFE installed - you may have encountered a bug that is fixed in a subsequent release.
 - Check the InaSAFE documentation to see if you are trying to do something unsupported.
 - Report the problem using the issue tracker at https://github.com/inasafe/inasafe/issues. Reporting an issue requires that you first create a free account at http://github.com. When you report the issue, please copy and paste the complete contents of this panel into the issue to ensure the best possible chance of getting your issue resolved.
 - Try contacting one of the InaSAFE development team by sending an email to info@inasafe.org. Please ensure that you copy and paste the complete contents of this panel into the email.

*****Details

    These additional details were reported when the problem occurred.

 - inasafe_fields

*****Diagnostics

 1. In file "/Users/lucernae/Documents/WorkingDir/Projects/InaSAFE/inasafe/safe/impact_function/impact_function.py", line 1017, in run self._run()
 2. In file "/Users/lucernae/Documents/WorkingDir/Projects/InaSAFE/inasafe/safe/utilities/profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
 3. In file "/Users/lucernae/Documents/WorkingDir/Projects/InaSAFE/inasafe/safe/impact_function/impact_function.py", line 1160, in _run self.aggregation_preparation()
 4. In file "/Users/lucernae/Documents/WorkingDir/Projects/InaSAFE/inasafe/safe/utilities/profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
[...]

PR #3825, opened 2017-02-13
avoid adding default ratios to a raster exposure layer

Gustry:
Ticket #3821


lucernae:
LGTM. Thank you very much @Gustry !

Issue #3824, opened 2017-02-13
Map extent not properly used in Map Report


lucernae:
### Expected Behaviour

- [x] Extent should hover to analysis area being used
- [ ] Extent should hover to aggregation area being selected for analysis

CC @Charlotte-Morgan @timlinux

Gustry:
I think you can just focus on the first checkbox

lucernae:
Closed

PR #3830, opened 2017-02-13
fix issue #3824 Map extent not properly used in Map Report

lucernae:
Map report now auto zoom in to analysis extent only, instead of all included layers.

screen shot 2017-02-13 at 22 57 32

### DIFF
--- a/safe/report/extractors/composer.py
+++ b/safe/report/extractors/composer.py
@@ -222,12 +222,16 @@ def qgis_composer_extractor(impact_report, component_metadata):
         # place exposure at the bottom
         layers.append(impact_report.impact_function.exposure)
 
+    # default extent is analysis extent
+    if not qgis_context.extent:
+        qgis_context.extent = impact_report.impact_function.analysis_extent
+
     map_elements = [
         {
             'id': 'impact-map',
             'extent': qgis_context.extent,
             'grid_split_count': 5,
-            'layers': layers
+            'layers': layers,
         }
     ]
     context.map_elements = map_elements
--- a/safe/report/processors/default.py
+++ b/safe/report/processors/default.py
@@ -268,25 +268,33 @@ def qgis_composer_renderer(impact_report, component):
         item_id = map_el.get('id')
         split_count = map_el.get('grid_split_count')
         layers = map_el.get('layers')
+        map_extent_option = map_el.get('extent')
         composer_map = composition.getComposerItemById(item_id)
         """:type: qgis.core.QgsComposerMap"""
         if composer_map:
             composer_map.setKeepLayerSet(True)
             composer_map.setLayerSet(
                 [l.id() for l in layers])
-            # composer_map.zoomToExtent(square_extent)
-            extent = QgsRectangle()
-            extent.setMinimal()
-            for l in layers:
-                extent.combineExtentWith(l.extent())
+            if map_extent_option and isinstance(
+                    map_extent_option, QgsRectangle):
+                # use provided map extent
+                extent = map_extent_option
+            else:
+                # if map extent not provided, try to calculate extent
+                # from list of given layers. Combine it so all layers were
+                # shown properly
+                extent = QgsRectangle()
+                extent.setMinimal()
+                for l in layers:
+                    # combine extent if different layer is provided.
+                    extent.combineExtentWith(l.extent())
 
-            canvas_extent = extent
-            width = canvas_extent.width()
-            height = canvas_extent.height()
+            width = extent.width()
+            height = extent.height()
             longest_width = width if width > height else height
             half_length = longest_width / 2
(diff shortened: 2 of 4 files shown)


safe/report/test/test_impact_report.py
```diff
@@ -980,8 +980,10 @@ def test_qgis_map_pdf_report(self):
         # insert layer to registry
         layer_registry = QgsMapLayerRegistry.instance()
         layer_registry.removeAllMapLayers()
-        rendered_layer = impact_function.exposure_impacted
+        rendered_layer = impact_function.impact
         layer_registry.addMapLayer(rendered_layer)
+        for layer in impact_function.outputs:
+            layer_registry.addMapLayer(layer)
```
Gustry: If I understand, you will add the impact layer twice. The first element in `outputs` is the impact layer too. Is-it expected ?

lucernae: Ah yes, I think I overlook this. I will check different parts too.
Thanks for the heads up

timlinux:
I personally think we should be using the analysis layer rather - can you please update to use analysis layer.

Gustry:
The analysis layer has the full analysis extent. By using the impact, you will have smaller area @lucernae

lucernae:
So, what should I use? analysis extent or impact extent?
I'm not clear on how it should be?
Shouldn't it display smaller area?

Gustry:
> Shouldn't it display smaller area?

No it should display the area requested by the user.

> So, what should I use? analysis extent or impact extent?

Analysis layer

lucernae:
Is this correct then?
I already used analysis extent, and for this screenshot, the area requested is the smaller one right (when we choose aggregation area)?

safe/report/extractors/composer.py
```diff
@@ -222,12 +222,16 @@ def qgis_composer_extractor(impact_report, component_metadata):
         # place exposure at the bottom
         layers.append(impact_report.impact_function.exposure)
 
+    # default extent is impact extent
```
Gustry: The analysis extent, not the impact extent. It's not correct I think

safe/report/extractors/composer.py
```diff
@@ -222,12 +222,16 @@ def qgis_composer_extractor(impact_report, component_metadata):
...
+    if not qgis_context.extent:
+        qgis_context.extent = impact_report.impact_function.analysis_extent
```
Gustry: You take the analysis extent, right

safe/report/processors/default.py
```diff
@@ -268,25 +268,29 @@ def qgis_composer_renderer(impact_report, component):
...
-            height = canvas_extent.height()
+            if map_extent_option and isinstance(
+                    map_extent_option, QgsRectangle):
+                extent = map_extent_option
+            else:
+                extent = QgsRectangle()
+                extent.setMinimal()
+                for l in layers:
+                    # combine extent if different layer is provided.
+                    extent.combineExtentWith(l.extent())
```
Gustry: I'm not sure  for the logic between L278 and L286

Gustry:
> Is this correct then?
I already used analysis extent

So I trust you ahah ! ;-)

safe/report/extractors/composer.py
lucernae: ah, ok. so I should change this comment.

safe/report/extractors/composer.py
lucernae: Yes, default to analysis_extent if qgis_context.extent was not set before report generation.
So, this is one is the correct approach, right?

safe/report/processors/default.py
lucernae: This is if the extent were not passed by extractor, for example if we use different extractor other than default (which pass analysis_extent as default). Then the map needs to find some extent to show as default, so I combined all extent for all layers passed by extractor.
For the dock use case, it will not reach the else part. This is only useful for realtime or headless where it needs custom extractor (not inasafe default).

lucernae:
😄 I'm just afraid if I misunderstood the problem, because there is also requested_area_extent in impact_function, and I'm not sure the difference between them.

So, conclusion (CMIIW): 
- Use analysis extent
- analysis extent is not the same as impact extent (I misunderstood this one, I guess. So I put impact extent in the comment by mistake :p)

Gustry:
To sum up :)
* requested_area_extent ≠ analysis extent (the first one is the extent requested by the user, the second one is the real extent used in the IF) You can't use requested_area_extent because it's null if you use an aggregation layer.
* analysis layer ≈ analysis extent (first one is the layer, the second one is a geometry)
* analysis extent > impact extent

lucernae:
Thank's @Gustry !
I will merge it once it passed.

PR #3831, opened 2017-02-13
Strings

Gustry:


safe/report/extractors/analysis_result.py
```diff
@@ -117,7 +117,7 @@ def analysis_result_extractor(impact_report, component_metadata):
 
     report_fields = [
         total_affected_field,
-        total_unaffected_field,
+        total_not_affected_field,
```
lucernae: This will conflicts with my PR: #3832.
I think you should merge first then.

Gustry: This is a good question. I will check

PR #3833, opened 2017-02-13
revamp of the hazard conditional styling

Gustry:
### What does it fix ?
* Description : Revamp of the hazard conditional styling.

screen shot 2017-02-13 at 19 49 20

As it breaks the alignment (bug in QGIS), it's not enabled by default. You need to go in inasafe settings to enable this feature.


Gustry:
@timlinux I'm confused.
I tried without formatting  it's also not aligned.
screen shot 2017-02-13 at 21 22 43

Even before this PR. Is-it aligned for you ? Maybe it's because the icon is aligned that we notice this bug which always there !

Issue #3678, opened 2017-01-25
Translation: we need to translate the code


Charlotte-Morgan:
# problem
InaSAFE v 4.0 has many new strings to be translated

# proposed solution
@felix-yew will help to translate short strings with no tags.
@ivanbusthomi will help to translate more complicated english terms and strings with tags

timlinux:
Thanks very much! @ismailsunni to provide support if they need any help with the process...

ismailsunni:
Sure, I will help :)

ismailsunni:
Some comments:

English text | Your translation |  Proper translation | Notes
----------- | ---------------- | ------------------ | ------
from {number_affected} affected | dari {jumlah_terdampak} terdampak | dari {number_affected} terdampak | Do not translate sting inside { } 
Analysis initialisation | Analisis Inisialisasi | Inisialisasi Analisis | 
Hazard preparation | Persiapan ancaman bencana | Persiapan ancaman | hazard = ancaman
Aggregate hazard |  | ancaman teragregasi | 
Global Default | Default Umum | Default Global | 

I have fixed the one that I reviewed.

I will add more cc @felix-yew @ivanbusthomi

ismailsunni:
Propossed date for the translation:
- [ ] 9 Feb 2017 : All strings are translated
- [ ] 10 - 13 Feb 2017 : Additional urgent strings / translation
- [ ] 14 Feb 2017 : Merging translation to develop branch

cc @timlinux @Charlotte-Morgan @ivanbusthomi @felix-yew @Gustry

Gustry:
> hazard = ancaman

@ismailsunni You should use the glossary for that : 
https://www.transifex.com/inasafe/inasafe-code/glossary/fr/

Sometimes, one english word can have many equivalents in French. Using the glossary, I put one word in French so I always use the same word later. I also put notes sometimes.

ismailsunni:
@Gustry ah thanks, I forget that transifex has it. I will check

Btw, I just push latest strings to transifex. Hopefully there won't be much change later.

Gustry:
> Btw, I just push latest strings to transifex. Hopefully there won't be much change later.

We should have merge this PR : https://github.com/inasafe/inasafe/pull/3775

ivanbusthomi:
Hi mas @ismailsunni ,
I check today there is no string to be translated for Indonesian, but some to be reviewed.
And the glossary feature in Transifex is a great help to standardize the translation of some words and terms.

[...]

PR #3842, opened 2017-02-14
Update translation 4.0

ismailsunni:
### What does it fix ?
* Ticket #3678 (partially)
* Description : 
   - Also fix typos
   - Also little update on compile string script


safe/impact_function/impact_function.py
```diff
@@ -696,9 +696,8 @@ def _check_layer(layer, purpose):
             message = generate_input_error_message(
                 tr('The %s layer do not have keywords.' % purpose),
                 m.Paragraph(tr(
-                    'The %s layer do not have keywords. Use the '
-                    'Use the wizard to assign keywords to the layer.'
-                    % purpose))
+                    'The %s layer do not have keywords. Use the wizard to '
+                    'assign keywords to the layer.' % purpose))
```
Gustry: I would have updated using `format` and a dictionary ;-)

Gustry:
Thanks

safe/impact_function/impact_function.py
ismailsunni: Yes, agree. I just fixed the typos, but I forgot to fix it also.

Issue #3846, opened 2017-02-14
BUG: Wizard mess up when setting exposure thresholds in key words


Charlotte-Morgan:
### problem ###
I get a python error when i try to edit valued for flood classes

Key words wizard for setting population classifications in flood hazard data (jakarta raster 2007)
![image](https://cloud.githubusercontent.com/assets/5743135/22913320/9b57ac0c-f29c-11e6-90bb-2f3d755ac837.png)

```An error has occurred while executing Python code: 

TypeError: min_max_changed() got an unexpected keyword argument 'index' 
TypeError: min_max_changed() got an unexpected keyword argument 'index'


Python version: 2.7.5 (default, May 15 2013, 22:44:16) [MSC v.1500 64 bit (AMD64)] 
QGIS version: 2.14.10-Essen Essen, 057b42a 

Python Path:
C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python/plugins\processing
C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python
C:/Users/AIFDR_ADMIN/.qgis2/python
C:/Users/AIFDR_ADMIN/.qgis2/python/plugins
C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python/plugins
C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\site-packages\matplotlib-1.3.1-py2.7-win-amd64.egg
C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\site-packages\nose-1.3.3-py2.7.egg
C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\site-packages\tornado-4.0.1-py2.7-win-amd64.egg
C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\site-packages\backports.ssl_match_hostname-3.4.0.2-py2.7.egg
C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\site-packages\certifi-14.05.14-py2.7.egg
c:\osgeo4~1\apps\python27\lib\site-packages\python_dateutil-2.1-py2.7.egg
c:\osgeo4~1\apps\python27\lib\site-packages\six-1.3.0-py2.7.egg
C:\Program Files (x86)\ArcGIS\Desktop10.2\bin
C:\Program Files (x86)\ArcGIS\Desktop10.2\arcpy
C:\Program Files (x86)\ArcGIS\Desktop10.2\ArcToolbox
[...]

PR #3848, opened 2017-02-14
Fix bug in thresholds classification.

ismailsunni:
### What does it fix ?
* Ticket #3846 
* Description : 
   - Fix bug in thresholds
   - Update min and max label

multihazard_wizard_fix_thresholds
   - Use True minimum value in raster data (no longer -99999)


Gustry:
LGTM

Issue #3849, opened 2017-02-14
Min Max Thresholds Discussion


ismailsunni:
### Problem
Based on my discussion with @Charlotte-Morgan this morning (14 Feb 2017), we agree that the minimum is excluded, but maximum value is included. Example:
```
Very low 0.01 - 0.1
Low >0.1 - 2
Medium >2 - 5
High >5 - 10
Very high >10
```

So, in the keyword wizard it will be:
`Min > [  ]  Max <= [  ]`

IIRC, in our current IF code, we use the reverse (include min, but exclude max). So, we need to update the code also. I am not sure if it's easy, but perhaps @Gustry knows better where to put the < or >= symbol.

Another discussion, in the hazard classification, we don't need the `small_number` anymore as long as we use the comparison consistently everywhere. So, I suggest to remove the small number.
For example:
```
'numeric_default_min': 0,
'numeric_default_max': (1 - small_number),
```

will be 
```
'numeric_default_min': 0,
'numeric_default_max': 1,
```

To Do:
- [x] @ismailsunni to update the wizard and the small_number
- [x] @Gustry  to update the IF

CC @Gustry @Charlotte-Morgan @timlinux

Gustry:
Is-it finished @ismailsunni ?

ismailsunni:
@Gustry almost. Waiting for #3890 to be merged.

ismailsunni:
- [ ] Add threshold and classification in the concepts.py (@ismailsunni)

PR #3856, opened 2017-02-14
fix minimum and maximum thresholds

Gustry:
### What does it fix ?
* Ticket #3849
* Description : It should fix minimum and maximum in the GIS overlay analysis.

### DIFF
--- a/safe/definitions/styles.py
+++ b/safe/definitions/styles.py
@@ -22,10 +22,10 @@ very_dark_red = QColor('#710017')
 # Map legend templates according to standards :
 # https://github.com/inasafe/inasafe/issues/3653#issuecomment-275011957
 # The space between count and exposure_unit will be automatically added if
-# the unit is not empty..
+# the unit is not empty.
 template_without_thresholds = u'{name} ({count}{exposure_unit})'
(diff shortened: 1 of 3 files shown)


safe/definitions/styles.py
```diff
@@ -22,10 +22,10 @@
...
 # https://github.com/inasafe/inasafe/issues/3653#issuecomment-275011957
 # The space between count and exposure_unit will be automatically added if
-# the unit is not empty..
+# the unit is not empty.
 template_without_thresholds = u'{name} ({count}{exposure_unit})'
 template_with_maximum_thresholds = (
-    u'{name} < {max} {hazard_unit} ({count}{exposure_unit})')
+    u'{name} <= {max} {hazard_unit} ({count}{exposure_unit})')
 template_with_minimum_thresholds = (
     u'{name} > {min} {hazard_unit} ({count}{exposure_unit})')
```
Gustry: I wanted to change this one to : 
`u'{name} > {min} <= {max} {hazard_unit} ({count}{exposure_unit})')`
But I'm not convinced.

This is the label used in the legend.

Issue #3834, opened 2017-02-13
PetaJakarta downloader fails with requests error


timlinux:
### Problem

Trying to use PetaJakarta downloader on QGIS Windows gives an error:

```
Traceback (most recent call last):
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\plugin.py", line 848, in add_petajakarta_layer
			    from safe.gui.tools.peta_jakarta_dialog import PetaJakartaDialog
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 572, in _import
			    mod = _builtin_import(name, globals, locals, fromlist, level)
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\gui\tools\peta_jakarta_dialog.py", line 18, in <module>
			    import requests
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 572, in _import
			    mod = _builtin_import(name, globals, locals, fromlist, level)
			ImportError: No module named requests
			
```

### Expected Behaviour

We should **never** use requests - all network access should go though QGIS Network Manager class to be proxy friendly.

### Environment
* InaSAFE : dev
* QGIS : 2.14.11
* OS : Windows 10

### Proposed Solution

Update to use QGIS Libs for network requests.

myarjunar:
closed in #3858

PR #3858, opened 2017-02-14
use FileDownloader instead of python's import

myarjunar:
### What does it fix ?
* Ticket #3834 
* Description : replace import with FileDownloader

### DIFF
--- a/safe/gui/tools/peta_jakarta_dialog.py
+++ b/safe/gui/tools/peta_jakarta_dialog.py
@@ -197,7 +197,7 @@ class PetaJakartaDialog(QDialog, FORM_CLASS):
         QtGui.qApp.setOverrideCursor(QtGui.QCursor(QtCore.Qt.WaitCursor))
 
         source = (
-            'https://data-dev.petabencana.id/floods'
+            'https://data.petabencana.id/floods'
             '?city=jbd&geoformat=geojson&format=json&minimum_state=1')
         # save the file as json first
         name = 'jakarta_flood.json'
@@ -213,9 +213,6 @@ class PetaJakartaDialog(QDialog, FORM_CLASS):
             overwrite)
 
         title = self.tr("Can't access API")
-        request_failed_message = self.tr(
-            "Can't access PetaBencana API: {source}").format(
-            source=source)
 
         try:
             self.download(source, output_base_file_path)
--- a/safe/gui/tools/peta_jakarta_dialog.py
+++ b/safe/gui/tools/peta_jakarta_dialog.py
@@ -11,15 +11,14 @@ Contact : ole.moller.nielsen@gmail.com
      (at your option) any later version.
 
 """
-import json
 import os
 import logging
 import time
-import requests
 
 from shutil import copy
 from PyQt4 import QtCore
 from PyQt4.QtCore import QVariant
+from PyQt4.QtNetwork import QNetworkReply
 # noinspection PyUnresolvedReferences
 # pylint: disable=unused-import
 from qgis.core import (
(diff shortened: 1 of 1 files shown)


safe/gui/tools/peta_jakarta_dialog.py
```diff
@@ -196,7 +197,7 @@ def accept(self):
         QtGui.qApp.setOverrideCursor(QtGui.QCursor(QtCore.Qt.WaitCursor))
 
         source = (
-            'https://data.petabencana.id/floods'
+            'https://data-dev.petabencana.id/floods'
```
myarjunar: ups you need to change this. We don't use data-dev.petabencana.id, we use dev.petabencana.id.

Gustry: how to talk to your self

safe/gui/tools/peta_jakarta_dialog.py
```diff
@@ -537,3 +528,26 @@ def reject(self):
...
+        :param url: URL of the zip bundle.
+        :type url: str
+
+        :param output_path: Path of output file,
+        :type output_path: str
+        """
+        downloader = FileDownloader(url, output_path)
+        try:
+            result = downloader.download()
+        except IOError as ex:
```
Gustry: ok, I see you copy/pasted from the OSM Downloader.
But doing this is a bad practice. We shouldn't re-raise an exception like this, because you are raising a new exception and we loose the previous traceback.

For now, let's remove this try/except. You can also do it in the OSM Downloader please.
Tim would like this PR to be merged.

But ideally, we should manage exception like last time. Let's do that later, ok ?

safe/gui/tools/peta_jakarta_dialog.py
```diff
@@ -537,3 +528,26 @@ def reject(self):
         # add our own logic here...
 
         super(PetaJakartaDialog, self).reject()
+
+    def download(self, url, output_path):
+        """Download file from url and write to output path.
+
+        :param url: URL of the zip bundle.
```
Gustry: This is not a zip I guess

safe/gui/tools/peta_jakarta_dialog.py
myarjunar: ok cool. so, probably I will replace this exception with the original one. Is it fine?

safe/gui/tools/peta_jakarta_dialog.py
```diff
@@ -537,3 +528,28 @@ def reject(self):
...
+        :param output_path: Path of output file,
+        :type output_path: str
+        """
+        request_failed_message = self.tr(
+            "Can't access PetaBencana API: {source}").format(
+            source=url)
+        downloader = FileDownloader(url, output_path)
+        result = downloader.download()
+        if not result:
+            raise Exception(request_failed_message)
```
Gustry: @myarjunar We should catch and display a nice error message.
Can you use a simple `display_critical_message_box` with the requested_fieled_message and a title ?

BTW, can also edit the OSM Downloader in the same way ? Or do you want to do it after this PR has been merged ?

myarjunar: maybe I will just create another PR for the OSM downloader

Gustry: ok no problem

safe/gui/tools/peta_jakarta_dialog.py
```diff
@@ -537,3 +528,32 @@ def reject(self):
...
+            "Can't access PetaBencana API: {source}").format(
+            source=url)
+        downloader = FileDownloader(url, output_path)
+        result = downloader.download()
+        if not result:
+            display_warning_message_box(
+                self,
+                self.tr('Download error'),
+                self.tr(request_failed_message))
+            raise Exception(request_failed_message)
```
Gustry: No, don't raise an exception. We never want to display an exception to the user. We want to avoid this as much as possible.

Gustry:
Looks good to me, thanks !

Issue #3861, opened 2017-02-14
Flood on buildings with no aggregation fails


timlinux:
### Problem

I get the following when trying to run flood (petajakarta flood classes) on buildings for Jakarta with no aggregation:

```

2017-02-14T17:10:28	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-02-14T17:10:55	0	Features which have been removed from exposure : 0
2017-02-14T17:11:01	0	Fields which have been renamed from exposure : {u'TYPE': 'exposure_type'}
2017-02-14T17:11:05	0	Fields which have been removed from exposure : [u'TYPE', u'BUILDING', u'TYPE', u'STRUCTURE', u'WALL_TYPE', u'ROOF_TYPE', u'LEVELS', u'ADMIN', u'ROOF_ACCES', u'CAPACITY', u'RELIGION', u'OSM_TYPE', u'FULL_ADDRE', u'NAME', u'AMENITY', u'LEISURE', u'USE', u'OFFICE', u'safe_ag', u'INUNDATED', u'safe_ag_1', u'safe_ag__2', u'safe_ag__3', u'safe_ag__4', u'safe_ag__5', u'safe_ag__6', u'safe_ag__7', u'safe_ag__8', u'safe_ag__9', u'safe_ag_10', u'safe_ag_11', u'safe_ag_12', u'safe_ag_13', u'safe_ag_14', u'safe_ag_15', u'safe_ag_16', u'safe_ag_17', u'safe_ag_18', u'safe_ag_19', u'safe_ag_20', u'safe_ag_21', u'safe_ag_22', u'safe_ag_23']
2017-02-14T17:11:05	0	inasafe_default_value is not present, we can not fill default values.
2017-02-14T17:11:22	0	Size Value Post Processor : Running
2017-02-14T17:11:27	0	Affected Post Processor : Running
2017-02-14T17:11:27	0	Displaced Post Processor : Could not run : Key population_exposure_count_field is missing in dynamic fields ['size_field', 'aggregation_name_field', 'hazard_class_field', 'affected_field', 'exposure_id_field', 'hazard_id_field', 'exposure_class_field', 'aggregation_id_field']
2017-02-14T17:11:27	0	Gender Post Processor : Could not run : Value gender_ratio is missing in keyword: ['inasafe_default_values', 'female_ratio_field']
[...]

PR #3863, opened 2017-02-14
fix issue #3861: Flood on buildings with no aggregation fails

lucernae:
### What does it fix ?
* Ticket #3861
* Description : handle big numbers in aggregation.

### DIFF
--- a/safe/report/extractors/aggregate_result.py
+++ b/safe/report/extractors/aggregate_result.py
@@ -99,7 +99,7 @@ def aggregation_result_extractor(impact_report, component_metadata):
         total_affected_value = format_number(
             feat[total_field_index],
             enable_rounding=is_rounded)
-        if total_affected_value == 0:
+        if total_affected_value == '0':
             # skip aggregation type if the total affected is zero
             continue
         item = {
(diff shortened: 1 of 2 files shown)


safe/report/extractors/aggregate_result.py
```diff
@@ -99,7 +99,7 @@ def aggregation_result_extractor(impact_report, component_metadata):
         total_affected_value = format_number(
             feat[total_field_index],
             enable_rounding=is_rounded)
-        if total_affected_value == 0:
+        if total_affected_value == '0':
```
Gustry: Maybe you should check `total_affected_value in ['0', 0]` ?
Can it happen ?

safe/report/extractors/aggregate_result.py
```diff
@@ -156,7 +156,8 @@ def aggregation_result_extractor(impact_report, component_metadata):
 
     if value_dict:
         for type_name in type_fields:
-            if int(value_dict[type_name]) == 0:
+            affected_value_string_formatted = value_dict[type_name]
+            if affected_value_string_formatted == '0':
```
Gustry: Same for here ?

safe/report/extractors/aggregate_result.py
lucernae: The value now always returned by format_number, so I think it's always a string (with decimal and thousand separator in debug mode)

safe/report/extractors/aggregate_result.py
lucernae: Same as above, it's now returned from format_number

Issue #3752, opened 2017-02-01
PetaJakarta is not working anymore


timlinux:
### Problem

In our Jan 2017 site visit we found out that PetaJakarta is moving to PetaBencana but the old API still seemed to work for fetching data. Now it seems to just return nothing....

### Expected Behaviour

Working PetaJakarta tool...

### Environment
* InaSAFE : 
* QGIS : 
* OS : 
* Data used :
  * Exposure : 
  * Hazard : 
  * Aggregation :

### Proposed Solution

@timlinux write to Tim from PJ and find out what the deal is. Consider removing this tool from 4.0 if there is no prospect of it working.

Charlotte-Morgan:
@timlinux - apologies for the delay getting you this info.
Here are contact details for the former PetaJakarta principle investigators:
Etienne e-mail: turpin@mit.edu or etienne@anexact.org
Tomas: tomasholderness@gmail.com

I hope you have success with these contacts; if not I am happy to follow up with UoW to find out if they are able to help.

lucernae:
The REST API now:

```
https://data.petabencana.id/floods?city=jbd&geoformat=geojson&format=json&minimum_state=1
```

It is returned as REST result (not as GeoJSON directly)

```
{"statusCode":200,"result":{"type":"FeatureCollection","features":[]}}
```

So we need to get the value in the "result" mapping.

Also they provided a testing server with fixed data:

```
https://data-dev.petabencana.id/floods?city=jbd&geoformat=geojson&format=json&minimum_state=1
```

timlinux:
Thanks @Charlotte-Morgan 

Just for follow up I did write to Tom who has replied with some hints. @lucernae is now looking into it. @myarjunar will you please make sure the desktop tool works too once @lucernae has figured out the procedure to follow?

Thanks

lucernae:
Ah, I already made the fix since it is straightforward. @myarjunar you can review and take over from there.

lucernae:
@myarjunar can you take over from here?
We need to refactor all PetaJakarta, petajakarta, Peta Jakarta references to PetaBencana in the code.

myarjunar:
@lucernae ok sip :+1: , I will take care of this issue

lucernae:
Thank's @myarjunar I merged the PR. You can start another one :)

Gustry:
@myarjunar Can we close this one too ?

timlinux:
[...]

PR #3866, opened 2017-02-15
Refactor. Peta Jakarta -> Peta Bencana

myarjunar:
### What does it fix ?
* Ticket #3752 
* Description : 
                     - change all occurrences of PetaJakarta, petajakarta, Peta Jakarta, and peta_jakarta with  PetaBencana, petabencana, Peta Bencana, and peta_bencana
                     - display message when there is no data available


safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -243,9 +243,19 @@ def accept(self):
 
         self.copy_keywords(output_base_file_path)
         layer = self.add_flooded_field(output_base_file_path)
-        # add the layer to the map
-        registry = QgsMapLayerRegistry.instance()
-        registry.addMapLayer(layer)
+
+        # check if the layer has feature or not
+        if layer.featureCount() <= 0:
+            message = 'There are no floods data available at this time.'
```
Gustry: at this time in this area ? I don't know if the tool takes care about extent ?
I just checked in QGIS, we don't know if it takes QGIS extent, all Jakarta extent (as it was call Peta**Jakarta**)
Maybe we should check with @lucernae or @timlinux , it something we should fix IMHO.

BTW, I think it's better to apply the tr() straight on the string, on this line.

myarjunar: the extent is defined by the requested url. `https://data.petabencana.id/floods?city=jbd&geoformat=geojson&format=json&minimum_state=1`, `city=jbd` will return the data only for Jabodetabek area. So, maybe it's better like this "at this time in Jabodetabek"?

Gustry: Thanks for giving me an answer. Is there only Jabodetabek supported ? If no, can't we add a combobox ?

timlinux: +1 for a combo to select city

lucernae: +1 for combo too

ismailsunni:
Yeah. I suggest to revert the changes in .ts file. I am not sure, but may be it can give problem if we edit the .ts file, then push to transifex. So, let's transifex do the work :) You can also ignore/exclude the i18n directory in your PyCharm. It will make PyCharm use less resource :)

myarjunar:
but I found in .ts file, there are `peta_jakarta` that refer to file name, `<location filename="peta_jakarta_help.py" line="38"/>`, I don't know but I think we need to change it maybe? since I also change every file name with `peta_jakarta` to `peta_bencana`.

myarjunar:
petabencana downloader_019
Here are the snippets. Maybe I still need some feedback until it fixed.

Gustry:
Ok, let us know when you need a review.

For the TS files, the translation script will take care of them.

ismailsunni:
> but I found in .ts file, there are peta_jakarta that refer to file name, <location filename="peta_jakarta_help.py" line="38"/>, I don't know but I think we need to change it maybe? since I also change every file name with peta_jakarta to peta_bencana.

Yes, the translation script will do that. You can try to do `make docker-update-translation-strings`

lucernae:
> but I found in .ts file, there are peta_jakarta that refer to file name

This is alright, the whole file will be overwritten anyway by translation script (the one that @ismailsunni mentioned). So you can exclude the whole *.ts files.

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -86,14 +92,24 @@ def __init__(self, parent=None, iface=None):
...
+
+        # signals
+        self.radioButton_prod.clicked.connect(self.populate_combobox)
+        self.radioButton_dev.clicked.connect(self.populate_combobox)
+
         # creating progress dialog for download
         self.progress_dialog = QProgressDialog(self)
         self.progress_dialog.setAutoClose(False)
-        title = self.tr('PetaJakarta Downloader')
+        title = self.tr('PetaBencana Downloader')
```
Gustry: Please use the same variable L95 and L112

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -554,3 +579,57 @@ def download(self, url, output_path):
...
+
+        display_warning_message_box(
+            self,
+            'output',
+            ', '.join(available_area)
+        )
+
+    def populate_combobox(self):
+        """Populate combobox."""
+        if self.radioButton_prod.isChecked():
```
Gustry: Follow our standards : radio_button_production, we don't use camelcase

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -554,3 +579,57 @@ def download(self, url, output_path):
...
+            self.source = production_api['url']
+            available_data = production_api['available_data']
+            self.comboBox.clear()
```
Gustry: We don't use camel case and please define the purpose of this combobox

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -86,14 +92,24 @@ def __init__(self, parent=None, iface=None):
...
 
-        self.setWindowTitle(self.tr('PetaJakarta Downloader'))
+        self.setWindowTitle(self.tr('PetaBencana Downloader'))
 
         self.iface = iface
 
+        self.source = None
+
+        self.buttonGroup.setExclusive(True)
+        self.radioButton_prod.setChecked(True)
```
Gustry: Please be sure that these radio buttons are hidden when we are not in developer mode.

safe/gui/tools/peta_bencana_dialog.py
myarjunar: so we need to rename all object that generated from QtDesigner? Or just the one that used in the code?

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -17,8 +17,13 @@
 
 from shutil import copy
 from PyQt4 import QtCore
-from PyQt4.QtCore import QVariant
-from PyQt4.QtNetwork import QNetworkReply
+from PyQt4.QtScript import QScriptEngine, QScriptValueIterator
```
myarjunar: this is what I've asked you before :sweat_smile: gotta change it

safe/gui/tools/peta_bencana_dialog.py
Gustry: yes you need to do it in QTDesigner too

safe/definitions/peta_bencana.py
```diff
@@ -0,0 +1,49 @@
...
+        {
+            'code': 'sby',
+            'name': 'Surabaya'
+        }
+    ]
+}
+
+production_api = {
+    'key': 'development_api',
+    'name': tr('Development API'),
```
myarjunar: how can I miss that :disappointed:

safe/gui/tools/help/peta_bencana_help.py
```diff
@@ -81,9 +81,9 @@ def content():
         'This tool requires a working internet connection and fetching '
         'data will consume your bandwidth.'))
     tips.add(m.Link(
-        'https://petajakarta.org/banjir/en/data/',
+        'https://data.petabencana.id/floods',
```
Gustry: Or maybe we should just link to `https://petabencana.id` ?

safe/gui/tools/peta_bencana_dialog.py
Gustry: Comment your block of code and remove this line I think. I don't have a strong opinion

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -243,11 +257,22 @@ def accept(self):
...
-        # add the layer to the map
-        registry = QgsMapLayerRegistry.instance()
-        registry.addMapLayer(layer)
-        self.disable_busy_cursor()
-        self.done(QDialog.Accepted)
+
+        # check if the layer has feature or not
+        if layer.featureCount() <= 0:
+            message = self.tr(
+                'There are no floods data available at this time.')
```
Gustry: Add the name of the place used.

safe/gui/tools/help/peta_bencana_help.py
```diff
@@ -8,7 +8,7 @@
 SUBSECTION_STYLE = styles.SUBSECTION_LEVEL_3_STYLE
 
 
-def peta_jakarta_help():
+def peta_bencana_help():
     """Help message for OSM Downloader dialog.
```
Gustry: Update this legacy line ;-)

safe/gui/tools/help/peta_bencana_help.py
myarjunar: or `https://docs.petabencana.id/`?

Gustry: Yes, it's fine

safe/gui/tools/help/peta_bencana_help.py
```diff
@@ -35,7 +36,7 @@ def heading():
     :returns: A heading object.
     :rtype: safe.messaging.heading.Heading
     """
-    message = m.Heading(tr('Peta Jakarta downloader help'), **SUBSECTION_STYLE)
+    message = m.Heading(tr('Peta Bencana downloader help'), **SUBSECTION_STYLE)
```
Gustry: It can be nice to be consistent. Here you use `Peta Bencana`.

safe/gui/tools/help/peta_bencana_help.py
```diff
@@ -81,9 +82,9 @@ def content():
         'This tool requires a working internet connection and fetching '
         'data will consume your bandwidth.'))
     tips.add(m.Link(
-        'https://petajakarta.org/banjir/en/data/',
+        production_api['help_url'],
         text=tr(
-            'Downloaded data is copyright the PetaJakarta contributors'
+            'Downloaded data is copyright the PetaBencana contributors'
```
Gustry: And here `PetaBencana`. According to their website https://docs.petabencana.id they always use one word `Petabencana` except one time `PetaBencana.id`. Maybe we should use `Petabencana`, what do you think ?Can you please uniform words ?

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -39,24 +40,29 @@
 from PyQt4.QtCore import QSettings, pyqtSignature, QRegExp, pyqtSlot
```
Gustry: Please use the same line between L40 and L20. You are importing QSettings twice

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -86,14 +92,36 @@ def __init__(self, parent=None, iface=None):
...
+        self.radio_button_group = QButtonGroup()
+        self.radio_button_group.addButton(self.radio_button_production)
+        self.radio_button_group.addButton(self.radio_button_development)
+
+        self.radio_button_group.setExclusive(True)
+        self.radio_button_production.setChecked(True)
+        self.populate_combo_box()
+
+        settings = QSettings()
+        developer_mode = settings.value('inasafe/developer_mode', False,
```
Gustry: We usually pull left.
```
developer_mode = settings.value(
    'inasafe/developer_mode', False, type=bool)
```

And BTW, can you please use this new class for reading settings : `safe.utilities.settings` thanks.

safe/gui/tools/peta_bencana_dialog.py
```diff
@@ -554,3 +593,53 @@ def download(self, url, output_path):
...
+    #         geometry_code = (
+    #             geometry.property('properties').property('code').toString())
+    #         available_area.append(geometry_code)
+
+    def populate_combo_box(self):
+        """Populate combo_box."""
+        if self.radio_button_production.isChecked():
+            self.source = production_api['url']
+            available_data = production_api['available_data']
+            self.combo_box.clear()
```
Gustry: Can you give a better name to this combobox ? like adding `city` for instance. We try to give explicit name.

safe/gui/tools/help/peta_bencana_help.py
myarjunar: true, they always use Petabencana,  let's use it

lucernae: I'm guessing it should be PetaBencana (it is from two words, Peta and Bencana).
Their docs API perhaps always use Petabencana, because it is using API docs generator, and auto-capitalize the project name or something.

I just assume it is PetaBencana since they even have a logo: PetaBencana.id. Or maybe even the official name is "PetaBencana.id", I don't know.

lucernae: Oops, I'm commenting right after you comment, so I didn't see your comment there. Either is fine, if we can be consistent, we can easily change it later :D.

safe/gui/tools/peta_bencana_dialog.py
myarjunar: I don't know why, but using `safe.utilities.settings` didn't give what we expected

safe/gui/tools/help/peta_bencana_help.py
myarjunar: Ah after small observation on their social media, even they are not consistent about using their name I guess :sweat_smile:  
> Either is fine, if we can be consistent, we can easily change it later

agree :+1:

safe/gui/tools/peta_bencana_dialog.py
Gustry: Yes it is ;-)
Thanks

Gustry: https://github.com/inasafe/inasafe/blob/develop/safe/gui/widgets/dock.py#L293

myarjunar: https://github.com/inasafe/inasafe/blob/develop/safe/utilities/settings.py#L75 I think this line will always make `value = True` if `expected_type = bool` and if `value` not empty or not 0. 
```
>>> bool(u'false')
True
>>> bool('false')
True
>>> bool('true')
True
>>> bool('')
False
>>> bool()
False
>>> bool(1)
True
>>> bool([])
False
```

Gustry: For your example of code above, I agree, it's normal. But 'developer_mode' is a boolean already in the QSettings.
screen shot 2017-02-16 at 17 19 57
It will not go into the `not isinstance(value, expected_type)`. So it should be fine, no ?

But I agree, there is a side effect of this kind of conversion. If you want to make a patch in another PR, you are welcome.

myarjunar: selection_021
but mine is unicode :(

safe/gui/tools/help/peta_bencana_help.py
ismailsunni: +1 to PetaBencana :)

Gustry:
Thanks @myarjunar

Issue #3812, opened 2017-02-10
Flood & Roads Analysis not appear the result when using aggregation


samnawi:
### Problem
I got the problem when run the analysis using flood and roads, but it has a little bit differ with issue #3729 .
It's working properly when I do analyze without aggregation, but the problem appeared when I run using aggregation.

Actually, the process is working, but show nothing in the end. 
1. When I use flood vector, it's just produce like the picture in below, without any information on the dock
![image](https://cloud.githubusercontent.com/assets/8871958/22813978/7cbce43c-ef82-11e6-86f5-7d38a7d035e4.png)
![image](https://cloud.githubusercontent.com/assets/8871958/22814032/cf09db5a-ef82-11e6-8d58-47e0caf44b90.png)

2. When I run the analysis using flood raster, it also produces the %s intersect with some error message. But, I can see the result on the table.
![image](https://cloud.githubusercontent.com/assets/8871958/22814207/277631e8-ef84-11e6-8e1c-afa1a94af4e1.png)
```
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1008, in run self._run()
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1183, in _run self.summary_calculation()
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1656, in summary_calculation self._aggregate_hazard_impacted, self._analysis_impacted)
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/kurnia/.qgis2/python/plugins\inasafe\safe\gis\vector\summary_3_analysis.py", line 196, in analysis_summary raise ComputationError
```

### Expected Behaviour
It's should working

### Environment
* InaSAFE : develop
* QGIS : 2.14.10
* OS : Win 7 64 bit
* Data used :
  * Exposure : DKI jakarta OSM roads
[...]

PR #3882, opened 2017-02-16
Checking why some layers wrong

Gustry:
### What does it fix ?
* Ticket #3812 
* Funded by : DFAT/WB
* Description : Add more sanity check along the IF

### DIFF
--- a/safe/gis/vector/test/test_intersection.py
+++ b/safe/gis/vector/test/test_intersection.py
@@ -25,10 +25,11 @@ class TestIntersectionVector(unittest.TestCase):
 
     def test_intersection_mask_vector(self):
         """Test we can intersect two layers, like exposure and aggregation."""
-
         exposure = load_test_vector_layer(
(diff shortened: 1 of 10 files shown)


safe/gis/vector/test/test_intersection.py
```diff
@@ -26,21 +26,21 @@ def tearDown(self):
     def test_intersection_mask_vector(self):
         """Test we can intersect two layers, like exposure and aggregation."""
 
-        hazard = load_test_vector_layer(
-            'gisv4', 'hazard', 'classified_vector.geojson')
+        exposure = load_test_vector_layer(
+            'gisv4', 'exposure', 'roads.geojson')
 
         aggregation = load_test_vector_layer(
```
ismailsunni: Is it right to have aggregation variable but in hazard directory?

Gustry: I added a comment about this. The test takes an aggregation layer, but I give its a hazard layer. THanks for review

PR #3884, opened 2017-02-17
needs profile patch. ignore empty root_directory

myarjunar:
### What does it fix ?
* Ticket #3810 
* Funded by : DFAT
* Description : # Ignore empty root_directory to avoid load min needs profile to test directory when test is running.

### DIFF
--- a/safe/gui/tools/minimum_needs/needs_profile.py
+++ b/safe/gui/tools/minimum_needs/needs_profile.py
@@ -187,26 +187,33 @@ class NeedsProfile(MinimumNeeds):
 
             return profiles_our_locale + profiles_remaining
 
-        locale_minimum_needs_dir = os.path.join(
-            str(self.root_directory), 'minimum_needs')
-        path_name = resources_path('minimum_needs')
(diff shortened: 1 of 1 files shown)


safe/gui/tools/minimum_needs/needs_profile.py
```diff
@@ -187,26 +187,33 @@ def sort_by_locale(unsorted_profiles, locale):
...
-            profile[-5:] == '.json']
-        profiles = sort_by_locale(profiles, self.locale)
-        return profiles
+        # We ignore empty root_directory to avoid load min needs profile
+        # to test directory when test is running.
+        if not self.root_directory:
+            profiles = []
+            return profiles
+
+        else:
```
ismailsunni: If you have `returns` in the  `if`, you don't have to add `else`.
Less indent :D

ismailsunni:
Thanks @myarjunar 
If you want to address my comment, you can do in the other PR. It's not mandatory.

I am getting a lot of minimum needs :D

PR #3886, opened 2017-02-17
fix impact function is None in Dock for reporting

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : Fix dock behaviour when the impact function is None for reporting


Gustry:
I'm going to merge this one. It was just one line updated. I made a copy/paste of the accept block.

Issue #3851, opened 2017-02-14
Population count and default ratios


Charlotte-Morgan:
### problem ###
I have raster population data (haha - quess where - Jakarta) and I want to run  a flood analysis. When I set the population key words i am not asked about the default age and gender profile on my data.

If I run without aggregation data, then I do not get an age gender report or detailed min needs for this analysis.

### proposed solution ###
allow users to set age gender defaults for raster population count data

This issue was mentioned  in #3763 but not addressed

cc
@samnawi

timlinux:
@Charlotte-Morgan in our discussion in JK in Jan we agreed to do ratios on exposure and global defaults only and not on the exposure layer. So it's not a regression its a requested enhancement :-P

Gustry:
So the graph is not finished : https://github.com/inasafe/inasafe/issues/3763#issuecomment-279165240 right ?
We missed if it's a raster population exposure. In this case, we don't take care of aggregations ratios, right ?

Charlotte-Morgan:
Ok. Sounds like i should not expect to set population attributes for raster population data. But the missing report implies that my global defaults were also not applied.  If they were, I should get an age gender report for any population analysis

ismailsunni:
Hi @Gustry 
I did a chat with @Charlotte-Morgan on [February 17, 2017 5:12 PM](https://gitter.im/inasafe/inasafe?at=58a6cca40ad50ac31529d521)

So, it seems I don't need to restore the raster default ratio step in the wizard. But, we just need to make sure the result has gender report for population analysis.

cc @timlinux also.

timlinux:
@Gustry @ismailsunni Yes correct - just make sure the defaults are read but don't add back support for defaults on the exposure layer. Thanks.

Gustry:
I think we can close this one @Charlotte-Morgan, it has been merged in develop.

ismailsunni:
Closing this.

PR #3888, opened 2017-02-17
manage default ratios with a raster exposure

Gustry:
### What does it fix ?
* Ticket #3851 
* Funded by : DFAT
* Description : 
  * Manage default ratios with a raster exposure

Wait for 3889 to be merged


Gustry:
Thanks, It's on my todo list too. But it's not code, only geojson inside.

Gustry:
Can we rename gisv4 to 'quick_tests' or something like that ? I would like to keep this small dataset in a folder. But I can merge these files into safe/test/data if you prefer.

timlinux:
@Gustry lets not worry about it right now if there is not any code still using v4 naming that is good for me.

PR #3889, opened 2017-02-17
add tsunami vector and population raster

Gustry:
### What does it fix ?
* Ticket #
* Funded by : DFAT
* Description : Add tsunami vector and population raster to test data


Gustry:
This PR was only about test data.

PR #3890, opened 2017-02-20
Remove small number hacky fix.

ismailsunni:
### What does it fix ?
* Ticket #3849 
* Funded by : DFAT
* Description : 
   - We no longer need small number as hacky number. We prefer to use > or < operator in the calculation.

### DIFF
--- a/safe/definitions/concepts.py
+++ b/safe/definitions/concepts.py
@@ -469,6 +469,33 @@ concepts = {
             }
         ],
     },
+    'thresholds': {
+        'key': 'thresholds',
+        'description': tr(
+            'A range that defined with minimum and maximum value. In InaSAFE '
+            'we exclude the minimum value but include the maximum value. In '
+            'mathematical expression: minimum value < x <= maximum value. It '
+            'is used for doing classification for continuous data.'),
+        'citations': [
+            {
+                'text': '',
+                'link': u''
+            }
+        ],
+    },
+    'value_maps': {
+        'key': 'value_maps',
+        'description': tr(
+            'A mapping of a set of values to another set of unique values. '
+            'Each unique value represents a particular class. It is used for '
+            'doing classification for classified (non-continuous) data'),
+        'citations': [
+            {
+                'text': '',
+                'link': u''
+            }
+        ],
+    },
     # Boilerplate for adding a new concept...
     #  '': {
     #    'description': tr(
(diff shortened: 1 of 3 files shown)


safe/definitions/hazard_classifications.py
```diff
@@ -4,7 +4,7 @@
 See https://github.com/inasafe/inasafe/issues/2920#issuecomment-229874044
 to have a table showing you classes of each kind of hazard.
 """
-from safe.definitions import concepts, small_number
```
Gustry: It could be nice to add a comment here about we define max and min.

ismailsunni: Yes, I agree. I will put in the `concepts.py` like @timlinux said in the stand up.

Gustry: If you can copy/paste also here with comments in this file, it's quicker, for developers.

ismailsunni: Done.

safe/definitions/concepts.py
```diff
@@ -469,6 +469,33 @@
             }
         ],
     },
+    'thresholds': {
+        'key': 'thresholds',
+        'description': tr(
+            'A range that defined with minimum and maximum value. In InaSAFE '
+            'we exclude the minimum value but include the minimum value. In '
+            'mathematical expression: minimum value < x <= maximum value. It '
+            'is used for doing classification for continuous data.'),
```
timlinux: +            'A range defined with minimum and maximum values. In InaSAFE '
 +            'we exclude the minimum value but include the maximum value. '
 +            'Expressed mathematically: minimum value < x <= maximum value. Ranges '
 +            'are used for performing classification of continuous data.'),

safe/definitions/concepts.py
```diff
@@ -469,6 +469,33 @@
...
+                'link': u''
+            }
+        ],
+    },
+    'value_maps': {
+        'key': 'value_maps',
+        'description': tr(
+            'A mapping of a set of values to another set of unique values. '
+            'Each unique value represents a particular class. It is used for '
+            'doing classification for classified (non-continuous) data'),
```
timlinux: 'A conceptual mapping between one set of unique values and another set of unique values. Each unique value represents a particular class. It is used to express terms or concepts from one classification system in another classification system and only applies to non-continuous data. For example a value map can be used to express local 
names for entities (e.g. street type: "jalan") into generic concepts (e.g. street type: "residential").'

ismailsunni: Done. Thanks @timlinux

Issue #3874, opened 2017-02-16
Earthquake Hazard MMI


Charlotte-Morgan:
### problem ###
Earthquake hazard is not well defined in terms of MMI classes

### proposed solution ###
We need to make sure that MMI classes are whole numbers (no decimal places) and preferably in roman numerals.
ie in this edit option in the KW wizard - only whole numbers should be available for classified data

We need to update the Earthquake MMI classes to include a complete MMI classification where
MMI I is MMI 1
MMI II is MMI II etc

ie each class has a unique MMI value that is mutually exclusive

I believe this should be used for both classified and continuous data

Charlotte-Morgan:
@hadighasemi - can we please discuss this when you are in Jakarta
@RikkiWeber - FYI

hadighasemi:
Sure :)


Sent from my Samsung GALAXY S5

-------- Original message --------

Charlotte-Morgan:
here is a source for the MMI scale descriptions
https://earthquake.usgs.gov/learn/topics/mercalli.php

timlinux:
@Gustry, @myarjunar  and @ismailsunni I chatted with @Charlotte-Morgan about this on Friday. We probably want to include this in 4.0. The main things to do will be:

* Update the EQ Shakemap Convertor tool to create a classified raster (where half decimal on either side of the mmi level get assigned eg. 0.5 - 1.5 get reclassed as 1)
* Ideally we should prevent any smoothing or interpolation of the imported grid dataset (currently we use invdist)
* Add classifications for MMI with descriptions of each MMI and displacement rate (>=6 displacement rate = 100% for now)
* Fataility rates will be left blank - they should still come from the IF - make sure to leave a note in hazard_classification
* Update EQ IF to read the displacement rates from MMI classification rather

Charlotte-Morgan:
I can do the "add hazard classifications with descriptions and displacement rates" if you want

timlinux:
That would be great thanks @Charlotte-Morgan 

See also https://en.wikipedia.org/wiki/Mercalli_intensity_scale#Modified_Mercalli_Intensity_scale (I am sure you have that anyway from some more official source). We do have the hazard class colours in our code somewhere. Hopefully we will stick with those since they seem to be somewhat standard (I think USGS uses them too).

ivanbusthomi:
Hi @Charlotte-Morgan , this is the color code that you asked. You can find it in the code in /safe/utilities/styling.py at line 564

[...]

PR #3891, opened 2017-02-20
Mmi classes

Charlotte-Morgan:
### part fix MMI scales, colours and displacement rates?
* Ticket #3874
* Funded by : DFAT
* Description : update of MMI intensity classes, with standard colours and descriptions. Displacement rates estimated.

Apologies if its not the way you want it integrated. Please change as required.


safe/definitions/styles.py
```diff
@@ -20,9 +20,9 @@
 very_dark_red = QColor('#710017')
 MMI_white = QColor('#ffffff')
 MMI_blue = QColor('#209fff')
-MMI_light_blue = QColo ('#00cfff')
+MMI_light_blue = QColor('#00cfff')
```
timlinux: I wonder if it would not be a better idea to name these by level 

```
MMI_0 = QColor('#ffffff')
```

I think it would carry more implicit meaning than colour names. For H/M/L legends we could make verbal aliases e.g.:

```
MMI_high = MMI_red
```

Anyway that is just a thought and certainly not a blocker to merging. LGTM

safe/definitions/hazard_classifications.py
```diff
@@ -211,6 +223,278 @@
...
+                {
+                    'text': None,
+                    'link': u'https://earthquake.usgs.gov/learn/topics/'
+                            u'mercalli.php'
+                }
+            ]
+        }
+    ],
+    'exposures': [
+        exposure_land_cover,
```
timlinux: Don't we want to limit EQ to roads/structures and population?

safe/definitions/hazard_classifications.py
```diff
@@ -123,9 +135,9 @@
         exposure_structure
     ]
 }
-
-earthquake_mmi_hazard_classes = {
-    'key': 'earthquake_mmi_hazard_classes',
+# do not use these old classes until we can trace the origin
+earthquake_mmi_old_hazard_classes = {
+    'key': 'earthquake_mmi_old_hazard_classes',
```
ismailsunni: Do we really need to change the `key` here? I saw @Charlotte-Morgan add new classes but with different `key`. So, it should be fine. And I think it's better to change the `name` since the user will not know about the `key` or the variable name.
Just bear in mind, changing the `key` means we need to update all layer's metadata that we already assigned with that key.

timlinux: +1 to keep the old key - we can add remove it from hazard types if we want to disabled it.

Charlotte-Morgan:
yay - Travis says yes !
@ismailsunni @Gustry  @timlinux - any further advice before merging this one?

safe/definitions/hazard_classifications.py
```diff
@@ -203,8 +215,276 @@
...
+            ]
+        },
+        {
+            'key': 'II',
+            'value': 2,
+            'color': MMI_2,
+            'name': tr('II'),
+            'affected': True,
+            'description':
+                tr('Felt only by a few persons at rest,especially on upper '
```
timlinux: rest, especially

safe/definitions/hazard_classifications.py
```diff
@@ -123,7 +135,7 @@
         exposure_structure
     ]
 }
-
+# do not use these old classes until we can trace the origin
```
Gustry: Can you clarify what should we do with this class ?
Should this class be available in the wizard ? Should we use only the new one now ?

Gustry:
@Charlotte-Morgan You also need to add the classification here : https://github.com/inasafe/inasafe/blob/develop/safe/definitions/hazard.py#L154

Charlotte-Morgan:
@Gustry - I think the old class can be used for classified vectors?

Charlotte-Morgan:
@timlinux - I have added earthquake_fatality_model_limitations in exposure.py 
Maybe you can find a good place to add these in the help ?

safe/definitions/exposure.py
```diff
@@ -47,22 +49,43 @@
...
     'earthquake_pager_notes': [   # these are earthquake Pager specific notes
         tr('Fatality model is from Population Vulnerability '
             'Pager Model.'),
     ],
     'earthquake_itb_notes': [   # these are earthquake ITB specific notes
-         tr('Fatality model is from Institut Teknologi Bandung 2012.'),
+        tr('Fatality model is from Institut Teknologi Bandung 2012.'),
 
     ],
+    'earthquake_fatality_model_limitations': [# notes provided by Hadi Ghasemi
```
Gustry: Travis said : 
> ./safe/definitions/exposure.py:71:47: E261 at least two spaces before inline comment

So you need : 
`'earthquake_fatality_model_limitations': [  # notes provided by Hadi Ghasemi`

safe/definitions/hazard_classifications.py
Gustry: You forgot to fix the comma @Charlotte-Morgan ? :D

timlinux:
> @timlinux - I have added earthquake_fatality_model_limitations in exposure.py
> Maybe you can find a good place to add these in the help ?

Done - please see #3906

Issue #3749, opened 2017-02-01
help_text should be the long information in definitions


timlinux:
### Problem

We need to review definitions and make sure that help_text is always the longer information for a field and description is always the shorter one. This will avoid issues like this:

![screen shot 2017-02-01 at 9 37 00 am](https://cloud.githubusercontent.com/assets/178003/22498592/0c6cbd64-e862-11e6-8d14-06df86a6e84f.png)

![screen shot 2017-02-01 at 9 37 08 am](https://cloud.githubusercontent.com/assets/178003/22498596/0fa3bfaa-e862-11e6-99b0-8fa70221df4a.png)

### Expected Behaviour

Collapsed help triangle should show short version of the help always.

### Proposed Solution

Tim review and fix.

timlinux:
@ismailsunni to make sure code usages correct, @timlinux to make sure definitions are correct.

ismailsunni:
![0fa3bfaa-e862-11e6-99b0-8fa70221df4a](https://cloud.githubusercontent.com/assets/1421861/22503638/39f5d91c-e8a5-11e6-8706-93d0173a2f22.png)

Just to confirm:
- red arrow = description, shorter
- blue arrow = help text, longer

isn't it?

timlinux:
Confirmed @ismailsunni

ismailsunni:
Done in code @timlinux

ismailsunni:
Hi @timlinux I made a unit test here to compare the help text and description. I am not sure if we need this or not : https://github.com/ismailsunni/inasafe/tree/unit_test_definitions

timlinux:
@ismailsunni thanks for that. I guess there could be some edge cases where the description is not actually longer than the text, but as long as the test fails with a clear message, it should be easily fixable by making the help_text more verbose....

PR #3892, opened 2017-02-20
Parameter and Wizard polishing

ismailsunni:
### What does it fix ?
* Ticket #3579 #3749 
* Funded by : DFAT
* Description : 
   - Swap between help_text and descriptions in wizard (#3749)
   - Add min max in custom value in InaSAFE field (#3579)
   - Always set first tab in InaSAFE option to be opened.


timlinux:
> Swap between help_text and descriptions in wizard (#3749)

Thanks for that @ismailsunni !

PR #3700, opened 2017-01-26
Fix batchrunner

ivanbusthomi:
Fix batchrunner map pdf not showing proper layer due to some strange bug where loaded layer is not recognised by InaSAFE or QGIS

PR #3894, opened 2017-02-20
change relative path to absolute path

myarjunar:
### What does it fix ?
* Ticket #3700 
* Funded by : DFAT
* Description : Currently batch runner cannot read relative path, for now we need the save scenario tool to give absolute path.


timlinux:
LGTM

PR #3896, opened 2017-02-20
Use better displacement rates for PetaBencana floods based on per. co…

timlinux:
…mm with @charlotte-morgan in gitter. Funded by DFAT.

### What does it fix ?

Uses better displacement rates for PetaBencana floods

* Ticket #3667
* Funded by :  DFAT
* Description : Set displacement rate for high to 1.0 (100%) and medium to 0.03 (3%)


Gustry:
Looks good to me, thanks !

Issue #3898, opened 2017-02-20
Detailed report section


lucernae:
I had some question about this section here in reporting standard:

<img width="927" alt="screen shot 2017-02-20 at 20 17 38" src="https://cloud.githubusercontent.com/assets/831865/23126598/fedc5716-f7a9-11e6-8594-0838bfb9160e.png">

These 3 cases were seemingly very different pattern. So, I had some questions:

1. In first sample (people or population) where can I get the values for male column? Also, was it coming from postprocessors calculations? If it is, then was it a duplicate of the information in aggregation report for population (the one with female and age section)
2. In the second sample, I guess we're not favoring this one since it is covered by the third sample? So I'm assuming you want a building report where the affected column contains each hazard breakdown?
3. In the third sample, I'm assuming, we also wants columns with total affected, total not affected, and total not exposed?

lucernae:
If it is possible, I would like to have all remain consistent. In reporting standard, the sentence/phrase: "Length of roads" does not have a unit (m) on it. So, where should we move the unit (m)?
In #2691 , there is a suggestion to put it on Road type. Is this still valid? Other option is to put it on table header, like in Affected (m).

Charlotte-Morgan:
Hi @lucernae - i will try to add some detail.
`In first sample (people or population) where can I get the values for male column? Also, was it coming from postprocessors calculations? If it is, then was it a duplicate of the information in aggregation report for population (the one with female and age section)`
 - we do not calculate male specifically, but you can get male by subtracting female from total.

The draft tables just give an idea of how we can progress from general information to more detailed information. In some cases the data can be combined into a single table.

Maybe refer to the overall examples for an idea of how we progress from general to detailed info in a single report.
![image](https://cloud.githubusercontent.com/assets/5743135/23151799/153c6f10-f830-11e6-96b7-d01089a1d681.png)

Charlotte-Morgan:
Regarding the question about units; number of people and buildings does not have a unit in english but a unit is required in Indonesian.
I prefer to use the unit in the table heading. You will see in the examples:
Estimated number of people 
Estimated length (m) of roads

I would not add units to type. I think it is much better to use a table title and include units there. If not it belongs with each hazard class as it is in the total in #2691

Gustry:
@lucernae Can we close this ticket ?

lucernae:
Yes we can close this I guess.

PR #3899, opened 2017-02-20
Improve report table

lucernae:
### What does it fix ?
* Tickets:
   - #2691
   - #3898
* Funded by : DFAT
* Description : 
  - fix issue #2691
  - add separate background highlight in detailed report
  - capitalize headers

### DIFF
--- a/resources/report-templates/standard-template/jinja2/aggregation-postprocessors.html
+++ b/resources/report-templates/standard-template/jinja2/aggregation-postprocessors.html
@@ -2,7 +2,7 @@
     {% for _, value in sections.iteritems() %}
         {% if value and value.rows %}
         <h5 class="info">
-            <i class="icon-info-sign icon-white"></i> {{ value.header }}
+            <i class="icon-info-sign icon-white"></i> {{ value.header|capitalize }}
         </h5>
         <table class="table table-striped condensed">
             <thead>
@@ -23,7 +23,7 @@
                 <tr>
                     {% for column in row %}
                         {% if loop.first %}
-                        <th class="text-left">{{ column }}</th>
+                        <th class="text-left">{{ column|capitalize }}</th>
                         {% else %}
                         <td class="text-right">{{ column }}</td>
                         {% endif %}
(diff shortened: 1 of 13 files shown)


lucernae:
Sample of highlighted affected column with double header:
screen shot 2017-02-21 at 00 05 14

timlinux:
Beautiful work @lucernae ! I wonder if the highlight is not a bit too heavy? Perhaps just putting an outline in charcoal around the group might look better?

timlinux:
Hmm maybe that will be tricky since you deal with it cell by cell I guess - perhaps just a lighter accent colour would work well for now....

Gustry:
It rocks @lucernae  !
I agree, a lighter color will do the trick I think

lucernae:
@timlinux @Gustry could you review definitions/units.py later?
I added 'measure' key because I needed words like 'Length' (something that we measure) in the report.

lucernae:
Is this lighter now?
screen shot 2017-02-21 at 02 34 33

Or, well since we can use css, we can do this too if you prefer:
screen shot 2017-02-21 at 03 21 53

Or both:
screen shot 2017-02-21 at 03 22 53

lucernae:
The previous one seems to be not a good example. Actually, if there is a not affected group, it will show up something like this:

screen shot 2017-02-21 at 03 32 13

We could also styled each group if you want (e.g, give a different background color).

ismailsunni:
Looks Beautiful For Me !!

Charlotte-Morgan:
hey @lucernae - this looks really neat.
The only other think in the standard is the title for the table - ie what is the table about.
But for adding double headers and highlighting - this is great. LVGTM

timlinux:
Can we go with this one @lucernae

lucernae:
@Charlotte-Morgan Can you point out what's still missing?
I did assumed that the table title is the one in blue.
Maybe that's not what you meant?
screen shot 2017-02-21 at 00 05 14

lucernae:
I will merge this for now and have the table title taken care in separate PR.

PR #3900, opened 2017-02-20
setting up docker with an environment variable

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : To be able to set the test package with an environment variable. It's very convenient when a test is not passing on travis, but the test is ok on the mac.

```
docker run -d --name qgis-testing-environment -v /Users/etienne/dev/inasafe/inasafe:/tests_directory -e ON_TRAVIS=True -e DISPLAY=:99 -e TESTING_PACKAGE='safe.impact_function' kartoza/qgis-testing:boundlessgeo-2.14.7

docker exec -it qgis-testing-environment sh -c "qgis_setup.sh inasafe"

docker exec -it qgis-testing-environment sh -c "qgis_testrunner.sh test_suite.test_environment"
```

@timlinux It could be very nice with can have QGIS in the UI according to your blogpost : http://kartoza.com/en/blog/how-to-run-a-linux-gui-application-on-osx-using-docker/
Is-it still up to date with the new docker ?


ismailsunni:
LGTM

Issue #3879, opened 2017-02-16
Make maps showing all layer related with analysis


lucernae:
### Problem

Only impact layer and exposure (toggled in InaSAFE option) that is shown.

### Proposed Solution

- Hazard, exposure and aggregation should be able to be shown in map report.
- Create a toggle in InaSAFE option to toggle showing up only impact layer, or all related layer in analysis.
- Proposed layer order (bottom to top): exposure, hazard, outputs, aggregation

PR #3901, opened 2017-02-20
fix issue #3879: Make maps showing all layer related with analysis

lucernae:
### What does it fix ?
* Ticket #3879
* Funded by : DFAT
* Description : 
   - Add new InaSAFE option: show only impact layer
   - Add new provenance items: original hazard, exposure, and aggregation layer. This is needed to draw map. Since after analysis, hazard, exposure and aggregation were cleaned up, we can't show it using that layer.

### DIFF
--- a/safe/definitions/default_settings.py
+++ b/safe/definitions/default_settings.py
@@ -7,6 +7,7 @@ inasafe_default_settings = {
     'visibleLayersOnlyFlag': True,
     'set_layer_from_title_flag': True,
     'setZoomToImpactFlag': True,
+    'set_show_only_impact_on_report': False,
     'setHideExposureFlag': False,
     'useSelectedFeaturesOnly': False,
     'useSentry': False,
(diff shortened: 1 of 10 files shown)


lucernae:
Show all related layers:
order (top to bottom): aggregation, impact, hazard, exposure
screen shot 2017-02-21 at 01 38 49

Show only impact (with exposure):
screen shot 2017-02-21 at 01 40 39

safe/definitions/default_settings.py
```diff
@@ -7,6 +7,7 @@
     'visibleLayersOnlyFlag': True,
     'set_layer_from_title_flag': True,
     'setZoomToImpactFlag': True,
+    'set_show_only_impact_on_report': False,
```
lucernae: the naming is not consistent but I'm trying to follow python convention.

ismailsunni: Yeah, +1 for lower case. The other are legacy.

timlinux:
Much better thanks @lucernae  - can you make the tests pass so we can merge?

PR #3919, opened 2017-02-22
fix notes failing when there is no displacement

lucernae:
### What does it fix ?
* Ticket #: No ticket number
* Funded by : DFAT
* Description : Small fix for possible key error.


lucernae:
Please just merge when it is passed

Issue #3713, opened 2017-01-27
Jakarta aggregation data errors


Charlotte-Morgan:
### problem
In Version 4.0 the Jakarta district and subdistrict boundary data used for aggregation show up  as being invalid.

### proposed solution
we need to fix the data (using spatial tools - please do not redigitise it!!)

###
make it happy :)

Gustry:
For subdistrict, I could have the error easily : 
![screen shot 2017-01-27 at 07 50 53](https://cloud.githubusercontent.com/assets/1609292/22362767/6c5652ea-e465-11e6-80a3-6e36ebb3575c.png)

But for the district layer, the processing tool 'Check validity' says everything is ok 🙁
But the problem is when you use the district layer in inasafe, QGIS can't compute the union.

Gustry:
Or should I try to make it valid on runtime ? @timlinux

Charlotte-Morgan:
hi @Gustry - this is a ticket to @samnawi to fix the data at this end, nothing for InaSAFE I don't think. Just letting you know we have a problem in the data and that I have not been able to fix it.
QGIS thinks it is invalid but Arc is happy when i run checks and fix. Even when I save and get back into QGIS its still not happy.

Gustry:
![screen shot 2017-01-27 at 11 11 03](https://cloud.githubusercontent.com/assets/1609292/22367381/8f01254c-e481-11e6-89ee-4089ee3dc412.png)

Charlotte-Morgan:
Yes - I'm sure he will fix it - or talk to the custodians

timlinux:
@Charlotte-Morgan these data have been broken since I can remember - I always left it like that because its an imperfect world and working with a bit of ropey data is a better reflection of how our users are probably working. I think the onus should be on us to either report these issues to the user, or deal with them seamlessly where we can....

Anyway happy to be overridden on this if you want to have clean data rather.

timlinux:
BTW when I investigated the data before it had interior rings that comprised only two vertices iirc. If you want to fix it I suggest using a tiny buffer and then a tiny inverse buffer - it will dissolve the interior rings.

samnawi:
OK @Charlotte-Morgan I'll try to looking up the problem and fix it.

samnawi:
Hi @Charlotte-Morgan I already fix the problem, and put it in the minion testing data

Gustry:
Hi @samnawi Where is the minion testing data ?
[...]

PR #3920, opened 2017-02-22
fix geometries on the fly for the aggregation layer

Gustry:
### What does it fix ?
* Ticket #3713 
* Funded by : DFAT
* Description : Fix geometries on the fly for the the aggregation layer

valid

### DIFF
--- a/safe/impact_function/create_extra_layers.py
+++ b/safe/impact_function/create_extra_layers.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+
 """Create extra layers in the impact function."""
 
 from qgis.core import (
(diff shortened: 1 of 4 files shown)


Gustry:
According to github, travis is still busy, but actually, travis is happy if we click on link. This PR is ready to be reviewed.

safe/impact_function/create_extra_layers.py
```diff
@@ -161,3 +162,31 @@ def create_profile_layer(profiling):
 
     tabular.commitChanges()
     return tabular
+
+
+def create_valid_aggregation(layer):
+    """Create a local copy of the aggregation layer and try to make it valid.
```
timlinux: Maybe good to add in the docstring why we need this (i.e. that geometries may need to be cleaned)

Gustry: I fixed this

Issue #3778, opened 2017-02-07
BUG: Type error


Charlotte-Morgan:
### problem ###
the following error occurred when running Ambon tsunami impact on population 

```
Problem

The following problem(s) were encountered whilst running the analysis.

TypeError
Suggestion

You can try the following to resolve the issue:

Check that you have the latest version of InaSAFE installed - you may have encountered a bug that is fixed in a subsequent release.
Check the InaSAFE documentation to see if you are trying to do something unsupported.
Report the problem using the issue tracker at https://github.com/inasafe/inasafe/issues. Reporting an issue requires that you first create a free account at http://github.com. When you report the issue, please copy and paste the complete contents of this panel into the issue to ensure the best possible chance of getting your issue resolved.
Try contacting one of the InaSAFE development team by sending an email to info@inasafe.org. Please ensure that you copy and paste the complete contents of this panel into the email.
Details
These additional details were reported when the problem occurred.

unsupported operand type(s) for +=: 'int' and 'NoneType'
Diagnostics

In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 981, in run self._run()
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1149, in _run self.summary_calculation()
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1611, in summary_calculation self._aggregate_hazard_impacted, self.aggregation)
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\gis\vector\summary_2_aggregation.py", line 112, in aggregation_summary exposure_class=key.replace(pattern, '')
In file "C:/Users/localadmin/.qgis2/python/plugins\inasafe\safe\utilities\pivot_table.py", line 46, in add_value self.data[key] += value
```

Gustry:
[...]

PR #3937, opened 2017-02-23
fill None values after zonal stats algorithm

Gustry:
### What does it fix ?
* Ticket #3778 
* Funded by : DFAT
* Description : After zonal stats, some values were set to None. We need to fill them with 0 or it breaks the summary calculation.


Gustry:
Feel free to merge if it's lgtm, I guess the DMI will try develop on monday morning.

PR #3938, opened 2017-02-23
add some default values into the hazard classifications

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : Improve the value mapping by adding some default values.

@ismailsunni Do we check these strings by making values in attribute value lowercase ? Do we also check if a translation is available ?


ismailsunni:
> Do we check these strings by making values in attribute value lowercase ? 
We do it by converting the value to uppercase

> Do we also check if a translation is available ?
Not yet. Should we do it?

Issue #3931, opened 2017-02-23
Note on displacement should only apply to pop exposure


Charlotte-Morgan:
### problem ###
Flood impact on building includes the following in notes:
For this analysis, the following displacement rates were used: High - 5.00%, Medium - 3.00%, Low - 1.00%, Use caution - 0.00%

### proposed solution ###
this note should only be used on pop exposure
a similar note stating thresholds for each zone would be useful

lucernae:
> this note should only be used on pop exposure

Ok. Noted, I haven't implement that one

Issue #3887, opened 2017-02-17
Missing some translations in exposure classification


Gustry:
### Problem

Exposure value mapping is translated in the wizard : 
![screen shot 2017-02-17 at 14 44 25](https://cloud.githubusercontent.com/assets/1609292/23067439/9dc9d462-f51f-11e6-8ae5-af322cfc7ac7.png)

* But then, in the dock, it's not translated anymore, for @ismailsunni : 
![screen shot 2017-02-17 at 14 43 08](https://cloud.githubusercontent.com/assets/1609292/23067458/b189cd0e-f51f-11e6-86db-e37b32264787.png)

* Not in the attribute table too. I'm not sure I will do it. We may have accents etc in French. I will check. 
![screen shot 2017-02-17 at 14 46 17](https://cloud.githubusercontent.com/assets/1609292/23067484/e0084412-f51f-11e6-9c48-6bb943d11afb.png)

* Not in the report too. @lucernae did you take values from the attribute table ? Can you check if there is a translation available ?
![screen shot 2017-02-17 at 14 36 13](https://cloud.githubusercontent.com/assets/1609292/23067502/f0d542e0-f51f-11e6-83fe-d50ceea8980d.png)

Gustry:
BTW, we should check that the value mapping and exposure breakdown is ordered following the exposure classification. For instance, 'Other' should always be at the bottom IMHO.  ( @lucernae  and @ismailsunni )

lucernae:
How do we sort it? Do we have something like definition for the order?

Gustry:
Yes we have ordering in definitions : 
https://github.com/inasafe/inasafe/blob/develop/safe/definitions/exposure_classifications.py#L43

Gustry:
Thanks @lucernae It's working for your side.

lucernae:
Nice!
Was translation working too?

Gustry:
@lucernae Yes it's working fine in the Analysis Details table : 
<img width="897" alt="screen shot 2017-03-06 at 22 24 03" src="https://cloud.githubusercontent.com/assets/1609292/23630539/cf0e6818-02bb-11e7-8e79-ffdab95ce9ee.png">

But translations are still missing on the aggregation table : 
<img width="906" alt="screen shot 2017-03-06 at 22 26 15" src="https://cloud.githubusercontent.com/assets/1609292/23630577/f53cb3be-02bb-11e7-907c-e5314d08c88c.png">

[...]

Issue #3876, opened 2017-02-16
Review Standard Report


lucernae:
### Problem

<img width="400" alt="screen shot 2017-02-16 at 15 32 59" src="https://cloud.githubusercontent.com/assets/831865/23013480/5eaf5422-f45d-11e6-9461-2d1dead349df.png">

Make this section obey standard.

### Solution

ongoing drafting solutions

ivanbusthomi:
hi @lucernae ,
according to the standard document, this is the changes that I proposed for this ticket. Feel free to add feedback or another suggestion
![image](https://cloud.githubusercontent.com/assets/7351977/23154821/48e82d60-f842-11e6-8243-142d406e1346.png)

Gustry:
@lucernae I still have the `hazard zone` for me (in French), did you remove it already ?

lucernae:
I haven't removed it.
The problem is, we got the label 'Hazard Zone' from hazard classification definition, and I'm not sure on the approach to fix this. We can:

1. Change the 'name' key-value to omit 'zone' in the hazard classification (the easiest). But perhaps it will conflict with other standards, such as in legends or in table attribute, we need to review.

2. Create a substitution rule in report_component metadata for all hazard classification, to customize the header. This way the changes only specific for the report standards. The example is like what I did in https://github.com/inasafe/inasafe/blob/develop/safe/definitions/reports/components.py#L113 to create a 'Not Affected' label, because we don't have definition based text for only 'Not Affected'

Both of them is trivial, I guess. Can @myarjunar take over from here, should be a good issue candidate for exercise 😄 ? We just need to decide which option is the best for the team.

PR #3939, opened 2017-02-23
Fix action notes

lucernae:
### What does it fix ?
* Ticket :
   - #3787 
   - #3931 
   - #3876 
   - #3887 
   - #3898 
* Funded by : DFAT
* Description : fixed report behaviour for displaced


lucernae:
I added extra messages:

If total displaced count 0:
for minimum needs
screen shot 2017-02-23 at 18 47 03

for aggregation postprocessors
screen shot 2017-02-23 at 18 57 30

lucernae:
changed column header to "Total displaced population" and add "male displaced count"
screen shot 2017-02-23 at 20 58 13

safe/definitions/minimum_needs.py
```diff
@@ -70,9 +68,9 @@ def _initializes_minimum_needs_fields():
             field_name = '{namespace}__{field_name}'.format(**format_args)
             # It's easier to debug if we have the real value from the
             # postprocessor. That's why I put a Double. ET.
-            field_type = QVariant.Double
+            field_type = qvariant_numbers
```
Gustry: So I guess you can remove my two comment lines just above, but it will not be true anymore.

lucernae: So, do you approve or not approve? 😄
If we want to make it consistent, we either make both displaced and minimum needs used either double or int.
My concern is just for the inconsistencies if the number added up to aggregation impacted.

lucernae:
Separate Analysis Question and General Report. Tidying up header (de-highlight hazard zone per @ivanbusthomi 's suggestions in #3876 ).
screen shot 2017-02-24 at 01 52 33

Make General Report split in larger screen per @ivanbusthomi 's suggestions.
screen shot 2017-02-24 at 01 52 12

Fix exposure class order and hopefully provides translation to class name (because it was taken from definitions)
screen shot 2017-02-24 at 01 52 44

safe/definitions/minimum_needs.py
Gustry: I though it might bring discrepancies if we lost some decimals. But I'm ok with integers.

Gustry: So it brings this kind of discrepancies now : 
screen shot 2017-02-27 at 15 08 04

We have now one transgender displaced.

Gustry: I think this one is also related now : 
screen shot 2017-02-27 at 15 49 57

I don't what we can do for these ones.

lucernae: Yes, @Gustry I think we should brought it up in the discussion. Like I said, it will bring number inconsistencies if it is summed, especially if the number is small.
The minimum needs and gender postprocessor is actually from a different cause.

- For the minimum needs it was caused by the change, because I made it as integers
- I suspect that it was always that case with the gender postprocessor (I didn't change that). It was one of the reason I propose to change displaced and postprocessors to decimals.

From this, we can either make all becomes integers and ignores small numbers discrepancies. Or all should be decimals, even if it doesn't make sense in the attribute table (like half people or half toilet 😄 ). CC @timlinux you might want to read this.

timlinux:
Definately -1 from me for half toilets :-P I think all minimum needs should be rounded up to the nearest whole unit at least.

PR #3942, opened 2017-02-23
additional customization for earthquake report

lucernae:
### What does it fix ?
* Ticket #3666 
* Funded by : DFAT
* Description : 
  - add fatalities and displaced rows in analysis results
  - add MMI Section to show exposed, displaced, and fatalities

### DIFF
--- a/safe/report/test/test_earthquake_report.py
+++ b/safe/report/test/test_earthquake_report.py
@@ -1,0 +1,87 @@
+# coding=utf-8
+import os
+
+import io
+
+import shutil
+import unittest
+
+from jinja2.environment import Template
+
+from safe.common.utilities import safe_dir
+from safe.definitions.constants import ANALYSIS_SUCCESS
+from safe.definitions.reports.components import \
+    standard_impact_report_metadata_html
+from safe.impact_function.impact_function import ImpactFunction
+from safe.report.impact_report import ImpactReport
+from safe.report.report_metadata import ReportMetadata
+from safe.test.utilities import (
+    get_qgis_app,
+    load_test_vector_layer, load_test_raster_layer)
+
+QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
+
+__copyright__ = "Copyright 2016, The InaSAFE Project"
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = ':%H$'
+
+
+class TestEarthquakeReport(unittest.TestCase):
+    """Test Earthquake Report."""
+
+    maxDiff = None
+
+    @classmethod
+    def fixtures_dir(cls, path):
+        directory_name = os.path.dirname(__file__)
+        return os.path.join(directory_name, 'fixtures', path)
+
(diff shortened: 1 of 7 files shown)


lucernae:
I included displaced and fatalities to Analysis result, in case we want to show it.

screen shot 2017-02-23 at 23 34 07

Also I prepared another section in case we want to show it:
For this one, I need mmi-related fields in analysis layer (for each mmi levels and each exposed/displaced/fatalities combinations).
screen shot 2017-02-23 at 23 34 47

Gustry:
> For this one, I need mmi-related fields in analysis layer (for each mmi levels and each exposed/displaced/fatalities combinations).

Just to clarify, we have now two EQ classifications for a hazard layer : 
* low, medium and high
* I, II, III, IV, V ...

I will always provide you this field in the analysis layer @lucernae : https://github.com/inasafe/inasafe/blob/develop/safe/definitions/fields.py#L1191
I think it's only what we will display. So the analysis summary report should be based on the EQ hazard classification which has been used for this layer. Sometimes it will be Low Medium and High and sometimes it will be MMI levels.

Please correct me @timlinux or @Charlotte-Morgan if I am wrong. I also need to be clear.

lucernae:
> So the analysis summary report should be based on the EQ hazard classification which has been used for this layer. Sometimes it will be Low Medium and High and sometimes it will be MMI levels.

I will note to myself:
- So, in one analysis report, it will be either one and not both.
- perhaps MMI needs to be displayed using Roman numerals.

Gustry:
We need to confirm this @lucernae. I'm not sure.

For Roman numerals, they will come automatically in the hazard_count_field.
screen shot 2017-02-23 at 20 39 45

safe/report/test/test_earthquake_report.py
```diff
@@ -0,0 +1,87 @@
...
+from safe.definitions.reports.components import \
+    standard_impact_report_metadata_html
+from safe.impact_function.impact_function import ImpactFunction
+from safe.report.impact_report import ImpactReport
+from safe.report.report_metadata import ReportMetadata
+from safe.test.utilities import (
+    get_qgis_app,
+    load_test_vector_layer, load_test_raster_layer)
+
+QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
```
Gustry: I know it's new, but I think we should start using safe.test.utilities.qgis_iface() instead of get_qgis_app() It will make our life easier when will switch to QGIS3. For next time ;-)

lucernae:
I will handle the merge conflict first

safe/report/test/test_earthquake_report.py
lucernae: Nice! I will change it. The test itself was not written yet. Not sure what to test at the moment.

lucernae: I try to use it, but I got this instead:

```
ImportError: No module named mock
```

Will not use it for now.

lucernae:
Just a comment, the proposed MMI Detail section will only be showed in Debug Mode for now.

Gustry:
Thanks @lucernae !

PR #3948, opened 2017-02-27
remove strings concatenation, they are missing in transifex

Gustry:
### What does it fix ?
* Ticket #3678 
* Funded by : World Bank
* Description : Remove string concatenation, these strings are missing on transifex.


lucernae:
LGTM

PR #3949, opened 2017-02-27
add displacement ratio to the attribute table

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : add displacement ratio to the attribute table

screen shot 2017-02-27 at 14 40 18

### DIFF
--- a/safe/definitions/fields.py
+++ b/safe/definitions/fields.py
@@ -817,6 +817,35 @@ affected_field = {
     'replace_null': False
 }
 
+# Displacement ratio
+population_displacement_ratio_field = {
+    'key': 'population_displacement_ratio_field',
+    'name': tr('Population Displacement Ratio'),
+    'field_name': 'population_displacement_ratio',
+    'type': QVariant.Double,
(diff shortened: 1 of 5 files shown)


timlinux:
I think this is one case where showing 3 decimal places would be good.

timlinux:
I wonder also if it would be more explicit to call the column "population_displacement_rate" so that in the future we can allow for e.g. "aged_displacement_rate" and other such vulnerability indicators....

Gustry:
> I think this is one case where showing 3 decimal places would be good.

It's showing only 0 or 1 because the displacement ratio is only 0 and 1 ;-) The field is already with 3 digits precision and it's working fine with a displacement ratio different than 0 and 1.

> I wonder also if it would be more explicit to call the column "population_displacement_rate"

Ok, I'm going to rename it.

Gustry:
> I wonder also if it would be more explicit to call the column "population_displacement_rate"

Ok done, I put : `population_displacement_ratio`, we said rate is for banking ;-) Is-it ok like this ?

PR #3950, opened 2017-02-27
enable test about minimum needs with a raster exposure

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : enable test about minimum needs with a raster exposure


lucernae:
We can close this, I have enabled the test again.

Issue #3742, opened 2017-01-31
We need to report on hectares, not on square meters


Gustry:
### Problem
![screen shot 2017-01-31 at 10 02 41](https://cloud.githubusercontent.com/assets/1609292/22458361/743e4366-e79c-11e6-80b4-3123d3dec7a7.png)

We do an analysis on landcover, we need to report with hectares, not, square meters.

timlinux:
@Gustry  and @Charlotte-Morgan should we not do the same for roads (km vs m when the length exceeds 1000m?)

Charlotte-Morgan:
sounds like a good idea - do I need to update the standards?

Gustry:
Just to be sure, in a legend, is-it ok if we have two units (km and m for instance) ?
* Length of roads
  * High hazard zone (900m)
  * Medium Hazard zone (1.2km)
  * Low Hazard Zone (120m)

Or should we check if every values can be converted first ? (all to km if it's possible).

I think we can't do that for reports as units are in headings. I'm talking about the legend.

Charlotte-Morgan:
hi @Gustry  - I think it would be better to use only one unit in the legend.  In the example above I would use meters- 1200m is better than 0.9 km if you are to be consistent.

Gustry:
Ok so, in the legend, only if all values are greater than 1000 we switch to kilometres if I understood.

Charlotte-Morgan:
I think its more about the range of values
if you have 20m in one class and 20,000m in another I would use metres for all.

However; if you get 1500m and 986000m then it would be better to use km ie 1.5 and 986Km

Gustry:
Yes, so my rule works ;-)

Gustry:
These 2 PR have been merged. I'm going to close this issue.

PR #3951, opened 2017-02-27
convert unit, for landcover from square metres to hectare for instance

Gustry:
### What does it fix ?
* Ticket #3742 
* Funded by : DFAT
* Description : Convert unit in size postprocessor if needed

@lucernae I noticed your were already displaying hectare. But values were never in hectare yet ;-)
We should read definitions.

### DIFF
--- a/safe/definitions/units.py
+++ b/safe/definitions/units.py
@@ -154,6 +154,24 @@ unit_metres = {
         }
     ],
 }
+unit_hectares = {
+    'key': 'hectare',
+    'name': tr('Hectare'),
+    'base_unit': False,
+    'plural_name': tr('hectares'),
+    'measure': tr('Area'),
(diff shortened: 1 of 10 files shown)


safe/definitions/units.py
```diff
@@ -154,6 +154,22 @@
...
 }
+unit_hectare = {
+    'key': 'hectare',
+    'name': tr('Hectare'),
+    'base_unit': False,
+    'plural_name': tr('hectares'),
+    'measure': tr('Area'),
+    'abbreviation': 'ha',
+    'description': tr(
+        '<b>Hectare</b> are a metric unit of measure.'),
```
ismailsunni: The description is almost the same as square metres. May be you can get the description from https://en.wikipedia.org/wiki/Hectare :) And add the citation from it also

PR #3965, opened 2017-02-28
fix default string in dock when we add new layers

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : Fix default string


Gustry:
This is just a bug fix, I'm going to merge this one

Issue #3956, opened 2017-02-28
Batch runner UI


myarjunar:
### Problem

https://files.gitter.im/inasafe/inasafe/YriY/screencapture.gif

### Expected Behaviour
good looking batch runner

### Environment
* InaSAFE : 4.0 develop
* QGIS : 2.14
* OS : Windows

myarjunar:
closed in #3966

Issue #3955, opened 2017-02-28
Fail to run batch runner scenario from save scenario tool


myarjunar:
### Problem
cannot find path generated from save scenario

### Expected Behaviour
path generated from save scenario tool is recognized by batch runner

### Environment
* InaSAFE : 4.0 develop
* QGIS : 2.14
* OS : Windows
* Data used (filenames and paths so as to reproduce, a project is better):
```
[Flood new scenario] 
exposure = E:\Sync\data.inasafe.org\InaSAFE 3.5 Training Data\DKI Jakarta\DKI Jakarta\02 Exposure Data\Jakarta_roads_WGS84.shp 
hazard = E:\gisdata\Jakarta\JakartaFloods-2April2016\flood_data.json 
extent = 106.803668, -6.325429, 106.886219, -6.237527 
extent_crs = EPSG:4326 
aggregation = E:\Sync\data.inasafe.org\InaSAFE 3.5 Training Data\DKI Jakarta\DKI Jakarta\03 Aggregation Data\Jakarta_Subdistrict_Boundary_WGS84.shp
```

myarjunar:
@Gustry @timlinux do we support json format for layer?

timlinux:
Yes sure - any valid QGIS supported format should be able to work....

myarjunar:
closed in #3966

PR #3966, opened 2017-02-28
handle json file on batch runner + ui polish

myarjunar:
### What does it fix ?
* Ticket #3955 #3956 
* Funded by : DFAT
* Description : 

batch runner_030

### DIFF
--- a/safe/gui/tools/batch/batch_dialog.py
+++ b/safe/gui/tools/batch/batch_dialog.py
@@ -402,7 +402,7 @@ class BatchDialog(QDialog, FORM_CLASS):
             else:
                 LOGGER.critical('Failed to create layer from scenario')
                 return
-        elif extension in ['.shp', '.geojson', '.gpkg']:
+        elif extension in ['.shp', '.json', '.geojson', '.gpkg']:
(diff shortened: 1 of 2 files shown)


safe/gui/tools/batch/batch_dialog.py
```diff
@@ -402,7 +402,7 @@ def define_layer(self, layer_path):
             else:
                 LOGGER.critical('Failed to create layer from scenario')
                 return
-        elif extension in ['.shp', '.geojson', '.gpkg']:
+        elif extension in ['.shp', '.json', '.geojson', '.gpkg']:
```
timlinux: ok this is just a bandaid fix for now ... we really should support all QGIS supported formats. We could try to load each layer first as a raster, then as a vector then fail.

Gustry: Yes +1 ;-)
It would be much better.
https://github.com/inasafe/inasafe/blob/develop/safe/datastore/datastore.py#L153

Gustry:
Great, that's better, we support every format now!

PR #3969, opened 2017-02-28
use kilometres if possible in the legend

Gustry:
### What does it fix ?
* Ticket #3742 
* Funded by : DFAT
* Description : use kilometres if possible in the legend

I think we need to do the same in reporting. (cc @lucernae @myarjunar )

### DIFF
--- a/safe/definitions/exposure.py
+++ b/safe/definitions/exposure.py
@@ -8,6 +8,7 @@ from safe.definitions.units import (
     unit_metres,
     unit_square_metres,
     unit_hectares,
+    unit_kilometres,
 )
 from safe.definitions.fields import (
     adult_count_field,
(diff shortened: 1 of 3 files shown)


safe/definitions/exposure.py
```diff
@@ -229,7 +230,7 @@
         'line'
     ],
     'size_unit': unit_metres,
-    'units': [unit_metres],
+    'units': [unit_metres, unit_kilometres],
```
lucernae: How about, the most preferred unit in the front?
what do you think, @Gustry ?

Gustry: Yes I can do that

Gustry: @lucernae Yes i can undertsand. I was doing by increasing order, it's also natural for me.
But currently, I have a problem if we have more than 2 units because I my code is doing only one conversion. It's low priority to have more than 2 units, I have some other blockers for now. Is-it ok like this ?

Issue #3967, opened 2017-02-28
[Bug] Cannot change Minimum Needs Profile


adissadis:
### Problem
When I ran InaSAFE analysis for flood raster impact on population raster, i cannot change minimum needs profile into different minimum needs profile (for example Sphere or Philippines).

Here's my video about the bug:
![gifrecord_2017-02-28_183708](https://cloud.githubusercontent.com/assets/2520159/23404198/534d656e-fde6-11e6-942a-b37df9bd894c.gif)

If i want to change the minimum needs profile, i need to restart my QGIS first. 

### Expected Behaviour
When the user change minimum needs profile, user can use the minimum needs profile directly to the new analysis. Without need to restart QGIS first.

### Environment
* InaSAFE : 4.0.0b
* QGIS : 2.14.12
* OS : Xubuntu 16.04
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : Java Population Raster
  * Hazard : Flood HKV Raster
  
Checklist (if applicable): 
* [V ] Did you try to run it in debug mode ?
* [ V] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? Data.inasafe.org

### Proposed Solution
If possible, could you give prompt or information window when the user select another minimum needs profile

timlinux:
Hi @adissadis I think it is written in the help that you need to restart QGIS after switching needs profiles. Perhaps @myarjunar can find away to remove the need for the restart....

Gustry:
We need to check why we need a restart.
Can't we make a manual refresh before we launch the IF ?

I agree, we should at least display a message. Not a lot of people read helps, especially when it's in another window.

When I give a training, it's also not obvious for users that the last profile selected in the combobox will be the default one. It could be nice if we can review this.

adissadis:
Hi @timlinux i'm sorry i didn't read the Help about restart QGIS when we want to change minimum needs. Because in previous InaSAFE version (i forget what version), you only need to click "Save" button to change minimum needs profile.

timlinux:
Hi @adissadis no I think its always been a requirement since it was first added. Lets see if @myarjunar can remove the need for doing that for us....

timlinux:
See #3976 - maybe we can take another look for 4.3

PR #3976, opened 2017-03-01
Direct effect after changing needs profile

myarjunar:
### What does it fix ?
* Ticket #3967 
* Funded by : DFAT

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -60,7 +60,6 @@ from safe.gis.raster.polygonize import polygonize
 from safe.gis.raster.zonal_statistics import zonal_stats
 from safe.gis.raster.align import align_rasters
 from safe.gis.raster.rasterize import rasterize_vector_layer
-from safe.definitions.post_processors import post_processors
 from safe.definitions.analysis_steps import analysis_steps
 from safe.definitions.utilities import definition
 from safe.definitions.exposure import indivisible_exposure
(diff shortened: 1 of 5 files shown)


safe/impact_function/impact_function.py
```diff
@@ -1766,6 +1765,7 @@ def post_process(self, layer):
         # Set the layer title
         layer_title(layer)
 
+        from safe.definitions.post_processors import post_processors
```
Gustry: So we can't import post_processors at the beginning of file as usual ? If yes, you should put a comment why you want to import not at the beginning.  Same comment where you import not at the beginning.

myarjunar: > So we can't import post_processors at the beginning of file as usual ?

I think yes if we want to have a direct effect after changing needs profile, `post_processors` value has to be changed to apply new minimum needs post processors. I'll write a comment for it.

myarjunar:
layer properties - number of people fields_032
layer properties - analysis_impacted fields_031

Need your help, I'm getting stuck here :sweat_smile: @Gustry  . If you look at those pictures, after changing needs profile then run the IF, on `analysis_impacted` layer I didn't get the same minimum needs as `aggregation_impacted` layer. What probably caused this?

lucernae:
I'm taking this with @myarjunar .

Gustry:
ok great, I was going to check. I will let you

ismailsunni:
1 month later.
How is it going @lucernae @myarjunar ?

Gustry:
Still in progress, we have an issue with these missing fields. I will try to fetch this code

Issue #3971, opened 2017-03-01
InaSAFE analysis results are incorrect


Charlotte-Morgan:
### problem ###
In a review of the analysis behavour for population exposure @samnawi has identified some incorrect analysis results.

The testing reviewed the analysis of population counts in exposure data and the application of ratios in aggregation data using both vector and raster population data.

Example 1.
In vector flood and raster population with vector aggregation - the global default age and gender ratios are applied regardless of whether the Do not use; Global default or Custom ratios are set in the aggregation key words.

Example 2.
In vector flood and vector population with vector aggregation; the results are inconsistent. The results are correct when no field is selected in exposure data and global defaults are used in aggregation layer.
The results are incorrect (approximately 10% of expected value) when fields are selected in the exposure data.
The results are incorrect when no field is selected in the exposure data and a ratio is selected from the aggregation layer

### proposed solution ###
example 1.
the radio buttons in the aggregation data should be applied in preference to global defaults if set.

example 2.
the results produced by different analysis methods should produce similar results.
The results should be 'correct' ish.

... more to follow ... the situation with vector population data is more complicated

samnawi:
I try to investigate the behavior of InaSAFE population analysis to calculate age & gender.
I the table below, show the result of the analysis that I doing today.

Hazard : Jakarta Flood (Peta Bencana.id)
Population : Dummy Vector Population
Aggregation : Dummy Aggregation

![image](https://cloud.githubusercontent.com/assets/8871958/23451209/8939923c-fe91-11e6-9950-c69054e7b56f.png)

Hazard : Jakarta Flood (Peta Bencana.id)
Population : Raster Population (InaRISK Data)
Aggregation : Dummy Aggregation

![image](https://cloud.githubusercontent.com/assets/8871958/23451229/99f7936c-fe91-11e6-887c-1520ce97ccc1.png)

According to the result in above, I try to create a conclusion of the InaSAFE Behaviour to calculate age and gender ratio (**Just focus on the red color**)
![image](https://cloud.githubusercontent.com/assets/8871958/23451285/d9abb9fc-fe91-11e6-876c-df5d1fdaa802.png)


And this is the attribute table of the population and aggregation
**Population**
![image](https://cloud.githubusercontent.com/assets/8871958/23451040/b876d52e-fe90-11e6-93fe-8b177bef2bd5.png)

**Aggregation**
[...]

PR #3977, opened 2017-03-01
add many logs in the IF

Gustry:
### What does it fix ?
* Ticket #3971 
* Funded by : DFAT
* Description : 

It adds many logs which are very useful to see what happen with ratios and counts : 
```
2017-03-01T17:34:11	0	aggregation: Aggregation layer already in exposure CRS
2017-03-01T17:34:11	0	female_ratio_field field detected in the aggregation layer AND the equivalent count has not been detected in the exposure layer. We will propagate this ratio to the exposure layer later when we combine these two layers.
2017-03-01T17:34:11	0	youth_ratio_field = 0.25 constant detected in the aggregation layer AND the equivalent count has not been detected in the exposure layer. We are adding this ratio to the exposure layer.
2017-03-01T17:34:11	0	adult_ratio_field is neither a constant nor a field AND the equivalent count has not been detected in the exposure layer. We will do nothing about it.
2017-03-01T17:34:11	0	elderly_ratio_field = 0.2 constant detected in the aggregation layer AND the equivalent count has not been detected in the exposure layer. We are adding this ratio to the exposure layer.
2017-03-01T17:34:11	0	aggregation: Convert the aggregation layer to the analysis layer
```

### DIFF
--- a/safe/gis/vector/prepare_vector_layer.py
+++ b/safe/gis/vector/prepare_vector_layer.py
@@ -112,6 +112,9 @@ def prepare_vector_layer(layer, callback=None):
     rename_remove_inasafe_fields(cleaned)
 
     if _size_is_needed(cleaned):
+        LOGGER.info(
+            'We noticed some counts in your exposure layer. Before to update '
+            'geometries, we compute the original size for each feature.')
         run_single_post_processor(cleaned, post_processor_size)
 
     if cleaned.keywords['layer_purpose'] == 'exposure':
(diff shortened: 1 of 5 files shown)


safe/gis/vector/prepare_vector_layer.py
```diff
@@ -234,7 +237,8 @@ def rename_remove_inasafe_fields(layer):
     # Remove useless fields
```
timlinux: Just a note for future, I noticed  the way you generally use the word 'useless' - it is better phrased 'unnecessary'. Useless is often derogatory 'good for nothing' versus unnecessary meaning 'not needed'....

timlinux: PS Dont do anything for now, just file in your vocab for future....:-P

PR #3978, opened 2017-03-02
handle inasafe default wizard behaviour if layer has keywords

myarjunar:
### What does it fix ?
* Ticket #3813 
* Funded by : DFAT
* Description : handle inasafe default wizard behaviour if layer has keywords


Gustry:
Thanks @myarjunar

PR #3983, opened 2017-03-02
Ratios and counts

Gustry:
### What does it fix ?
* Ticket #3971 
* Funded by : DFAT
* Description : Managa ratios when we use an aggregation layer on a raster layer


Gustry:
With this kind of logs, we can just "read" logs in QGIS and understand how we compute ratios and counts along the process. Because it's quite tricky all these uses cases. Having this kind of logs would have been much appreciated for @samnawi's work.

I would love having logs in the datastore, I don't think I will have time for this.

timlinux:
Sure no problem @Gustry we can do it some other release....

Issue #3973, opened 2017-03-01
Male displaced count show the negative results


felix-yew:
### Problem
I run the flood vector vs population vector and find a negative number of male displaced count in the detailed gender report.
![image](https://cloud.githubusercontent.com/assets/13621886/23448593/822676b6-fe84-11e6-8b95-152e28e4737c.png)

### Expected Behaviour
Show the right total displace for male.

### Environment
* InaSAFE : v.4.0.0b0
* QGIS : 2.14.10
* OS : Win 7
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : [DKI_Jakarta_Population_Dukcapil_2013_WGS84.zip](https://github.com/inasafe/inasafe/files/809869/DKI_Jakarta_Population_Dukcapil_2013_WGS84.zip)
  * Hazard : [Jakarta_Flood_18113_WGS84.zip](https://github.com/inasafe/inasafe/files/809872/Jakarta_Flood_18113_WGS84.zip)
  * Aggregation : -
  * Extent : 
  
Checklist (if applicable): 
* [x] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? 

### Proposed Solution
Make it happy :smile: 

Cc: @Charlotte-Morgan @timlinux @Gustry @ismailsunni @lucernae @samnawi @ivanbusthomi

Gustry:
@felix-yew Thanks for reporting this ticket and sharing your data.
But none of your zip (exposure and hazard) have a shapefile in (.shp).
I will try another dataset.

felix-yew:
woops :pray:  sorry, I upload again
* Exposure: [DKI_Jakarta_Population_Dukcapil_2013_WGS84.zip](https://github.com/inasafe/inasafe/files/810297/DKI_Jakarta_Population_Dukcapil_2013_WGS84.zip)
* Hazard: [Jakarta_Flood_18113_WGS84.zip](https://github.com/inasafe/inasafe/files/810303/Jakarta_Flood_18113_WGS84.zip)
or you can check it in Real World Data in Flood folder.

let me know if there is some mistake in my zip file, again.

Gustry:
I used the data from RealWorldData : 
I run the keywords wizard and everything runs fine for me : 
[...]

Issue #3972, opened 2017-03-01
Python error: exposure layer combobox in the dock is empty


Charlotte-Morgan:
### problem ###
analysis fails to complete

### log message ###
```
2017-03-01T13:23:31	0	Sentry logging enabled in safe
2017-03-01T13:24:28	0	Datastore : C:/AIFDR/AIFDR II - SDA/temp/FloodPolygonOnPopulationPolygon_01March2017_13h24-28.855000
2017-03-01T13:24:28	0	aggregation: provided: False
2017-03-01T13:24:29	0	hazard: use_same_projection: True
2017-03-01T13:24:29	0	Features which have been removed from hazard : 0
2017-03-01T13:24:29	0	Fields which have been renamed from hazard : {u'state': 'hazard_value'}
2017-03-01T13:24:29	0	Fields which have been removed from hazard : [u'state', u'area_id', u'geom_id', u'area_name', u'parent_nam', u'city_name', u'state', u'last_updat', u'floodclass', u'flooded']
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
2017-03-01T13:24:29	0	GEOS geoprocessing error: One or more input features have invalid geometry.
[...]

PR #3984, opened 2017-03-02
Fix 3973 fix negative number

ismailsunni:
### What does it fix ?
* Ticket #3973 
* Funded by : DFAT
* Description : 
   All exposure for volcanic ash are enabled now.
screen shot 2017-03-01 at 07 54 54


safe/gis/vector/from_counts_to_ratios.py
```diff
@@ -69,19 +76,17 @@ def from_counts_to_ratios(layer, callback=None):
                     count_field=count_field['key']))
 
     if len(mapping) == 0:
-        # There is not a count field. Let's skip this layer.
+        # There is not a subset count field. Let's skip this layer.
         layer.commitChanges()
         return layer
```
Gustry: Can you add an information debug message about what we are doing here, with the textual formula?

Gustry:
Great thanks. I put one small comment.
So we will have to revamp the size when we will use feature value? To divide with the size, right ?

Gustry:
I guess you are not working on ticket #3972 as mentioned in this PR. It's #3984

felix-yew:
Hi mas @ismailsunni I try your PR for #3973 
image
:+1: LGTM, thank you

samnawi:
Yeah.... It's work 
image

It's also fix the wrong calculation in the count ratio @ismailsunni ?

safe/gis/vector/from_counts_to_ratios.py
ismailsunni: Done.

ismailsunni:
> So we will have to revamp the size when we will use feature value? To divide with the size, right ?

Yes. But I think you have already done that. I check in the `test_ratios_with_vector_exposure`. Divided exposure polygon has the count proportional to the size / area. Not sure about feature value, though.

> I guess you are not working on ticket #3972 as mentioned in this PR. It's #3984

ah yes, typos :D 

Thanks @samnawi @felix-yew I will merge this after Travis is passing.

Issue #3860, opened 2017-02-14
Deprecated API calls being used


timlinux:
### Problem
Lots of warnings are showing when running Flood on buildings using flood classes from #3667 :

```
2017-02-14T16:42:20	1	warning:C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\site-packages\PyQt4\uic\uiparser.py:844: PendingDeprecationWarning: This method will be removed in future versions.  Use 'elem.iter()' or 'list(elem.iter())' instead.
			  for include in elem.getiterator("include"):
			
			traceback:  File "<string>", line 1, in <module>
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 315, in startPlugin
			    plugins[packageName].initGui()
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\plugin.py", line 524, in initGui
			    self._create_dock()
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\plugin.py", line 500, in _create_dock
			    from safe.gui.widgets.dock import Dock
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 572, in _import
			    mod = _builtin_import(name, globals, locals, fromlist, level)
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\gui\widgets\dock.py", line 69, in <module>
			    from safe.impact_function.impact_function import ImpactFunction
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 572, in _import
			    mod = _builtin_import(name, globals, locals, fromlist, level)
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 55, in <module>
			    from safe.gis.raster.clip_bounding_box import clip_by_extent
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 572, in _import
			    mod = _builtin_import(name, globals, locals, fromlist, level)
			  File "C:/Users/Tim/.qgis2/python/plugins\inasafe\safe\gis\raster\clip_bounding_box.py", line 6, in <module>
			    import processing
			  File "C:/PROGRA~1/QGIS2~1.14/apps/qgis-ltr/./python\qgis\utils.py", line 572, in _import
[...]

PR #3987, opened 2017-03-02
remove deprecated functions from the QGIS api

Gustry:
### What does it fix ?
* Ticket #3860
* Funded by : DFAT
* Description : remove deprecated functions from the QGIS API


safe/report/processors/default.py
```diff
@@ -242,7 +242,7 @@ def qgis_composer_renderer(impact_report, component):
         """:type: qgis.core.QgsComposerPicture"""
         if image is not None and path is not None:
             try:
-                image.setPictureFile(path)
+                image.setPicturePath(path)
```
lucernae: Thanks for doing this too!

Issue #3803, opened 2017-02-09
KeyError in Earthquake Raster vs Population Raster


ivanbusthomi:
### Problem
I got this error when running earthquake raster on population raster.
```
Problem
------------
The following problem(s) were encountered whilst running the analysis.
KeyError

Details
------------
These additional details were reported when the problem occurred.
classification

Diagnostics
------------
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1021, in run self.style()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1678, in style self.debug_mode)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\style.py", line 117, in generate_classified_legend hazard_classification = hazard.keywords['classification']
```

### Environment
* InaSAFE : 4.0 develop
* QGIS : 2.14.10
* OS : Windows
* Data used : West Sumatra dataset
  * Exposure : Population Raster
  * Hazard : Earthquake raster
  * Aggregation : villages

### Proposed Solution
Make it Happy
___

CC @ismailsunni @timlinux @Charlotte-Morgan @samnawi @felix-yew

samnawi:
Hi @ivanbusthomi I also found the "Key Error" problem in Flood raster VS Population Raster and Tsunami raster VS population raster.

Hi @Gustry Should I open new ticket for it, or it's still relevant with this ticket?

Gustry:
Can you open a new one @samnawi ?

EQ IF is a special one so your error might be different. Thanks

Gustry:
This has been fixed. I'm going to close this one.

Charlotte-Morgan:
hi @Gustry - i got key error for earthquake raster on pop raster with aggregation boundaries this morning

Gustry:
Is-it the same KeyError ?
Can you copy/paste logs ?

ivanbusthomi:
Hi @Gustry , actually I tried using west sumatra dataset in Real world data and it works. But when I tried using another data set, I got a key error. these are the diagnostics
```
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1067, in run self._run()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
[...]

PR #3991, opened 2017-03-02
Key error eq

Gustry:
### What does it fix ?
* Ticket #3803 
* Funded by : DFAT
* Description : 
  * Fix a unicode problem in French
  * Fix some small inner rings in the analysis extent


lucernae:
LGTM

Issue #3934, opened 2017-02-23
Infographics behaviour


lucernae:
### Problem

Right now we had this kind of infographics:

<img width="1066" alt="screen shot 2017-02-23 at 16 18 19" src="https://cloud.githubusercontent.com/assets/831865/23252542/c7cfaafc-f9e3-11e6-86c7-26fe44c43287.png">

We had some behaviour that needs to be clarified:
Affected or displaced?
- Pie chart is showing total population affected for each hazard zone (perhaps change it to displaced?)
- Number of people in People section is total affected population (perhaps change it to displaced?)
- Vulnerability showed "from x people affected" (perhaps change it to displaced?)

Is the rounding behaviour as expected?
- All numbers in hazard zones in pie chart are rounded. But since pie chart needs to be total 100%. Total numbers in center of pie chart will be different than the number in People section.
- All the percentage in Vulnerabilities section might be different from the actual ratio used to calculate them, because of rounding.

Icons?
- What to show if user chooses minimum needs profile other than default BNPB_en. Some icons in minimum needs will be missing.

felix-yew:
Hi, mas @lucernae - maybe I can add some suggestion about the:
> Affected or displaced?
> * Pie chart is showing total population affected for each hazard zone (perhaps change it to displaced?)

Maybe we still show the total population in pie chart (Affected + Not Affected + Not Exposed)

> * Number of people in People section is total affected population (perhaps change it to displaced?)

In the people section we can add the new icon and information for displaced, maybe we can add in the right side of affected icon. I think the number of displaced is the key to the vulnerability and minimum needs section (cmiiw).

![image](https://cloud.githubusercontent.com/assets/13621886/23257431/59969eb6-f9f6-11e6-926d-7fe7b89fee6a.png)

> * Vulnerability showed "from x people affected" (perhaps change it to displaced?)

After we have the number of displaced, so the vulnerability section will show "from x people displaced". 
![image](https://cloud.githubusercontent.com/assets/13621886/23257501/b89b0ffa-f9f6-11e6-9413-81249d012bb0.png)

> Is the rounding behaviour as expected?
> All numbers in hazard zones in pie chart are rounded. But since pie chart needs to be total 100%. Total numbers in center of pie chart will be different than the number in People section.

I think this number will be different because in people section we will take the number of affected and displaced. But, in total numbers in the center of pie chart will be the total of "Affected + Not Affected + Not Exposed"

[...]

PR #3992, opened 2017-03-02
fix issue #3934: tidying up infographics:

lucernae:
### What does it fix ?
* Ticket #3934
* Funded by : DFAT

### Description : 
- Don't tag infographic pdf as final product until
  the problem with png generation in composer is resolved
  (html is still working fine)
- Split vulnerabilities into 2 group
- Report displaced number
- Add notes from where displaced number comes from
- Hide slice label if the slice is less than 10%


lucernae:
Mas @felix-yew I still can't make that triangle in Gender group and Age group 😄 .
I just make a border for now.
I put the from {number} displaced side by side with Vulnerability, since it seems make sense
I modified the displacement notes, since the rate might be different

screen shot 2017-03-03 at 06 05 29

timlinux:
@lucernae there is still a problem here:

inked0764ea70-ffd8-11e6-878e-8f1aaa7fb792_li

lucernae:
@timlinux Can you hand me the icons :smile:

felix-yew:
mas @lucernae I try your PR and the result like this
image
there are only displaced icon? I think in the #3934 we also show the affected and then the displaced in the people section.

> I just make a border for now.
I put the from {number} displaced side by side with Vulnerability, since it seems make sense

LGTM :+1: 


> I modified the displacement notes, since the rate might be different

Can you make like a wrap text and make the font little smaller (if it still can read)


This is the icon for displaced, you can download from this [link](https://thenounproject.com/ochaavmu/collection/ocha-humanitarian-icons/?i=4244)

lucernae:
New preview
screen shot 2017-03-03 at 16 59 36

lucernae:
Using the new icon
screen shot 2017-03-03 at 17 56 43

PR #3995, opened 2017-03-03
anticipate inactive exposure layer

myarjunar:
### What does it fix ?
* Ticket #3972 
* Funded by : DFAT
* Description : Make sure current selected exposure layer is used on current IF.


Gustry:
LGTM thanks

Issue #3947, opened 2017-02-27
Landcover Classification


ivanbusthomi:
### Problem
We need to make the classification for landcover exposure match with the volcanic ash landcover classification. The landcover classification came from Badan Geologi.
![lc](https://cloud.githubusercontent.com/assets/178003/14305277/e0fd1d78-fbe4-11e5-80ae-b92335877840.png)
We might also need to automatically assign the value map from BIG landcover data to the classes because the value in BIG land cover data has already been standardised using either the Remark Field or the FCODE field.
![image](https://cloud.githubusercontent.com/assets/7351977/23359852/d96dee4c-fd1a-11e6-9a3b-60a7a6a998e8.png)

### Expected Behaviour
the landcover classification matches the Badan Geologi classification.

### Environment
* InaSAFE : 
* QGIS : 
* OS : 
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : 
  * Hazard : 
  * Aggregation :
  * Extent : 

### Proposed Solution
Maybe by adding new land cover classification. This is the volcanic ash landcover classification and the value map for BIG land cover data

| Value Map      | Remark | FCODE
| -------------    | ------------- | ------------- |
| Settlement      | Permukiman dan Tempat Kegiatan | 50102 |
| Rice Field        | Sawah | 50306 |
| Plantation       | Perkebunan / Kebun  | 50304 |
| Water Supply  | Air Danau / Situ, Air Empang, Air Penggaraman, Air Tambak, Air Tawar Sungai, Air Waduk, Perairan Lainnya | 50404, 50420, 50418, 50416, 50408, 50406, 50400|
| Forest              | Hutan Rimba  | 50202 |

Gustry:
I put the label release blocker because I though it was. If it's not, you can remove it. Should we go with this classification ? Should we remove the old one ?

ivanbusthomi:
Hi @Gustry,
I am thinking maybe we can just add another classification for BIG landcover, and make it automatically value mapped if using the remark or FCODE field based on the table above.
And I believe it is not a release blocker. What do you think @Charlotte-Morgan ?

Gustry:
Yes, good.
Do you want to do it @ivanbusthomi ?
You need to add new classification here : https://github.com/inasafe/inasafe/blob/develop/safe/definitions/exposure_classifications.py
(CC @myarjunar )

Gustry:
Check #3626

myarjunar:
[...]

PR #3999, opened 2017-03-03
add landcover definition from badan geologi for volcanic ash

myarjunar:
### What does it fix ?
* Ticket #3947 
* Funded by : DFAT

### DIFF
--- a/safe/definitions/exposure_classifications.py
+++ b/safe/definitions/exposure_classifications.py
@@ -674,10 +674,10 @@ generic_landcover_classes = {
 }
 
 badan_geologi_landcover_classes = {
-    'key': 'volcanic_ash_landcover_classes',
-    'name': tr('Volcanic Ash Landcover Classification'),
+    'key': 'badan_geologi_landcover_classes',
+    'name': tr('Badan Geologi Landcover Classification'),
     'description': tr(
-        'Classification of landcover for volcanic ash based on Badan Geologi'),
+        'Classification of landcover based on Badan Geologi'),
     'type': exposure_classification_type,
     'citations': [
         {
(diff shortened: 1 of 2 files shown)


safe/definitions/exposure_classifications.py
```diff
@@ -673,6 +673,101 @@
     ]
 }
 
+badan_geologi_landcover_classes = {
+    'key': 'volcanic_ash_landcover_classes',
+    'name': tr('Volcanic Ash Landcover Classification'),
```
Gustry: @myarjunar and @ivanbusthomi Is-it only specific for volcanic ash ? I guess no, we can use this landuse classification with tsunami,... no ?

myarjunar: I don't know either, I quote from here:
> We need to make the classification for landcover exposure match with the volcanic ash landcover classification.

CMIIW @ivanbusthomi

safe/definitions/exposure_classifications.py
```diff
@@ -673,6 +673,101 @@
     ]
 }
 
+badan_geologi_landcover_classes = {
+    'key': 'badan_geologi_landcover_classes',
+    'name': tr('Badan Geologi Landcover Classification'),
+    'description': tr(
+        'Classification of landcover for volcanic ash based on Badan Geologi'),
```
Gustry: Ash ?

safe/gui/widgets/dock.py
```diff
@@ -1177,9 +1177,9 @@ def accept(self):
         if self.zoom_to_impact_flag:
             self.iface.zoomToActiveLayer()
 
+        qgis_exposure = self.get_exposure_layer()
```
Gustry: This is an old commit

safe/definitions/exposure_classifications.py
ivanbusthomi: Hi @Gustry and @myarjunar , sorry for the late reply,
I believe it is not only for the volcanic ash, and also can be used for the tsunami hazard because Badan Geologi who uses this classification also develops Tsunami hazard modelling.

Issue #3762, opened 2017-02-02
Add displacement rates and notes


Charlotte-Morgan:
We need to consult experts to find displacement rates for all hazards and then update notes to explain the rates that are assigned

timlinux:
Hi @Charlotte-Morgan can we move this to 4.1 or 4.0.1? I don't think we will have all the info we need by 4.0?

Charlotte-Morgan:
Hi Tim, 
I have consulted earthquake, tsunami, flood and volcano people.  I think we are making good progress and should get these numbers in for 4.0. Its not a big deal to put something in even if we add a note that it is an estimate.
If displacement rates are left as 0 there will be no min needs or age gender reports.
Null reports is a bigger issue for me than documented guesses.

I have a few more to add from the walk through with @RikkiWeber last week.
I think we need some one to contribute info for cyclone @wcarthur @lpower @mbernasocchi

wcarthur:
Proposed displacement rates for TC:

cat 1: 0.00
cat 2: 0.06
cat 3: 0.55
cat 4: 0.97
cat 5: 1.00

These are based on examination of building fragility curves developed for the Greater Metro Manila Area Risk Assessment Project, which was undertaken by University of the Philippines Institute for Civil Engineering. The rates are based on the assumption of low-rise, timber framed construction, with population in buildings suffering moderate or higher damage state.

The damage states give a qualitative indication of the level of damage sustained to a buiding (or components of a building). The damage index quantifies this as a fraction of the complete replacement cost of the structure (component). 


| Damage state | Description                                |
|--------------|:-------------------------------------------|
| Negligible   | Little or no visible damage from the outside. No broken windows, or failed roof deck. Minimal loss of roof over, with no or very limited water penetration. |
| Slight       | Moderate roof cover that can be covered to prevent additional water ingress. One window, door or garage door broken. |
| Moderate     | Major roof damage, moderate window breakage. Minor roof sheathing failure. Some water damage to interior.            |
| Extensive    | Major window damage or roof sheathing loss. Major roof cover loss. Extensive damage to interior from water.          |
| Complete     | Complete roof failure and/or failure of wall frame. Loss of more than 50% of roof sheathing.                         |


[...]

PR #4000, opened 2017-03-03
Displaced rate

ismailsunni:
### What does it fix ?
* Ticket #3762 
* Funded by : DFAT
* Description : 
   - Update displaced rate for cyclone
   - Remove fatality rate for every hazard


safe/definitions/hazard_classifications.py
```diff
@@ -164,7 +161,6 @@
             'string_defaults': ['high'],
             'displacement_rate': 1.0,
             # Not used because EQ algs take care of this
```
Gustry: Can you remove these comments @ismailsunni ? They were linked to fatality_rate

PR #4007, opened 2017-03-03
Test tweaks - improve testing on windows

timlinux:
### What does it fix ?

* Funded by :  DFAT
* Description :  Fixes some tests that fail in windows. We still have an issue that gui package tests crash python and the test suite does not run through all in one go. We also have an issue with some geopackage and CSV loading tests.


timlinux:
I'm going to merge this  - there should not be any side effects.

PR #4010, opened 2017-03-05
Fix failing dock tests (which are anyway expected to fail

timlinux:
### What does it fix ?

* Funded by :  DFAT
* Description :   Removing expected failures allows the full test suite to be run successfully on windows. Also these tests are moribund and I think we will be better off implementing new test cases since the tests used old API logic.


Gustry:
LGTM !
Thanks we wanted to do that too !

Issue #3538, opened 2017-01-05
Some cosmetics on the default values window


Gustry:
### Problem
![screen shot 2017-01-05 at 18 16 14](https://cloud.githubusercontent.com/assets/1609292/21690175/6628e598-d373-11e6-987e-c67a99770aaa.png)

* [x] Items are not ordered. Adult, Elderly and Youth are at the top. Youth is at the bottom.
* [ ] The UI with the sentence just after each item is misleading. I often think the text is a title for the next item but it's not. The text explains the item before. I think we can drop it.
* [x] Replace `mˆ2` per `²`.
* [ ] It could be nice to not be able to set elderly, adult and youth to one and check if the total is equal to 1.

Gustry:
@ismailsunni I think you fixed some of these issues, right ?

ismailsunni:
Hmm, I don't think so @Gustry . I was mostly working on the wizard lately. I just change some text and fix the typo in `eldery`

ismailsunni:
Second point will be addressed in #4029 
4th point will be addressed in #4015 

I am closing this ticket then.

PR #4030, opened 2017-03-08
Fix description option dialog

ismailsunni:
### What does it fix ?
* Ticket : No ticket number
* Funded by : DFAT
* Description : 
   Previously it was like this (missing description)
   screen shot 2017-03-08 at 12 16 59
   Now it has description back
   
   screen shot 2017-03-08 at 12 18 07


Gustry:
Can you fix some of https://github.com/inasafe/inasafe/issues/3538 ?
For instance, the order should be Female, Youth, Adult, Elederly I think.

ismailsunni:
> Can you fix some of #3538 ?

@Gustry yes, I can. But you know, it will be a dirty hack. Anyway, I will do it.

Issue #4033, opened 2017-03-08
Error: Fatality rounding - EQ analysis


Charlotte-Morgan:
### problem ###
there is an error in the rounding of earthquake fatalities.
I have got a big earthquake planned. In the debug analysis - the results show 18589 fatalities but the rounded report shows 0 - 100.

![image](https://cloud.githubusercontent.com/assets/5743135/23698311/af0e829c-041d-11e7-9d5b-cb77585182e0.png)

### proposed solution ###
make it happy - kill more people

Gustry:
It seems the problem comes in the reporting.
I will have a quick look this afternoon if I know where is it done in the reporting engine from @lucernae

Gustry:
Fixed in PR above.

PR #4034, opened 2017-03-08
fix fatality range

Gustry:
### What does it fix ?
* Ticket #4033 
* Funded by : DFAT
* Description : Fix the fatality range


lucernae:
LGTM! You're da best man.
I guess I'm really really sleepy at that time.
Stupid mistake 😟

Gustry:
TDD is the best practice ! :D

Issue #3980, opened 2017-03-02
BUG: gdal.exe window pops up


Charlotte-Morgan:
when i run earthquake raster on pop raster the dgal.exe window pops open and then closes.

Solution: this should not be visible to users

Gustry:
@timlinux @Charlotte-Morgan Can you try on windows if you still have this issue ?

Charlotte-Morgan:
I had this again on 4.0.0 yesterday - ie using master version

Gustry:
I meant on develop, if you pulled this morning. It will be in 4.0.1 if it's working

Gustry:
Can you check if you still have this window @Charlotte-Morgan ?
Or close the ticket, thanks.

Charlotte-Morgan:
I think we can close this

PR #4035, opened 2017-03-08
disable GDAL bash window on windows

Gustry:
### What does it fix ?
* Ticket #3980 
* Funded by : DFAT
* Description : Disable the bash window

Can you try @timlinux on Windows® ?


Gustry:
Could you try in Windows ?

Issue #4026, opened 2017-03-08
Error: Invalid layer


Charlotte-Morgan:
I am getting this error when running raster earthquake on vector population and vector aggregation.

```
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1081, in run self._run()
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1245, in _run self.gis_overlay_analysis()
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1344, in gis_overlay_analysis self.hazard_preparation()
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1665, in hazard_preparation self.hazard = clip_by_extent(self.hazard, extent)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\gis\raster\clip_bounding_box.py", line 86, in clip_by_extent check_layer(clipped)
In file "C:/Users/AIFDR_ADMIN/.qgis2/python/plugins\inasafe\safe\gis\sanity_check.py", line 70, in check_layer raise InvalidLayerError(tr('The layer is invalid.'))
```

Charlotte-Morgan:
'''
			
2017-03-08T09:27:36	0	ANALYSIS : The impact function is starting.
[...]

Issue #3655, opened 2017-01-20
The raster is clipped with a smaller extent than the analysis extent.


Gustry:
### Problem
![screen shot 2017-01-20 at 15 50 37](https://cloud.githubusercontent.com/assets/1609292/22142305/5aad4afe-df28-11e6-8926-4b467a37a220.png)

This is the Maumere scenario.
We can see that one building is marked as 'not exposed' as the tsunami raster is clipped with a smaller extent that the analysis area.

### Expected Behaviour
The clipped raster should have one row more to cover the analysis area.

### Environment
* InaSAFE : 4.0b

timlinux:
Maybe make a 1pix distance buffer around the edge of the aggregation area?

Gustry:
I fixed this issue in the PR above.

PR #4037, opened 2017-03-08
skip clip raster by extent if the layer is invalid

Gustry:
### What does it fix ?
* Ticket #4026 #3655 
* Funded by : DFAT
* Description : Manage raster which are upside-down, make a buffer on the analysis raster


ismailsunni:
Is there possibility where is the layer is leftside right?

timlinux:
I think that is less likely @ismailsunni

Issue #4015, opened 2017-03-06
Check Sum for Age Ratio in Options Dialog


ismailsunni:
### Problem
In the wizard, we have checker for age ratio, we need to apply this also in the options dialog.

### Expected Behaviour
Wizard and options dialog have the same behaviour.

### Proposed Solution

PR #4038, opened 2017-03-09
Add sum age ratio checker in option dialog.

ismailsunni:
### What does it fix ?
* Ticket #4015 
* Funded by : DFAT
* Description : 
   - Add sum age ratio checker in option dialog
screen shot 2017-03-09 at 09 14 13


Gustry:
LGTM

Please merge, so we can translate these new strings ;-)

ismailsunni:
Ah shame, the message box makes the travis not happy. I should check my PR earlier today.

@Gustry You can write the translation in that message box here, and I will translate it for you if you want.

Gustry:
It's not so urgent. I will translate it on transifex when it will be there ;-)

BTW, I would say : 
`You have set age ratio whose sum is not equal to 1.`

PR #4041, opened 2017-03-09
Review provenance to have a better translation in French

Gustry:
### What does it fix ?
* Ticket #
* Funded by : WorldBank
* Description :

### DIFF
--- a/safe/definitions/hazard_exposure_specifications.py
+++ b/safe/definitions/hazard_exposure_specifications.py
@@ -1,13 +1,21 @@
 # coding=utf-8
 
-"""Notes which are specific for an exposure and a hazard."""
+"""Actions, notes, questions which are specials for a given hazard/exposure."""
 
-from safe.utilities.i18n import tr
+from safe.utilities.i18n import tr, locale
 from safe.definitions.hazard import (
     hazard_volcanic_ash,
     hazard_earthquake,
+    hazard_flood,
 )
-from safe.definitions.exposure import exposure_population
+from safe.definitions.exposure import (
+    exposure_population,
+    exposure_road,
+)
+
+# Notes, for the analysis question, this is specific for French. I put the
+# singular and masculine sentence on transifex.
+# I put here when it's plural and/or feminine.
 
 ITEMS = (
     {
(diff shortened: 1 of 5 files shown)


safe/definitions/hazard_exposure_specifications.py
```diff
@@ -55,7 +63,15 @@
                 'model.'),
             # end notes provided by Hadi Ghasemi
         ]
-    }
+    },
+    {
+        'hazard': hazard_flood,
+        'exposure': exposure_road,
+        'analysis_question': {
+            'fr': u"Dans l'événement d'une inondation, quelle longueur de "
```
lucernae: I'm just asking the possibility here.
Does it have to be hardcoded?
Or can we use tr('Some unique statement about this case'),
then translate it (put the translated string) on transifex?

safe/impact_function/impact_function.py
```diff
@@ -856,12 +855,12 @@ def prepare(self):
                 return status, message
 
             # Set the name
-            self._name = tr('%s %s On %s %s' % (
-                self.hazard.keywords.get('hazard').title(),
-                self.hazard.keywords.get('layer_geometry').title(),
-                self.exposure.keywords.get('exposure').title(),
-                self.exposure.keywords.get('layer_geometry').title(),
-            ))
+            self._name = tr('%s %s On %s %s') % (
```
lucernae: can we use {name} format on this (py3 style with named variable)?

safe/impact_function/impact_function.py
```diff
@@ -1997,6 +1996,8 @@ def provenance(self):
         If the impact function is not ready (has not called prepare method),
         it will return empty dict to avoid miss information.
 
+        The impact function will call generate_provenance at the end of the IF.
+
         List of keys (for quick lookup):
 
         [
```
lucernae: @Gustry and @ismailsunni It's hard to keep track what can be expected to be inside the provenance for now.
So, if possible, if there's any changes in the future, please change this list accordingly (for example, we might want to delete report_question?).
Best solution would be to declare another container class for provenance (for self._provenance object).
That way, pycharm can detect its properties without having to run it.

Gustry: Yes, agree!

safe/definitions/hazard_exposure_specifications.py
Gustry: I didn't want you to translate this kind of strings on transifex. If you want, we can put the 'tr' if you think we need it. But as you told me these generated strings works well for you, I though it's fine.

Issue #3998, opened 2017-03-03
Too Large Font Size in PDF Map Legend


uswatunkh21:
### Problem
Font size in the PDF Map legend is too large. It then overlaps with other text in the PDF Map.
![too big font size](https://cloud.githubusercontent.com/assets/2535397/23543778/5769e04a-0026-11e7-85c2-9ba4135ae666.png)

### Expected Behaviour
Font size in the PDF Map legend should be in proportional size in order it do not overlaps with other text in the PDF Map.
![true font size](https://cloud.githubusercontent.com/assets/2535397/23543914/180a3552-0027-11e7-8e86-a92cdaa19748.png)

### Environment
* InaSAFE : 4.0
* QGIS : 2.14.8
* OS : Winddows 10
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : Padang building point (download with OSM downloader
  * Hazard : Padang_EQ_2009_WGS84.tif
  
### Proposed Solution
Minimize the font size, or split the legend into two columns

lucernae:
@ivanbusthomi could you kindly take this over? I think this is about the template. You can give the template and assign it to me later.

ivanbusthomi:
@lucernae sorry I missed this ticket yesterday.
I will create the template quickly and post it here

ivanbusthomi:
@lucernae this is the zip template, sorry for the delay
[20170306_fix_template_v4.zip](https://github.com/inasafe/inasafe/files/820022/20170306_fix_template_v4.zip)

Gustry:
@lucernae Can you have a look ? We will release 4.0.1 at the end of the week.

lucernae:
Sorry, I will have a look at Thursday.

lucernae:
Nice!
<img width="566" alt="screen shot 2017-03-09 at 21 49 11" src="https://cloud.githubusercontent.com/assets/831865/23755263/4b6cd27a-0512-11e7-9407-eb153142b7cd.png">

Gustry:
@lucernae Can we close this ticket ?

lrpower:
The legend text is overflowing for Cyclone IF landscape report. Maybe the (severe tropical cyclone) could be removed?
![image](https://cloud.githubusercontent.com/assets/10228025/24383474/c959c284-13a8-11e7-9167-0a509b26aec3.png)

ivanbusthomi:
[...]

PR #4042, opened 2017-03-09
Report fixes

lucernae:
### What does it fix ?
* Ticket #3887,  #3998
* Funded by : DFAT
* Description : 

Changes:
- Refactor analysis provenance details to metadata
- Add ordering for aggregation results by exposure class
- Update QGIS composer map template
- Update infographics


safe/definitions/reports/components.py
```diff
@@ -346,6 +348,49 @@
...
+                'jinja2/'
+                'analysis-provenance-details.html',
+    'extra_args': {
+        'defaults': {
+            'source': tr('source not available'),
+            'reference': tr('reference unspecified'),
+            'aggregation_not_used': tr('not used')
+        },
+        'header': {
+            'analysis_detail': tr('Analysis details')
```
Gustry: Thanks ;-)

safe/definitions/reports/components.py
```diff
@@ -346,6 +348,49 @@
...
+
+            'aggregation_header': tr(
+                'Aggregation source'),
+            'aggregation_format': tr(
+                '{layer_name} - {source} - '),
+
+            'impact_function_header': tr(
+                'Impact Function'),
+            'impact_function_format': tr(
+                '{impact_function_name}'),
```
Gustry: IMHO, we don't need tr() here on these lines 379, 384.

lucernae: Yes for 379, I think we don't need that.
For 384, I guess impact function name were already translated.

Gustry: In transifex we will copy paste the block because there is nothing to translate here.
So to have this tr to works, we would need to put every combinations of hazard/exposure in https://github.com/inasafe/inasafe/blob/develop/safe/common/dynamic_translations.py
This would be a waste of time ;-)

 I just put a commit to translate the name of the IF before your side : https://github.com/inasafe/inasafe/pull/4041/commits/1670822f2ac3de1f5bc9db28d773e1e9865977cc 

So i think it's safe if you try to translate this ;-)

lucernae:
Tada!
Now it can display 100% slice with some 'hacky' way.
I also cleaned up the svg icons, now it can be scaled using center anchor points.
I'd like to go to Himalaya to make myself an SVG monk.

P.S. for @timlinux if ubuntu font not exists, it will fallback to Segoe UI (windows metro font default).
But I haven't tested it yet. Maybe you could test it later.
infographic

resources/css/InaSAFE.css
```diff
@@ -1,5 +1,5 @@
 body {
-  font-family: 'Ubuntu', Fallback, sans-serif;
+  font-family: 'Ubuntu', 'Lucida Grande', 'Segoe UI', sans-serif;
```
lucernae: Lucida Grande is fallback for MacOS.
Segoe UI is for Windows 8+.
sans-serif is the default ugly one.

Issue #4039, opened 2017-03-09
BUG: negative value in min needs report


Charlotte-Morgan:

Gustry:
@Charlotte-Morgan Can we have more inputs ? datasets, extent ...

I tried to replicate with Maumere tsunami raster on population raster : 
![capture d ecran 2017-03-09 a 08 34 13](https://cloud.githubusercontent.com/assets/1609292/23740325/60f732da-04a3-11e7-8bf3-08cbf68f201b.png)

I tried with EQ raster on population raster too : 
![capture d ecran 2017-03-09 a 08 37 29](https://cloud.githubusercontent.com/assets/1609292/23740403/b05c1a2a-04a3-11e7-8933-ae0e28394281.png)

Everytime I use the biggest area as possible. I can't replicate.

Charlotte-Morgan:
hi @ivanbusthomi - can you please replicate?
use Sunda WorldPop 2015, Sunda Slab EQ
share the world pop if necessary, I already shared hazard data yesterday.
Thanks

Charlotte-Morgan:
my negative is only in the general report not the detailed report

samnawi:
Hi @Gustry I got the same problem with @Charlotte-Morgan when I doing analysis for whole area.
![image](https://cloud.githubusercontent.com/assets/8871958/23744566/4b225cca-04e7-11e7-91e7-1ed216d96d4b.png)

But, when I tried in the smaller area, the result is fine
![image](https://cloud.githubusercontent.com/assets/8871958/23744579/5a60ccc6-04e7-11e7-8893-f11da1687cd9.png)

ivanbusthomi:
Hi @Gustry 
I can replicate the problem. I already put the exposure data in `Real_world_data` BtSync, under `Sunda` folder

Gustry:
Thanks, I will check

Gustry:
I could replicate the issue. We need a big dataset for this one.

The number of displaced : 47 122 267
In the BNPB profile, we distribute 67 L of clean water for each displaced person : 
67 * 47 122 267 = 3 157 191 916

There is somehow a maximum of the number. We use memory layer and geojson layer during the process.

Solutions : 
* Switch to m³ in the profile, the easiest one for us.
* Switch automatically to m³ if needed but we can't do that now, InaSAFE is not aware about the original unit. We have to update the minimum needs to be able to set the unit from a combobox ans we will be able to know which units are compatible each other.
[...]

PR #4044, opened 2017-03-10
Allow QLongLong length 11 in minimum needs fields

Gustry:
### What does it fix ?
* Ticket #4039 
* Funded by : DFAT
* Description : Increase length size in minimum needs fields

capture d ecran 2017-03-10 a 09 39 30


ismailsunni:
LGTM

Issue #3607, opened 2017-01-17
Needs value for hygiene packs and weekly rice for pregnant and lactating women


lucernae:
CC @Charlotte-Morgan 

I just need to know/confirm the value of these minimum needs (based on number of females displaced):

1. Weekly hygiene packs:

    ```
    weekly hygene packs = affected pop * fem_ratio * 0.7937 * week / intended day-of-use
    ```

   in the code:
   displaced female * 0.7937 * ( 7 days / 7 days)

2. Additional weekly rice for pregnant and lactating women:
    in the code:
    rice for lactating women = displaced female * 2 * 0.033782
    rice for pregnant women = displaced female * 2 * 0.01281
    total additional rice = rice for lactating women + rice for pregnant women

Charlotte-Morgan:
@samnawi  - can you please check this one for @lucernae ?

samnawi:
OK @Charlotte-Morgan I'll check it

Charlotte-Morgan:
hi @lucernae  - seems like this is a good question.
The minimum needs for this are not specified in the BNPB Perka 7 / 2008 nor in sphere standard.
We need to review some old tickets to find out when this was added to the code and where the advice came from.
@timlinux

samnawi:
Hi @Charlotte-Morgan I still not found about the regulation of the stuff in above. I just found some link in your comment https://github.com/inasafe/inasafe/issues/1230#issuecomment-60903150 that refers to some standard. But I'm not find anything about the specific number like in above.

Charlotte-Morgan:
@samnawi  - thanks for looking. This could go way back to some of the original work.

samnawi:
Hi @Charlotte-Morgan I found a ticket that seems like an initial discussion about a vulnerable group in here https://github.com/inasafe/inasafe/issues/312, but I still try to find a ticket of the further discussion about it.

lucernae:
@samnawi I don't know if this will help, but you can find the value in InaSAFE code here:

https://github.com/lucernae/inasafe/blob/e97187fda0f91ecc388995600abfe444ce0fa389/safe/postprocessors/gender_postprocessor.py#L110-L110

There is also some comments related to it.

ismailsunni:
@samnawi 
Sorry, I missed to read this thread, perhaps because of the full of notifications.
[...]

PR #4045, opened 2017-03-10
regarding issue #3607: change calculation of additional rice

lucernae:
### What does it fix ?
* Ticket #3607 
* Funded by : DFAT
* Description : change additional rice calculation

### DIFF
--- a/safe/definitions/post_processors.py
+++ b/safe/definitions/post_processors.py
@@ -422,18 +422,28 @@ post_processor_additional_rice = {
     'input': {
         'female_displaced':
             {
-                'value': female_displaced_count_field,
+                'value': displaced_field,
                 'type': field_input_type,
             },
         'additional_rice_ratio':
             {
                 'type': constant_input_type,
-                'value': 2 * (0.033782 + 0.01281),
+                'value': (0.024 + 0.026) * 500 * 7 * 0.1 / 129,
             }
     },
     'output': {
         # The formula:
-        # displaced_female * 2 * (0.033782 + 0.01281)
+        # See: https://github.com/inasafe/inasafe/issues/3607
+        # for reference
+        #
+        # displaced_population * (pregnant_rate + breastfeeding_rate) *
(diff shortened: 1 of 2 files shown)


lucernae:
Sample calculations:

Without rounding:
screen shot 2017-03-10 at 16 14 27

With rounding:
screen shot 2017-03-10 at 16 14 48


Formula that is being used:
22110 is the total displaced
3299 = 22110 * (0.024 + 0.026) * 550 * 7 * 0.1 / 129

Gustry:
Can you update the post processor tree at the bottom of the file? Thanks

fredychandra:
Hi @lucernae 

I try to calculate the extra rice as below. Please compare with the formula that you have shown above. I find a different result with your calculation. 

Simulation : 
Total Displaced "Jati"= 22200
Ratio Breastfeeding and Pregnant = (2,4% + 2,6%) = 5 %
Extra Rice need for BreastFeeding and Pregnant Women is 0,3876 Kg/day

Extra Rice per week = 22200 * 0.05 * 0,3876 Kg/day * 7 (days)
                                = 3011.652 Kg/week
                                = 3011,7 Kg/week (Rounded)

lucernae:
Hi mas @fredychandra , sorry. My mistake, I read the extra calories as 550Kkal instead of 500. 👍

lucernae:
For @Gustry , I think the execution order doesn't change. Semantically it should be calculated after gender postprocessors.

Gustry:
The order doesn't change, but the branch is not the same in the tree.
I think you need to remove one tab, it will be like this L789 : 
```
# |      |--- gender
# |      |--- hygiene packs
# |      |--- additional rice
```
This is cosmetics, but is usefull to quickly see which processor needs to run before.

lucernae:
Ah you mean that one. What does the tree represents? Is it dependencies? If that is so, then I only need to move additional rice and let hygiene packs be.

PR #4046, opened 2017-03-10
add other group to Badan Geologi Landcover

Gustry:
### What does it fix ?
* Funded by : DFAT
* Description : Add other group in Badan Geologi classification

### DIFF
--- a/safe/definitions/exposure_classifications.py
+++ b/safe/definitions/exposure_classifications.py
@@ -1,5 +1,7 @@
 # coding=utf-8
+
 """Definitions relating to exposure classifications."""
+
 from safe.utilities.i18n import tr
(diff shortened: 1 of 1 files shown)


safe/definitions/exposure_classifications.py
```diff
@@ -764,16 +766,24 @@
...
+                    'text': None,
+                    'link': None
+                }
+            ]
         }
     ]
 }
 
-# Classes order
-structure_class_order = [
```
ismailsunni: Do we not use it anymore?

Gustry: Not anymore. It was legacy from 3.5

Issue #4055, opened 2017-03-16
Hazard Keyword from Peta Bencana Downloader


samnawi:
### Problem
The hazard keyword for flood data that we download from petabencana.id set as Generic Class.
![image](https://cloud.githubusercontent.com/assets/8871958/23981960/dcb1b302-0a3c-11e7-8133-1c12b31c26a0.png)

Meanwhile, if we look at the flood classification in petabencana.id, similar with the classification in the flood classes at the keyword 
![image](https://cloud.githubusercontent.com/assets/8871958/23982101/c8809c9e-0a3d-11e7-9aae-bf97d1900bdd.png)

### Expected Behaviour
I just think, should we align the keyword from petabencana downloader with the hazard classification on petabencana.id (flood classes)

### Proposed Solution
Make it happen

cc : @Charlotte-Morgan

Gustry:
@myarjunar Can you have a look to this issue ? Thanks

myarjunar:
closed in #4056

PR #4056, opened 2017-03-16
match petabencana's keyword with its classification

myarjunar:
### What does it fix ?
* Ticket #4055 
* Funded by : DFAT
* Description : Change the petabencana keywords so the keywords match with its classification.

selection_042


Gustry:
LGTM thanks !

Issue #4077, opened 2017-04-03
Metadata and Definitions Support for #4069


ismailsunni:
### Problem
To be able support feature request on #4069 , we need to update metadata and defintions. Some requirements:
1. Some InaSAFE fields (e.g. youth, adult, female, etc) can be in several fields. It will be used in 
2. We need grouping mechanism for InaSAFE fields. E.g. Age ratio for youth_ratio_field, adult_ratio_field, and elderly_ratio_field in age_ratio_group
3. The group should have a constraint mechanism. E.g. the sum of age ratio group is at most 1.

### Proposed Solution
I think the solution is quite simple. We just need to make the inasafe fields also able to store list (currently, it's only a string). It will solve first point above.
Point 2 should be handled in defintions and the last one is more complicated, currently it's hard coded.

ismailsunni:
Hi @Gustry 

With my PR above, we will have not only string for the fields, but also list of string (list of fields). I think you just need to sum the value in the fields if it's a list of fields to create the standard InaSAFE fields (e.g. youth_count_field). And the process should be the same after that. Is that ok for you?

CMIIW, the grouping will not affect your impact function, but rather in the reporting, won't it?

Please check my PR above and if it's ok, I can continue my work and you can create dummy test data according to this ticket.

cc @timlinux

ismailsunni:
Per #4079 merged, we have addressed point 1 and 2. For point 3, since it's more complicated and can be implemented later after the tools / wizard has been finished, I will move to the new issue. Closing this.

PR #4079, opened 2017-04-03
Fix 4077 metadata support

ismailsunni:
### What does it fix ?
* Ticket : #4077
* Funded by : DFAT
* Description : 
  - Make sure the metadata able to support field groups
  - Add group in the field

### DIFF
--- a/safe/definitions/field_groups.py
+++ b/safe/definitions/field_groups.py
@@ -20,7 +20,7 @@ __revision__ = '$Format:%H$'
 
 age_count_group = {
     'key': 'age_count_group',
-    'name': 'Age Count Group',
+    'name': tr('Age Count'),
     'description': tr(
         'The group of fields that consists of population count per age '
         'class.'),
@@ -33,7 +33,7 @@ age_count_group = {
 
 age_ratio_group = {
     'key': 'age_ratio_group',
-    'name': 'Age Ratio Group',
+    'name': tr('Age Ratio'),
     'description': tr(
         'The group of fields that consists of population ratio per age '
         'class.'),
--- a/safe/definitions/field_groups.py
+++ b/safe/definitions/field_groups.py
@@ -1,0 +1,50 @@
+# coding=utf-8
+
+"""Definitions relating to group of fields."""
+
+from safe.utilities.i18n import tr
+
+from safe.definitions.fields import (
+    youth_count_field,
+    adult_count_field,
+    elderly_count_field,
+    youth_ratio_field,
+    adult_ratio_field,
+    elderly_ratio_field
+)
+
+__copyright__ = "Copyright 2017, The InaSAFE Project"
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = '$Format:%H$'
+
+age_count_group = {
+    'key': 'age_count_group',
+    'name': 'Age Count Group',
+    'description': tr(
+        'The group of fields that consists of population count per age '
+        'class.'),
+    'fields': [
+        youth_count_field,
+        adult_count_field,
+        elderly_count_field
+    ]
+}
+
+age_ratio_group = {
+    'key': 'age_ratio_group',
+    'name': 'Age Ratio Group',
+    'description': tr(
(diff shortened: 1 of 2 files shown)


safe/definitions/field_groups.py
```diff
@@ -0,0 +1,50 @@
...
+    'fields': [
+        youth_count_field,
+        adult_count_field,
+        elderly_count_field
+    ]
+}
+
+age_ratio_group = {
+    'key': 'age_ratio_group',
+    'name': 'Age Ratio Group',
```
Gustry: no translation on these two names ?

safe/utilities/test/test_metadata.py
```diff
@@ -24,6 +24,7 @@
 
 class TestMetadataUtilities(unittest.TestCase):
     """Test for Metadata Utilities module."""
+    maxDiff = None
```
Gustry: Landscape likes when there is one blank line before and after class docstring ;-)

safe/definitions/field_groups.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+__copyright__ = "Copyright 2017, The InaSAFE Project"
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = '$Format:%H$'
+
+age_count_group = {
+    'key': 'age_count_group',
+    'name': 'Age Count Group',
```
timlinux: Remember Charlotte didn't like the word 'Group' too much, so for the user facing name, just call it 'Age counts' (and similar for the others).

safe/definitions/field_groups.py
```diff
@@ -0,0 +1,50 @@
...
+    'description': tr(
+        'The group of fields that consists of population count per age '
+        'class.'),
+    'fields': [
+        youth_count_field,
```
timlinux: You are going to add the other two age classes here later?

safe/definitions/field_groups.py
```diff
@@ -0,0 +1,50 @@
...
+}
+
+age_ratio_group = {
+    'key': 'age_ratio_group',
+    'name': 'Age Ratio Group',
+    'description': tr(
+        'The group of fields that consists of population ratio per age '
+        'class.'),
+    'fields': [
+        youth_ratio_field,
```
timlinux: You are going to add the other two age classes here later?

safe/utilities/test/test_metadata.py
```diff
@@ -34,7 +35,7 @@ def test_write_iso19115_metadata(self):
         keywords = {
             'date': '26-03-2015 14:03',
             'exposure': 'structure',
-            'keyword_version': '3.2',
+            'keyword_version': '',
```
timlinux: 4.1?

safe/definitions/field_groups.py
ismailsunni: Ah yes, sorry. Too much c++ lately :)

safe/definitions/field_groups.py
ismailsunni: Ok.

safe/definitions/field_groups.py
ismailsunni: Yes, for now I will just work on prove on concept. Adding more classess or group should be easy if everything if the framework has been finished.

safe/definitions/field_groups.py
ismailsunni: Yes, same as above

safe/utilities/test/test_metadata.py
ismailsunni: It was meant to be inasafe_keyword_version like the other two in the same file. Thanks for pointing this.

safe/utilities/test/test_metadata.py
ismailsunni: Ah yes, thanks. Class from the old time :)

ismailsunni:
> I guess you are just making it work with the existing classes first and infant etc. are coming later?

@timlinux yes, exactly. Make it work first, then adding new classes should be simple since we don't hard code anything.

Issue #3802, opened 2017-02-08
Hazard Classes: Cyclone needs work


Charlotte-Morgan:

timlinux:
cc @mbernasocchi

Charlotte-Morgan:
hi @samnawi - can you please look at BNPB perka 2 / 2012 and update this ticket with hazard classification used for strong wind in Indonesia?  How many classes and what are the thresholds and what units are used.?

Charlotte-Morgan:
hi @maning - in InSAFE version 4.0 we are introducing cyclone impact analysis for partners in the Pacific.  Do you have any advice on hazard classification for wind hazard in the Philippines?

mbernasocchi:
Hi @Charlotte-Morgan I asked @wcarthur to have a look at this. 
re the indonesia scale, IIRC it is the same as the Aus scale

Charlotte-Morgan:
hey @mbernasocchi - thanks for getting back so quickly. I will get a chance to talk to Craig and Martine tomorrow. I'm fairly confident there is a problem, but let me discuss with them over a whiteboard for more clarity. The current scale suggests that you have multiple categories in a single event and that is not my understanding.

mbernasocchi:
how do you mean @Charlotte-Morgan? 
in one event you can have multiple zones with different wind speed, is that what you mean with multiple categories?

Charlotte-Morgan:
You would not have multiple categories in a single event.
A cyclone would be described as Category 3 at any given time; it may be upgraded or downgraded during its approach and as it crosses land. At landfall it will be a certain category with wind speed predicted up to a certain speed.  
Its a scale issue: the category classification is between events; we need classes within events.
Have a look at warnings for northern australia - there is a cyclone approaching now

mbernasocchi:
well not really @Charlotte-Morgan usually the cyclone is classified to its maximum intensity (usually the eye wall) but then at further distance from the eye it is not the same wind speeds that are expected. this is exactly what tcrm does calculate.

this is a TCRM screenshot with a "random" scale where the buildings are assigned the intensiti of cyclone that should touch them. Does this make sense?

![image](https://cloud.githubusercontent.com/assets/233663/22761465/1706597a-ee5b-11e6-9a59-0ca9a3ab37d6.png)

mbernasocchi:
[...]

PR #4096, opened 2017-04-11
Standardise wind speed averaging time periods

wcarthur:
### What does it fix ?
* Ticket : 3802
* Funded by : Geoscience Australia/PacSAFE
* Description : Standardise wind speed averaging time period, per comments in #3802. All wind speed classifications now assume the winds are representative of a 0.2 second gust wind speed. Users will need to convert TC hazard layers to this wind speed averaging time period before importing into InaSAFE. Recommend there be a pre-processor that converts between different wind speed averaging time periods in future releases.


wcarthur:
For the merge conflicts, suggest removing reference to the Beaufort scale, which is somewhat antiquated and only a qualitative classification of winds.

timlinux:
Thanks @wcarthur I have merged your work via https://github.com/inasafe/inasafe/pull/4133

Please start a new PR if you want to make any further changes. It might be good if you take a quick scan for #4133 changes to make sure the merge conflicts were all handled satisfactorily.

PR #4097, opened 2017-04-11
for not population, display real value if <1000

### LINKED ISSUE #4062: UNDO: rounding on infrastructure report
### problem ###
This is a ticket to request that a previous feature request is reversed :)
When I run InaSAFE analysis on infrastructure, the results are rounded (as per user requirements). However; if I have only 1 nuclear power plant in the hazard area it can be alarming for users to see results suggesting that there are 10 nuclear power plants in the hazard area.

### proposed solution ###
that we modify the rounding applied to buildings, roads and infrastructure to ensure that small numbers are not rounded in a way to mislead users.

Open to suggestions: but maybe we could round values greater than 1000 but use analysis estimates for values less than 1000.

@timlinux @Gustry @ivanbusthomi @samnawi @felix-yew - any suggestions?

Gustry:
### What does it fix ?
* Ticket : part fix #4062
* Funded by : DFAT
* Description : This will display the real value when it's not population and value <1000 in the legend only. For the report, @myarjunar (cc @lucernae ), can you have a look after this PR is merged?

### DIFF
--- a/safe/utilities/rounding.py
+++ b/safe/utilities/rounding.py
@@ -13,14 +13,19 @@ __email__ = "info@inasafe.org"
 __revision__ = '$Format:%H$'
 
 
-def format_number(x, enable_rounding=True, coefficient=1):
+def format_number(
+        x, enable_rounding=True, is_population=True, coefficient=1):
     """Format a number according to the standards.
 
     :param x: A number to be formatted in a locale friendly way.
     :type x: int
 
-    :param enable_rounding: Flag to enable a population rounding.
+    :param enable_rounding: Flag to enable a rounding.
     :type enable_rounding: bool
+
+    :param is_population: Flag if the number is population. It needs to be
+        used with enable_rounding.
+    :type is_population: bool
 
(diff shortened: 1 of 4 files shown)


safe/utilities/rounding.py
```diff
@@ -13,15 +13,20 @@
 __revision__ = '$Format:%H$'
 
 
-def format_number(x, enable_rounding=True, coefficient=1):
+def format_number(
+        x, enable_rounding=True, is_population=True, coefficient=1):
```
Gustry: is_population should be False by default I think. But I don't want to break tests in reporting for now.

safe/utilities/rounding.py
```diff
@@ -31,7 +36,7 @@ def format_number(x, enable_rounding=True, coefficient=1):
     :rtype: basestring
     """
     if enable_rounding:
-        x = population_rounding(x)
+        x = rounding(x, is_population)
 
     x /= coefficient
```
ismailsunni: :D, long time no see this kind of code

Issue #4071, opened 2017-03-31
Global Ratio defaults are not being used in the IF


lrpower:
### Problem

I am changing the Ratio Global Default values but they are not carrying through to the IF
![image](https://cloud.githubusercontent.com/assets/10228025/24532479/8a7b9e82-160c-11e7-9e1b-d0720cd89cb2.png)
![image](https://cloud.githubusercontent.com/assets/10228025/24532580/51de1cfc-160d-11e7-8697-fd9f2d5e7a7a.png)

### Expected Behaviour
The global ratio values set will be used.
 
### Environment
* InaSAFE : 4.0-develop
* QGIS : 2.18.3
* OS : windows
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : Population raster
  * Hazard : Tsunami
  * Aggregation :n/a
  * Extent : Tongatapu
  
Checklist (if applicable): 
* [ ] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? 

### Proposed Solution
Use the global ratio values that are set

ismailsunni:
@lrpower did you re-assign your layer's metadata using keywords wizard again? You need to re-assign the metadata, since the metadata only stores the ratio, not global or not (to avoid changing result unexpectedly)

lrpower:
Yes. deleted .xml and reran keywords. 
Still not evident. :(

lrpower:
And a QGIS reboot did not help either.

Gustry:
* Is the value updated in your keywords? (when you click on layer in the legend, you should see keywords on the right)

* Can you copy paste InaSAFE logs here? Open the log panel with the small icon in the bottom right corner of QGIS, it looks like a bubble comic strip, run the IF and then copy/paste all logs from the InaSAFE tab. Thanks.

lrpower:
No, the value is not updated in the keywords. 
The log isn't that descriptive
`
2017-04-07T16:01:43	0	Showing No Keywords Message
2017-04-07T16:01:43	0	Showing No Keywords Message
2017-04-07T16:02:07	0	Showing No Keywords Message
'

Gustry:
I'm trying to replicate your issue. Values are updated in my keywords:
* I changed my global defaults in InaSAFE
* I run the keyword wizard to update values
[...]

PR #4099, opened 2017-04-11
Global default

Gustry:
### What does it fix ?
* Ticket : #4071 
* Funded by : DFAT
* Description : Use global default which are set by the user, not from InaSAFE

Need to backport the commit to master

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -83,6 +83,7 @@ from safe.impact_function.provenance_utilities import (
 )
 from safe.definitions.constants import (
     inasafe_keyword_version_key,
+    GLOBAL,
     ANALYSIS_SUCCESS,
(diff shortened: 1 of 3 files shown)


safe/impact_function/impact_function.py
```diff
@@ -1442,16 +1444,17 @@ def aggregation_preparation(self):
                         if count_key not in keywords['inasafe_fields']:
                             # The exposure hasn't a count field, we should add
                             # it.
-                            default = definition(ratio_field)['default_value']
+                            default_value = get_inasafe_default_value_qsetting(
```
ismailsunni: Just trying to remember, so even if there is no ratio (e.g. no youth_ratio), it will take from the setting?

Gustry: I hope so, I will check

Issue #4069, opened 2017-03-24
RFC: Population analysis and reporting


Charlotte-Morgan:
## Problem
We would like to enhance InaSAFE to provide a high level of support for Contingency Planning. To do this we need to enhance the ability of InaSAFE to analyse and report on population data; including  reporting against age, gender and vulnerability measures.

This work is a high priority under DFAT contract for InaSAFE development in Indonesia.

This diagram summarises the current capacity of InaSAFE to report against population variables and the requirements for a BNPB contingency plan. This RFC addresses the age, gender and vulnerability requirements for contingency planning. Population displacement rates and fatalities are managed through hazard data and any hazard specific fragility functions. These are not included in this RFC.
![image](https://cloud.githubusercontent.com/assets/5743135/24278085/2b95f7e8-1074-11e7-8636-18f333584cf1.png)

In some cases; a hazard scenario may be homogeneous over a large area and appear to affect the whole population equally. The ability to identify and locate the most vulnerable sectors of the population allows government and humanitarian responders to prioritise contingency and response planning.

## Duration
1 week
DMI will meet with humanitarian partners on 11th April to test and finalise requirements
## Proposers
DMI Minions on behalf of BNPB, UNOCHA and the Humanitarian sector in Indonesia.
@samnawi 
@ivanbusthomi 
@felix-yew 
@fredychandra 

## Details
InaSAFE 4.0 has 3 ways of estimating the impact of a hazard event on population.
1. Global defaults - rates by age and gender
2. Population exposure data - with counts for total population (raster) and by age and gender (vector)
3. Aggregation data - with rates for age and gender

Users are able to specify if they want to report on age and or gender and how variables in the exposure and aggregation data will be used. Details of the affected population are used to generate the minimum needs and infographic reports.

## Proposed solutions
**1.** update global defaults to include more age, gender and vulnerability measures using international 'standards' where possible.
**2.** enhance InaSAFE to allow value mapping of population variables (counts in vector population exposure data) against a broader range of age, gender and vulnerability measures. 
**3.** enhance InaSAFE to allow value mapping of population variables (rates in vector aggregation data) against age, gender and vulnerability measures.

These proposed enhancements will:
* align well with contingency planning requirements in accordance with government regulations and international humanitarian practice.
* allow users to analyse a subset of the population - ie just youth, just women, just disabled.
* allow users to report against a subset of the population and in combination with customised minimum needs - create humanitarian, sector specific reports - ie shelter needs for disabled

[...]

Issue #4103, opened 2017-04-13
Polishing Field Mapping Tool


ismailsunni:
### Problem
The field mapping tool to address #4069 is already in this PR #4094. Most of the functionalities are working. But we can do some polishing before the release. This ticket will list all possible polishing to make the ticket more granullar:

- Write Metadata
    - [x] drag and drop
    - [x] Check spin box is stored properly
    - [x] Save to metadata
    - [x] Unit test for tab and widget
- Read metadata
    - [x] Handle single field name from old keywords
    - [x] Populate field based on metadata
    - [x] Clear used field from the metadata
- Misc
    - [x] Text for description and header text
    - [x] Help page
    - [ ] Better size
    - [x] Filter layer to only show exposure and aggregation layer 
    - [x] Auto select the current selected layer
    - [x] Filter fields —> int and float only
    - [x] Show 10 unique values
- Add to wizard
    - [x] Check the wizard flow #4120
    - [x] Add the step to the wizard #4120
    - [x] Remove unused inasafe fields or default fields
    - [x] Make wizard able to select more than one field for population count #4119 
- Others
    - [x] Show pretty fields in the dock
    - [x] Update definitions.py to list non-group fields
    - [x] Update parameters repository
    - [x] @timlinux to update help definition  #4105
    - [x] Add test layer
    - [x] Add icon for the field mapping tool cc @MariaSolovyeva #4104

ismailsunni:
We can close this now. The size is not high priority.

PR #4114, opened 2017-04-27
Fix 4069 handle new metadata

ismailsunni:
### What does it fix ?
* Ticket : #4069 
* Funded by : DFAT
* Description : 
   - Sum fields that fall to the same inasafe_fields
   - Fix some bugs in the Field Mapping Dialog


safe/common/test/test_version.py
```diff
@@ -58,8 +58,9 @@ def test_get_version(self):
     def test_get_current_hash(self):
         """Test for get_current_hash."""
         git_hash = current_git_hash()
-        if 'win32' in sys.platform:
+        if 'win32' in sys.platform or 'darwin' in sys.platform:
```
Gustry: `if sys.platform in ['win32', 'darwin']:` ?

Gustry: What happened today to get this test pass? Something changed?

ismailsunni: It was `'win32' in sys.platform`, so the sys.platform has substring `win32`. So, it can be failed if I check the reverse you suggested. I don't know the output of sys.platform in Windows, so just making sure.

Gustry: The output of is `win32`: http://stackoverflow.com/questions/446209/possible-values-from-sys-platform
ok

samnawi:
Yes @ismailsunni it's work. 
But I still curious to test in debug mode, because it's not work at that mode.
Also some inappropriate error message notification after I choose the wrong layer (no keyword layer). Could the field mapping just recognize the layer with keyword inside?

field mapping

ismailsunni:
> Yes @ismailsunni it's work. But, I still not see the result for vulnerability group.

We need to wait for @myarjunar to finish. It's working in the background. We just need to show it in the report... 

> But I still curious to test in debug mode, because it's not work at that mode.
Also some inappropriate error message notification after I choose the wrong layer (no keyword layer). Could the field mapping just recognize the layer with keyword inside?

I will check this, since I haven't got the same problem

Thanks.

ismailsunni:
> Could the field mapping just recognize the layer with keyword inside? 

Yes, it's already in my to do list https://github.com/inasafe/inasafe/issues/4103 in item: Filter layer to only show exposure and aggregation layer

ismailsunni:
Hi @samnawi and @inasafe/developers 

I will just merge this PR so that @myarjunar can try to see what's the result with real data to make the report. The main functionality is working. I will continue to polish it here and there.

There is a bug that @samnawi mentioned above, but after I checked, it's already existed. I made a ticket for it in https://github.com/inasafe/inasafe/issues/4116

myarjunar:
hi @samnawi , here is the snippet of how the new grouping report will look like (based on dummy data and keywords). And since this PR already been merged, I will make it works with real data. It would be nice if there is any feedback from you. cc: @ismailsunni @timlinux @Charlotte-Morgan 

selection_059

ismailsunni:
Hi @myarjunar I think it's better to make a specific ticket for the reporting and have a discussion in the ticket rather starting discussing in closed PR.

PR #4118, opened 2017-04-28
Fix 4103 polish field mapping

ismailsunni:
### What does it fix ?
* Ticket : #4103 
* Funded by : DFAT
* Description : 
   - Filter out layer that doesn't have layer groups.
   - Tidy up definitions.


ismailsunni:
I will merge this since it's quite straight forward to fix a bug.

PR #4122, opened 2017-05-02
Reporting global default

myarjunar:
### What does it fix ?
* Ticket : #4069 
* Funded by : DFAT
* Description : 
We can add group notes in field groups definition. 

selection_063


timlinux:
LGTM thanks

Issue #4113, opened 2017-04-26
Polishing Option Dialog


ismailsunni:
### Problem
Current state:
![Option Dialog](https://files.gitter.im/inasafe/inasafe/hfm9/Screen-Shot-2017-04-26-at-9.18.11-AM.png)

Suggestions from @timlinux :
- [x] Tab name: Global Default ->Demographic Defaults  #4121 
- [x] The whole tab content needs to go into a scrollbox widget #4121 
- [ ] If you have time, I think we should tweak the parameters widget so that each input has a little (?) next to the input widget which when clicked will show the extended help_text --> https://github.com/inasafe/parameters/issues/23
- [x] Add basic validation #4126

Gustry:
Nice preview.
Is-it not better having only one scrollbar ? (with Age, Gender, Vulnerability... inside)

ismailsunni:
@Gustry I hope it will be gone when we added scroll bar for the whole widget. Otherwise, I will set it to not use scroll bar. I agree, it's better to have only one.

ismailsunni:
Closing this, since all posible polishing have been implemented

PR #4126, opened 2017-05-05
Add dynamic validator for field group

ismailsunni:
### What does it fix ?
* Ticket : #4113
* Funded by : DFAT
* Description : 
   - Add feature to create validator in field group. Used in Field mapping tool and keywords wizard.
   - Fix python error when select no keywords found layer
field_mapping_validation

### DIFF
--- a/safe/common/parameters/group_select_parameter_widget.py
+++ b/safe/common/parameters/group_select_parameter_widget.py
@@ -67,12 +67,12 @@ class GroupSelectParameterWidget(GenericParameterWidget):
                 double_spin_box = QDoubleSpinBox()
                 self.radio_button_layout.addWidget(double_spin_box, i, 1)
                 double_spin_box.setValue(value.get('value', 0))
-                double_spin_box.setMinimum(value.get('constraint', {}).get(
-                    'min', 0))
-                double_spin_box.setMaximum(value.get('constraint', {}).get(
-                    'max', 1))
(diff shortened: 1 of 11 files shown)


myarjunar:
LGTM

safe/common/parameters/group_select_parameter_widget.py
```diff
@@ -73,6 +73,14 @@ def __init__(self, parameter, parent=None):
                     'max', 1))
                 double_spin_box.setSingleStep(value.get('constraint', {}).get(
                     'step', 0.01))
```
myarjunar: We usually pull left when we can.

safe/common/parameters/validators.py
```diff
@@ -0,0 +1,65 @@
+# coding=utf-8
+"""List of custom validations for parameter container.."""
```
Gustry: Double full stop ?

safe/common/parameters/validators.py
```diff
@@ -0,0 +1,65 @@
...
+    :param validation_message: The message if there is validation error.
+    :type validation_message: str
+
+    :param kwargs: Keywords Argument.
+    :type kwargs: dict
+
+    :returns: Dictionary of valid and message.
+    :rtype: dict
+
+    Note: The code is not the best I wrote, since there are two alternatives.
```
Gustry: 😝

Issue #4128, opened 2017-05-08
Keywords Changed After Analysis


ismailsunni:
### Problem
I was running analysis with aggregation layer. When I tried to open Field Mapping Tool for the aggregation layer, the FMT doesn't show anything beside the dialog.

I tried to track down the problem, the I found out that the keyword monkey patching for the layer gave the wrong keywords.
The keywords/metadata for the aggregation layer become `aggregation_summary`. 

The keyword wizard also read the monkey patch, so it loses the keywords also from the layer.

The dock is fine, since it reads the metadata file.

### Expected Behaviour
The keywords from the monkey patch shouldn't changed after running the analysis. It should only be changed after keywords assignment or doing field mapping tool.

### Environment
* InaSAFE : develop
* QGIS : 2.14.10
* OS : MacOS
* Data used (filenames and paths so as to reproduce, a project is better):
   I think it's happening for all layers.    

### Proposed Solution
Make sure the on the fly keyword is not assigned to the original layer.

ismailsunni:
Fixed in #4132

PR #4132, opened 2017-05-09
Fix 4128 keyword changed

ismailsunni:
### What does it fix ?
* Ticket : #4128 
* Funded by : DFAT
* Description : 
   - Solution: make sure wizard and FMT always read from file, not from layer properties

### DIFF
--- a/safe/gui/tools/wizard/wizard_dialog.py
+++ b/safe/gui/tools/wizard/wizard_dialog.py
@@ -257,26 +257,32 @@ class WizardDialog(QDialog, FORM_CLASS):
         self.lblSubtitle.setText(self.tr(
             'Use this wizard to run a guided impact assessment'))
 
-    def set_keywords_creation_mode(self, layer=None):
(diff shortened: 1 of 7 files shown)


safe/gui/tools/wizard/wizard_dialog.py
```diff
@@ -257,17 +257,18 @@ def set_mode_label_to_ifcw(self):
         self.lblSubtitle.setText(self.tr(
             'Use this wizard to run a guided impact assessment'))
 
-    def set_keywords_creation_mode(self, layer=None):
+    def set_keywords_creation_mode(self, layer=None, keywords=None):
         """Set the Wizard to the Keywords Creation mode.
 
         :param layer: Layer to set the keywords for
```
timlinux: Need docstring for keywords

ismailsunni: Done in the next commit.

PR #4141, opened 2017-05-09
Use value from field to calculate additional rice. Fix #4136

### LINKED ISSUE #4136: Update post_processor_additional_rice with value from layer's keywords
### Problem
Currently, the formula to calculate the additional rice is using hard code value for `pregnant_rate` and `breastfeeding_rate`. Since in v4.1 we introduce pregnant and lactating field, it will be great if we can use it to calculate the additional rice.

For the first implementation, we will merge the rate of pregnancy and lactating.

cc @timlinux @Charlotte-Morgan @samnawi

ismailsunni:
### What does it fix ?
* Ticket : #4136 
* Funded by : DFAT
* Description : 
   - Use value from field to calculate additional rice for pregnant and lactating.


safe/definitions/post_processors.py
```diff
@@ -512,15 +512,15 @@ def post_processor_population_displacement_function(
...
             {
-                'value': displaced_field,
+                'value': pregnant_lactating_displaced_count_field,
                 'type': field_input_type,
             },
         'additional_rice_ratio':
             {
                 'type': constant_input_type,
-                'value': (0.024 + 0.026) * 500 * 7 * 0.1 / 129,
+                'value': 500 * 7 * 0.1 / 129,
```
timlinux: Urf - magic numbers. Can you rather declare constants with meaningful names then create the formula from those?

ismailsunni: There is a comment for it actually. But I agree, I don't like magic number also. Fixed it in the following commit.

Issue #4157, opened 2017-05-15
Add Flag to Make Field Groups not Exclusive


ismailsunni:
### Problem
Currently, when we move a field from field list (in the left) to any field group, it will make the field in the list gone (it's moved). The field is exclusive to a field group. In some case, we want to have different behaviour where we can assign the same field to several field groups.

PR #4161, opened 2017-05-16
Fix 4157 non exclusive field

ismailsunni:
### What does it fix ?
* Ticket : #4157 
* Funded by : DFAT
* Description : 
  - We can put the same field into several field group.
non_exclusive_fmt

cc @Charlotte-Morgan @samnawi

### DIFF
--- a/safe/definitions/field_groups.py
+++ b/safe/definitions/field_groups.py
@@ -66,6 +66,7 @@ age_ratio_group = {
         adult_ratio_field,
         elderly_ratio_field
     ],
+    'exclusive': False,
     'notes': [],
     'constraints': {
         'sum': {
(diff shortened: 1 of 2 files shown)


safe/definitions/field_groups.py
```diff
@@ -66,6 +66,7 @@
         adult_ratio_field,
         elderly_ratio_field
     ],
+    'exclusive': False,
```
timlinux: We should probably annotate these with a comment - I think in 6 months time we might forget what it is for - or a new user may not intuitively understand its purpose.

ismailsunni: Done in the incoming commit.

PR #4172, opened 2017-05-17
remove deprecated calls in the QGIS api

Gustry:
### What does it fix?
* Funded by: DFAT
* Description: Remove deprecated calls in the QGIS api

### DIFF
--- a/safe/gis/vector/prepare_vector_layer.py
+++ b/safe/gis/vector/prepare_vector_layer.py
@@ -9,6 +9,7 @@ from qgis.core import (
     QgsField,
     QgsFeatureRequest,
     QGis,
+    QgsExpressionContext,
     QgsExpression
 )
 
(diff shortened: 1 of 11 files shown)


safe/gis/vector/prepare_vector_layer.py
```diff
@@ -464,22 +465,26 @@ def sum_fields(layer, output_field_name, input_fields):
     # Creating expression
     string_expression = ' + '.join(input_fields)
     sum_expression = QgsExpression(string_expression)
-    sum_expression.prepare(layer.pendingFields())
+    context = QgsExpressionContext()
```
ismailsunni: Is it the correct way to use QgsExpression? Or is it the new way? Just asking since I copied from peta_bencana_dialog.py

Gustry: Yes. The previous code is deprecated.

ismailsunni: Ah I see. Thanks @Gustry

Gustry: Ahah, I rolled back about this change. I will check later again.

PR #4175, opened 2017-05-17
Show memory profile setting

ismailsunni:
### What does it fix?
* Ticket: No
* Funded by: DFAT 
* Description: 
  - Add option to disable memory profile
screen shot 2017-05-17 at 16 00 31

2x faster without profiling.

### DIFF
--- a/safe/definitions/default_settings.py
+++ b/safe/definitions/default_settings.py
@@ -15,6 +15,7 @@ inasafe_default_settings = {
     'showOrganisationLogoInDockFlag': False,
     'developer_mode': False,
     'generate_report': True,
+    'memory_profile': False,
 
     'ISO19115_ORGANIZATION': 'InaSAFE.org',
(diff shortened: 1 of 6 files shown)


safe/definitions/default_settings.py
```diff
@@ -15,6 +15,7 @@
     'showOrganisationLogoInDockFlag': False,
     'developer_mode': False,
     'generate_report': True,
+    'memory_profile': False,
```
Gustry: This will create the entry in the QSettings if the key doesn't exist?

ismailsunni: It will be the default value, if the key doesn't exist. If the user Accept the options dialog, it will be stored in the QSettings. So, more or less, yes.

Gustry: ok, let's merge this so ;-)

PR #4182, opened 2017-05-19
Report rounding

myarjunar:
### What does it fix?
* Ticket: #4062 
* Funded by: DFAT
* Description:      Only round the number when it is population exposure and it is not in debug mode


Gustry:
Thanks @myarjunar

Issue #4093, opened 2017-04-09
Earthquake on vector population should calculate fatalities


Charlotte-Morgan:

timlinux:
@Gustry so if I understand @Charlotte-Morgan correctly, we should first remove fatalities from the population and then calculate displacement.

timlinux:
To support this we need to add fatality rates as part of the definitions. We discussed this at the sprint meeting today and propose to add a new keyword to the definitions for hazard_classes:

```
    'classes': [
        {
            'key': 'X',
            'value': 10,
            'color': MMI_10,
            'name': tr('X'),
            'affected': True,
            'description':
                tr('Some well-built wooden structures destroyed; most masonry '
                   'and frame structures destroyed with foundations. '
                   'Rails bent.'),
            'string_defaults': ['extreme'],
            # function to be called at load time, takes hazard class as param
            'fatality_rate': earthquake_fatality_rate(10),
            'displacement_rate': 1.0,
            'numeric_default_min': 9.5,
            'numeric_default_max': 10.5,
            'citations': [
                {
                    'text': None,
                    'link': u'https://earthquake.usgs.gov/learn/topics/'
                            u'mercalli.php'
                }
            ]
        },
```

For each hazard we can then implement a fatality rate function - in the case of EQ that function could look at settings to see which EQ algorithm is active and computers the rate accordingly. Later when we support regionalization of PAGER (#3855) this same function could use the coeficients  to calculate regionalized displacement rates.

timlinux:
@Gustry also please be sure to add fatality rate to definitions_help.py too

Gustry:
This is going to be fixed in #4101. This PR brings two new columns fatality ratio and fatality counts in the exposure summary when you do an EQ on vector pop.
@myarjunar Can you add the fatality count in the report?

Charlotte-Morgan:
@myarjunar ... remembering that are specific rounding rules applied to the way fatalities are reported :)

Gustry:
Thanks @Charlotte-Morgan !
@myarjunar You will find the another rounding for fatalities : `fatalities_range` in the rounding file

Gustry:
The fatality count is now available in the attribute table.
@myarjunar will add the field into the report.

myarjunar:
Thanks @Gustry , but I'm not getting the 'population_fatality_ratio_field' in the attribute table, it supposed to be there right?

samnawi:
[...]

PR #4186, opened 2017-05-22
add fatality vector to EQ on pop vector

Gustry:
### What does it fix?
* Ticket: #4093
* Funded by: DFAT
* Description: Add estimated fatality with EQ on vector pop


safe/definitions/fields.py
```diff
@@ -1448,6 +1448,36 @@
...
+    'name': tr('Fatality Ratio'),
+    'field_name': 'fatality_ratio',
+    'type': QVariant.Double,
+    'length': default_field_length,
+    'precision': 10,  # I think we need some precision for this field.
+    'absolute': False,
+    'description': tr(
+        'The population fatality ratio for a given hazard class.'),
+    'help_text': tr(
+        '"Killed" is defined as: {concept} In cases where population data '
```
ismailsunni: The field name is `Fatality Ratio`, but the help text or description doesn't explain about it. It explains about `killed` people.

safe/impact_function/postprocessors.py
```diff
@@ -331,6 +336,21 @@ def enough_input(layer, post_processor_input):
...
+            elif is_keyword_value:
+                try:
+                    value = reduce(
+                        lambda d, k:
+                        d[k], input_value[
+                            'value'], layer.keywords)
+                    if value == input_value['expected_value']:
+                        break
+                    else:
+                        msg = 'Value %s is not expected in keyword: %s' % (
```
ismailsunni: Do we pass this message to the user? If so, we need to translate this.

safe/impact_function/postprocessors.py
```diff
@@ -331,6 +336,21 @@ def enough_input(layer, post_processor_input):
...
+                            input_key,
+                            input_value['value'])
+                except KeyError:
+                    msg = 'Value %s is missing in keyword: %s' % (
```
ismailsunni: Do we pass this message to the user? If so, we need to translate this also.

safe/impact_function/postprocessors.py
Gustry: No we don't, only in logs

safe/impact_function/postprocessors.py
Gustry: No we don't, Only in logs

safe/definitions/fields.py
Gustry: They are both synonyms. In `concepts.py` it's referenced as killed but we usually use fatalities. I'm going to update it, thanks

PR #4190, opened 2017-05-23
first try to add our customs expressions in QGIS

Gustry:
For fun, InaSAFE can now provide some customs expressions.

expressions

I don't think if it's the best expression to have in InaSAFE, this is just a proof of concept. Do you think about better expressions we should provide? (@Charlotte-Morgan @timlinux )

### DIFF
--- a/safe/gis/expressions.py
+++ b/safe/gis/expressions.py
@@ -1,0 +1,16 @@
+# coding=utf-8
+from qgis.core import qgsfunction
+from safe.utilities.rounding import fatalities_range
+
+__copyright__ = "Copyright 2016, The InaSAFE Project"
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = '$Format:%H$'
+
+
+@qgsfunction(
+    args='auto', group='InaSAFE', usesGeometry=False, referencedColumns=[])
+def fatality_rounding(number, feature, parent):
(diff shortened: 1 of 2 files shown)


safe/gis/expressions.py
```diff
@@ -0,0 +1,16 @@
...
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = '$Format:%H$'
+
+
+@qgsfunction(
+    args='auto', group='InaSAFE', usesGeometry=False, referencedColumns=[])
+def fatality_rounding(number, feature, parent):
+    """Round population fatalities according to InaSAFE standards."""
+    _ = (feature, parent)
```
ismailsunni: __[flake8]__
* [F841] local variable '_' is assigned to but never used

safe/plugin.py
```diff
@@ -659,6 +662,13 @@ def unload(self):
         self.dock_widget.destroy()
         self.iface.currentLayerChanged.disconnect(self.layer_changed)
 
+        # Unload QGIS expressions loaded by the plugin.
+        qgis_expressions = [
+            o for o in getmembers(expressions) if isfunction(o[1])]
```
ismailsunni: What is `o` here?

Gustry: object ? It was a copy/paste ;-)

Issue #4194, opened 2017-05-23
Options: add vulnerability groups


Charlotte-Morgan:
Sunni - pls fix what Tim forgot :)
I think the non-group fields is not required and could sections for age vulnerability ratio and then gender vulnerability ratio. 
![image](https://cloud.githubusercontent.com/assets/5743135/26344055/7989d41e-3fc8-11e7-9f39-f176ec060336.png)

ismailsunni:
Done in #4195

PR #4195, opened 2017-05-23
Fix #4195, add missing field group options.

ismailsunni:
### What does it fix?
* Ticket: #4194 
* Funded by: DFAT
* Description: 
   - Add missing field groups that are missing
screen shot 2017-05-23 at 15 16 10

cc @Charlotte-Morgan


Charlotte-Morgan:
The options look good (ie issue is fixed) but your default values look dodgy. I think your values is not an issue

ismailsunni:
@Charlotte-Morgan ah yes, I set random number for the default value :)

Issue #4200, opened 2017-05-23
TypeError when running Flood vector on Population vector


ivanbusthomi:
### Problem
When I tried running flood vector on population vector, I got this error message
```
Problem
The following problem(s) were encountered whilst running the analysis.
TypeError

Details
These additional details were reported when the problem occurred.
unhashable type: 'list'

Diagnostics
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1083, in run self._run()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1244, in _run self.gis_overlay_analysis()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1357, in gis_overlay_analysis self.hazard_preparation()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1715, in hazard_preparation self.debug_layer(self.hazard)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1036, in debug_layer check_inasafe_fields(layer)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\gis\sanity_check.py", line 33, in check_inasafe_fields difference = set(inasafe_fields.values()).difference(real_fields)
```
When I tried updating the keyword of the hazard layer used in this case, the error disappeared, and I got a result. But when I tried updating it again, the error somehow reappeared.

### Expected Behaviour
It can be run without error.

### Environment
* InaSAFE : 4.1.0b0 (develop)
* QGIS : 2.14.10
* OS : Windows
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : [Exposure.zip](https://github.com/inasafe/inasafe/files/1021725/Exposure.zip)
[...]

PR #4202, opened 2017-05-23
check if layers have some inasafe_fields are up to date

Gustry:
### What does it fix?
* Ticket: #4200
* Funded by: DFAT
* Description: Check if inasafe_fields is correct before to run the IF

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -98,6 +98,7 @@ from safe.definitions.versions import inasafe_keyword_version
 from safe.common.exceptions import (
     InaSAFEError,
     InvalidExtentError,
+    InvalidLayerError,
     WrongEarthquakeFunction,
     NoKeywordsFoundError,
     NoFeaturesInExtentError,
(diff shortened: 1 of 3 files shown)


safe/impact_function/impact_function.py
```diff
@@ -739,6 +740,24 @@ def _check_layer(layer, purpose):
...
+
+        if is_vector_layer(layer):
+            try:
+                check_inasafe_fields(layer, keywords_only=True)
+            except InvalidLayerError:
+                parameters = {
+                    'source': layer.publicSource()
+                }
+                message = generate_input_error_message(
+                    tr('The %s layer is not up to date.' % purpose),
```
ismailsunni: We should put `tr('foo bar hojah') % purpose` to make the translation works properly. It's better to use `format` also.

safe/impact_function/impact_function.py
```diff
@@ -739,6 +740,24 @@ def _check_layer(layer, purpose):
...
+                    m.Paragraph(
+                        tr('The layer {source} must be updated with the '
+                           'keyword wizard. Your fields which have been set '
+                           'in the keywords previously are not matching your '
+                           'layer.'.format(**parameters))))
```
ismailsunni: Use `format` after `tr`

PR #4206, opened 2017-05-24
Fix 4200

ismailsunni:
### What does it fix?
* Ticket: #4200
* Funded by: DFAT
* Description: 
   - Add unit test to replicate the problem
   - Add clear to make sure the state of the field mapping step is clean.

cc @samnawi @felix-yew @ivanbusthomi


samnawi:
LGTM

felix-yew:
Hi mas ismail - I get the result but show 0
image

ivanbusthomi:
Hi mas @ismailsunni - I can confirm that I got the same result as mas @felix-yew

Gustry:
For code, it's ok

ismailsunni:
@felix-yew and @ivanbusthomi the different number is caused by the type of field (Integer vs Integer64. But we still don't know yet, why. I will merge this PR since it's not related to the problem.

PR #4222, opened 2017-05-25
Fix #4210

### LINKED ISSUE #4210: Behaviour of Pregnant, Lactating, and Additional Rice
### Problem
1. For Additional Rice
Currently, If I set Pregnant and Lactating as combine parameter, I can see additional rice in the result
image
image

If I set Pregnant and Lactating separated, I cannot see additional rice in the result
image
image

2. Also if the user set the pregnant and lactating (both of joined and separated), how should InaSAFE do the analysis, to calculate the additional rice? Moreover, if the user set both of them with the different sum (sum of separated P & L is not same with joined P & L). 
Currently, seems like InaSAFE calculate additional rice using joined P & L.
image
image

**Relevant Ticket**
#4069 #4092

### Expected Behaviour
Make it make sense

How about you @Charlotte-Morgan?

### Environment
* InaSAFE : 4.1
* QGIS : 2.14
* OS : Win 7 64 bit
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : Jakarta Population
  * Hazard : Flood
  * Aggregation : Jakarta Village
  * Extent : 
  
### Proposed Solution
Make it Happy

cc : @ismailsunni @timlinux @ivanbusthomi @felix-yew @fredychandra

ismailsunni:
### What does it fix?
* Ticket: #4210 
* Funded by: DFAT
* Description: 
  - Use lactating + pregnant to calculate additional rice
  - Remove lactating_and_pregnant concept.


safe/impact_function/test/test_impact_function.py
```diff
@@ -625,7 +625,7 @@ def test_scenarios(self):
             'raster_classified_on_classified_raster': False,
             'raster_classified_on_indivisible_polygons_with_grid': False,
             'raster_classified_on_line_with_grid': False,
-            'raster_classified_on_vector_population_multi_fields': False,
+            'raster_classified_on_vector_population_multi_fields': True,
```
Gustry: Can you set it back to False?

ismailsunni: ah yes, sure. Thanks :)

PR #4225, opened 2017-05-25
Small changelog update

timlinux:
### What does it fix?

Added provenance to changelog. Also indicates which EQ model is used in definitions help.

screen shot 2017-05-26 at 10 21 45 am


Gustry:
LGTM ! 3 commits for one line, well done ;-)

PR #4227, opened 2017-05-26
Add duration to impact function.

ismailsunni:
### What does it fix?
* Ticket: No ticket
* Funded by: DFAT
* Description: 
   - Add duration of impact function in the provenance


Gustry:
This is now an API break ;-)

LGTM anyway, I think it's fine

PR #4230, opened 2017-05-26
Show eq fatality model 

timlinux:
### What does it fix?

* Funded by: DFAT
* Description: Show eq fatality function in definitions help

screen shot 2017-05-26 at 11 27 04 am


Charlotte-Morgan:
LGTM.
What about a little more seeing as this is help.
"Users can select the earthquake fatality model in the InaSAFE options."

PR #4245, opened 2017-05-31
Add codecov.

ismailsunni:
### What does it fix?
* Ticket: #
* Funded by: 
* Description:


Gustry:
I don't expect it to work out-of-the-box because we are using docker.

ismailsunni:
@Gustry ah shame. Yes, I forget about that. Will check again some time later.

PR #4255, opened 2017-06-08
Fix #4248. Expression.

### LINKED ISSUE #4248: The result became zero in age breakdown when setting count field in exposure KW
### Problem
When I try to run the flood hazard raster with population vector, and setting keyword wizard of exposure then fill the age count tab in the field of infant, child, adult, and elderly, except youth field. 
image
But, I get zero result in the impact report in demographic breakdown section, except youth get the result although I have not set the count field.
image

### Expected Behaviour
The impact report should get the result when the user setting the count field in the exposure keyword.

### Environment
* InaSAFE : 4.1.0
* QGIS : 2.14.14
* OS : Win 7 Pro-64 bit/ RAM 12GB
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : I have a problem to upload the zip file :sweat: 
  * Hazard : 

### Proposed Solution
Make it happy :smile: 

### Cc : @Charlotte-Morgan @samnawi @ivanbusthomi @ismailsunni @timlinux @Gustry

ismailsunni:
### What does it fix?
* Ticket: #4248 
* Funded by: DFAT
* Description: 
   - The field name in the expression needs to be inside a double quote. But the test is also working without double quote.
screen shot 2017-06-09 at 06 50 03

### DIFF
--- a/safe/gis/vector/prepare_vector_layer.py
+++ b/safe/gis/vector/prepare_vector_layer.py
@@ -195,6 +195,7 @@ def clean_inasafe_fields(layer):
     :param layer: The layer
     :type layer: QgsVectorLayer
     """
+    fields = []
     # Exposure
     if layer.keywords['layer_purpose'] == layer_purpose_exposure['key']:
         fields = get_fields(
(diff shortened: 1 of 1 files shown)


felix-yew:
Hi @ismailsunni 
I have try this PR and set the exposure keyword without use the youth count.
image

When I use with no aggregation, it work properly
image

But, when I try with aggregation, the result became like this
image
The count of youth is not calculate when run with aggregation (that also use the ratio).

felix-yew:
Sorry @ismailsunni I change the keyword of aggregation in the last trial. I have run again the aggregation with global default ratio.
image

And, yes LGTM :+1: 
image

Issue #4249, opened 2017-06-08
Replace EQ workflow with standard vector analysis


timlinux:

Gustry:
@ismailsunni We need to remove the "allow resample" checkbox in the KW wizard. Thanks

ismailsunni:
See my PR for removing `allow resample` in #4273

Gustry:
All done !

PR #4257, opened 2017-06-09
use the main workflow for EQ on population analysis

Gustry:
### What does it fix?
* Ticket: #4249 
* Funded by: DFAT
* Description: use the main workflow for EQ on population analysis

Work in progress


Gustry:
@timlinux @Charlotte-Morgan So what do we do with the 'allow resample' option in population keywords? Should we remove it? Or should we keep the algorithm about raster alignment?

timlinux:
> @timlinux @Charlotte-Morgan So what do we do with the 'allow resample' option in population keywords? Should we remove it? Or should we keep the algorithm about raster alignment?

We can remove the allow resample and raster alignment options / logic.

Gustry:
Ok, this PR needs a review.
I have some differences in results (total number of people) using the EQ function in InaSAFE 4.1 and this new one in 4.2. I can check deeper. I will appreciate feedback from @felix-yew about this new way to compute an EQ analysis.

Gustry:
I'm going to merge it. I will do some more testing. At least we are now using the same GIS workflow. Maybe the previous one was not correct. @felix-yew Will you do some testing?

PR #4259, opened 2017-06-10
Fix #3751. Set to use default db path from definition.

### LINKED ISSUE #3751: We have keywords.db and metadata.db, we need to remove one
### Problem

We have metadata.db : https://github.com/inasafe/inasafe/blob/develop/safe/metadata/metadata_db_io.py#L61

And keyword.db : 
https://github.com/inasafe/inasafe/blob/develop/safe/utilities/keyword_io.py#L249

### Expected Behaviour
We should remove one I think. @ismailsunni Can you confirm ?

ismailsunni:
### What does it fix?
* Ticket: #3751 
* Funded by: DFAT
* Description: 
  - Use default db path from definition
  - Clean KeywordIO

### DIFF
--- a/safe/definitions/default_settings.py
+++ b/safe/definitions/default_settings.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 """Definitions relating to default setting of fresh InaSAFE."""
-from os.path import expanduser, abspath, join
+from os.path import join
+from qgis.core import QgsApplication
 # from safe.utilities.resources import resources_path
 
 inasafe_default_settings = {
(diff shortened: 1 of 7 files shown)


Gustry:
Can it go in 4.1.1? I will check your pr later. I'm in the airport. Nearly in paris

ismailsunni:
@gustry yes, it's fine to go 4.1.1. Thanks

safe/definitions/default_settings.py
```diff
@@ -23,7 +23,7 @@
     'ISO19115_LICENSE': 'Free use with accreditation',
 
     'keywordCachePath': abspath(join(
-        expanduser('~'), '.inasafe', 'keywords.db')),
+        expanduser('~'), 'inasafe', 'metadata.db')),
```
Gustry: Why do you remove the hidden folder ?
Anyway, we shouldn't use a folder in the user directory. We should use the QGIS user directory

ismailsunni: Ah yes, thanks. I copy and paste from the metadata. But it seems I pasted the wrong one.

Gustry: Use `QgsApplication.qgisSettingsDirPath()`

Gustry:
LGTM thanks
it fixes https://github.com/inasafe/inasafe/issues/3214 too I think

Issue #3633, opened 2017-01-18
Create Metadata for All Analysis Output Layers


ismailsunni:
### Problem
Currently, we only add metadata to exposure_impacted_layer after the analysis. We should add to the other outputs also.

### Expected Behaviour
All output layer has metadata.

ismailsunni:
I just found that we already creating metadata for all the output layers even though there is only `ExposureSummaryLayerMetadata` class. Do you know something about this @Gustry ?

Gustry:
Yes, it's creating an XML file, using the base class I think. You know the metadata engine better than me.

ismailsunni:
Fixed in #4265

PR #4265, opened 2017-06-13
Fix 3633

ismailsunni:
### What does it fix?
* Ticket: #3633
* Funded by: DFAT
* Description: 
  - Use explicit mapping between layer purpose and metadata class
  - Rename output layer metadata
  - Currently, we don't have special output layer metadata, so all output layer has the same mapping. If in the future we have a special one, we just need to inherit from `OutputLayerMetadata` class, and update the mapping.

### DIFF
--- a/safe/metadata/__init__.py
+++ b/safe/metadata/__init__.py
@@ -5,8 +5,8 @@
 # pylint: disable=unused-import
 from safe.metadata.base_metadata import BaseMetadata
 from safe.metadata.generic_layer_metadata import GenericLayerMetadata
-from safe.metadata.exposure_summary_layer_metadata import (
-    ExposureSummaryLayerMetadata)
+from safe.metadata.output_layer_metadata import (
+    OutputLayerMetadata)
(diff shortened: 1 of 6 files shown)


safe/metadata/__init__.py
```diff
@@ -5,8 +5,8 @@
 # pylint: disable=unused-import
 from safe.metadata.base_metadata import BaseMetadata
 from safe.metadata.generic_layer_metadata import GenericLayerMetadata
-from safe.metadata.exposure_summary_layer_metadata import (
-    ExposureSummaryLayerMetadata)
+from safe.metadata.output_layer_metadata import (
```
ismailsunni: __[flake8]__
* [F401] 'safe.metadata.output_layer_metadata.OutputLayerMetadata' imported but unused

Gustry:
I'm not familiar with this code here, but I'm gonna trust you ;D LGTM

ismailsunni:
@Gustry :D it's just refactoring and make the code tidier :D Thanks

Issue #4047, opened 2017-03-10
Legend is not properly formatted on reports generated.


NyakudyaA:
### Problem
When running IF on earthquake population on census the legend is not properly formatted and hence unreadable in reports. The report is A4-potrait

![fix_layout](https://cloud.githubusercontent.com/assets/2510900/23792915/93f92a0a-0591-11e7-9b14-4aeb51555561.PNG)

### Expected Behaviour
Generate reports with proper legends regardless of report format.

### Environment
* InaSAFE : dev
* QGIS : 2.14.10
* OS :  Windows 10
* Data used (filenames and paths so as to reproduce, a project is better):  Test data from inasafe repo
  * Exposure :  `data\exposure\census.geojson`
  * Hazard :  `data\hazard\earthquake.tif`
  * Aggregation : `data\aggregation\grid_jakarta.geojson`
  * Extent : 
  
Checklist (if applicable): 
* [ ] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * The data is available in the inasafe repo.

### Proposed Solution
Fix legend layout especially when the legend contains multiple objects.

lucernae:
Nice! Mas @ivanbusthomi , I think we should make some space for the landscape. Maybe it's caused by big affected numbers in the legend.

ivanbusthomi:
Hi mas @lucernae and @NyakudyaA 
I don't remember splitting it into three columns. I will check it quickly

timlinux:
I still think we should not show mmi <= 5 and just put a small note saying they are excluded

lucernae:
At the moment, legend is coming directly from QGIS Layer. So, if we want to change that, we have to decide does it affect qgis layer legend too.

ivanbusthomi:
So our option for this issue are either
* hide MMI below 5, and add some notes in the report too. Or,
* use only two columns in the legend.

In my oppinion, whichever is fine as long as it is happy :smile: 
CMIIW
___
@Charlotte-Morgan @samnawi @felix-yew

lrpower:
This is also happening for the Cyclone IF A4 portrait output.
I think the legend could be condensed to one column.
![image](https://cloud.githubusercontent.com/assets/10228025/24382724/f638153a-13a3-11e7-9357-dc118ec5904b.png)

timlinux:
Maybe a good solution for now is to use layer filtering in the map composer:

[...]

PR #4276, opened 2017-06-22
update map template with filtered legend

myarjunar:
### What does it fix?
* Ticket: #3998 #4047 
* Funded by: DFAT
* Description: Only shoe legend with the feature that only appears in the map

selection_097

selection_098


timlinux:
LGTM thanks

Issue #4277, opened 2017-06-22
Blocking bug with the FR translation 


severinmenard:
Seems there is a typo in the FR translation that blocks the load of InaSAFE when starting QGIS. It provides the error msg below. Same blocking isue if you try to install InaSAFE from QGIS with locale set in French. It works fine if locale forced in US English

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe8' in position 5: ordinal not in range(128) 
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 306, in startPlugin
    plugins[packageName] = package.classFactory(iface)
  File "/home/severin/.qgis2/python/plugins/inasafe/__init__.py", line 86, in classFactory
    from safe.plugin import Plugin
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 572, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/severin/.qgis2/python/plugins/inasafe/safe/__init__.py", line 28, in 
    from safe.common.custom_logging import setup_logger
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 572, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/severin/.qgis2/python/plugins/inasafe/safe/__init__.py", line 28, in 
    from safe.common.custom_logging import setup_logger
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 572, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/severin/.qgis2/python/plugins/inasafe/safe/common/custom_logging.py", line 31, in 
    from safe.common.utilities import log_file_path
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 572, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/severin/.qgis2/python/plugins/inasafe/safe/common/utilities.py", line 25, in 
    from safe.utilities.rounding import (
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 572, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/home/severin/.qgis2/python/plugins/inasafe/safe/utilities/rounding.py", line 6, in 
    from safe.definitions.units import unit_mapping
  File "/usr/lib/python2.7/dist-packages/qgis/utils.py", line 572, in _import
[...]

PR #4278, opened 2017-06-22
fix translations in French about unicode

Gustry:
### What does it fix?
* Ticket: #4277
* Funded by: World Bank
* Description: Fix unicode exception when InaSAFE is starting.

@ismailsunni @timlinux I also remove some old test. I think we could never update the language in a test. QGIS might needs to restart, no ?


myarjunar:
> Same failing test @ismailsunni

@Gustry you can rebase to develop and it supposed to be green now

Gustry:
Thanks @myarjunar 

Quick note : @inasafe/developers Please don't put the `format` function inside the `tr`. Translations won't work because it's gonna be a new string with the content of the `format` function.

PR #4279, opened 2017-06-22
rename test_earthquake_report to test_impact_report_earthquake

myarjunar:
### What does it fix?
* Ticket: 
* Funded by: DFAT 
* Description: we change the filename so this test will run after test_impact_report because if not, this eq report test will break test_minimum_needs_ouputs. we can find the better way to fix this later.

### DIFF
--- a/safe/report/test/test_impact_report_earthquake.py
+++ b/safe/report/test/test_impact_report_earthquake.py
@@ -30,6 +30,10 @@ __email__ = "info@inasafe.org"
 __revision__ = '$Format:%H$'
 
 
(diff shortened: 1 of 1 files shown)


myarjunar:
I will merge it to make develop green

Gustry:
Hum, I didn't follow the problem, I just heard it this morning quickly. This hack looks weird !
Thanks anyway @myarjunar!

myarjunar:
@Gustry I updated the description so you can follow the problem. That's what we've found so far :sweat_smile:

Issue #4271, opened 2017-06-19
'QgsComposerItem' object has no attribute 'setKeepLayerSet'


mbernasocchi:
### Problem
```
composer_map = composition.getComposerItemById(item_id)
    """:type: qgis.core.QgsComposerMap"""
    if composer_map:```
```
this assumes getComposerItemById returns a QgsComposerMap but in my case it returns a QgsComposerItem. why do we make this assumption? Am I missing something?

```
In file "/home/marco/.qgis2/python/plugins/inasafe/safe/report/impact_report.py", line 763, in process_components output = _renderer(self, component)
In file "/home/marco/.qgis2/python/plugins/inasafe/safe/report/processors/default.py", line 403, in qgis_composer_renderer composer_map.setKeepLayerSet(True)
```

### Environment
* InaSAFE :  4.1.1
* QGIS :  2.18.9
* OS : ubuntu 17.4
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : 
  * Hazard : 
  * Aggregation :
  * Extent : 
  
Checklist (if applicable): 
* [ ] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ?

Gustry:
Is-this fixed?
I know you did a PR to avoid the python exception, but does it fix the original problem?

Thanks

mbernasocchi:
nope. 
his does not fix the problem which is that getComposerItemById returns
QgsComposerItem and we erroneously expect QgsComposerMap and
QgsComposerLegend.

can't find the comment now, but @lucernae  mentioned it was a bug in qgis api?

lucernae:
I mentioned it on the Pull Request #4280 

The problem comes from Python Bindings from getComposerItemById will always return QgsComposerItem for any (or maybe some) element because it was not type casted in the bindings, so python code doesn't know the original type.
I don't know if it's possible to cast the type from python side. If that's not possible then perhaps we have to fix the binding?

Gustry:
Ok, thanks for your explanations. I will merge the backport to the master branch, but we keep this ticket. We need to fix this as 2.18 is the next LTR in 3 months.

Gustry:
I put the label `Release blocker`, because even if it might be an upstream bug, it's a high priority for us.
I have the same issue with pictures, no more pictures in my report:
```
        image = composition.getComposerItemById(item_id)
[...]

PR #4280, opened 2017-06-26
Protect from issue #4271

mbernasocchi:
### What does it fix?
* Addresses: #4271  
* Funded by: WB GFDRR
* Description: This commit protects from having issue 4271 to happen.
This does not fix the problem which is that getComposerItemById returns
QgsComposerItem and we erroneously expect QgsComposerMap and
QgsComposerLegend.

### DIFF
--- a/safe/report/processors/default.py
+++ b/safe/report/processors/default.py
@@ -25,7 +25,10 @@ from qgis.core import (
     QgsComposerHtml,
     QgsRectangle,
     QgsLegendRenderer,
-    QgsComposerLegendStyle)
+    QgsComposerLegendStyle,
+    QgsComposerMap,
+    QgsComposerLegend
+    )
 
 from safe.common.exceptions import TemplateLoadingError
 from safe.common.utilities import temp_dir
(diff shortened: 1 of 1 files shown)


lucernae:
Hi @mbernasocchi , by any chance were you using QGIS 2.18?
I've run into this problem before and that's because some wrong binding in QGIS 2.18. All getComposerItemById will return QgsComposerItem, even if it should have been casted first.
In my case, I tried to take QgsComposerHtml, but got QgsComposerItem instead, even if it should return that.

This PR is good, but probably still won't solve the real problem. Because we can't cast QgsComposerItem to the class that we need (and what it should be). For now we only support QGIS 2.14 LTR version, and @timlinux probably want to skip 2.18 and go straight to the next LTR version.

Anyway, LGTM.

mbernasocchi:
@lucernae yes in 2.18. as you say, it's not a fix, but at least we don't get the error.

mbernasocchi:
btw, 2.18 is the next LTR https://www.qgis.org/en/site/getinvolved/development/roadmap.html

Gustry:
Yes, we will support 2.18 when 3.0 will be out. For now, we focus on 2.14, but I agree, we shouldn't have exceptions in 2.18 too as it's one of the stable qgis version.

mbernasocchi:
this was included in 9df8e515a3a7642fbf1c66c6960710646ab1d9a2

Gustry:
Better to keep one commit when fixing bugs, because we can cherry-pick the commit. We need to backport this patch on the master branch for the 4.1.2 release.

Gustry:
This is backported in the PR above

PR #4295, opened 2017-06-29
do not refresh the IF we are using an aggregation layer

Gustry:
### What does it fix?
* Funded by: WB
* Description: do not refresh the IF we are using an aggregation layer

It seems I got some problems by trying to connect/disconnect. I just found that I could use one function always connected to pan/zoom. It makes it much easier now.

stop_refresh

### DIFF
--- a/metadata.txt
+++ b/metadata.txt
@@ -26,6 +26,7 @@ status=final
 
 changelog=
     Version 4.1.2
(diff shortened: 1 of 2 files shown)


timlinux:
Thank you very much for this @Gustry

Issue #4127, opened 2017-05-05
Investigate Rule for Ratio vs Fields vs Global Default


ismailsunni:
### Problem
Investigate if these rule are valid:

#### Rules

1. Population raster with no aggregation will use global defaults for demographics (based on ratios)
2. Population raster with aggregation will use aggregation defaults for demographics (based on ratios), falling back to globals if set in agg keywords
3. Population vector with no aggregation will use exposure keywords based on field mappings or global defautls based on ratios
4. Population vector with vector aggregation will use aggregation defaults for demographics (based on ratios), falling back to globals if set in agg keywords

cc @timlinux @Gustry

Gustry:
Read the IF logs, it's quite detailed inside about what is used, falling back default ...
Maybe add new logs with the new demographic system. We should be able to understand what is going on by reading the log.

ismailsunni:
Here is my investigation's result @Gustry @timlinux 
I made more detail cases here.

<img width="689" alt="screen shot 2017-05-08 at 14 21 25" src="https://cloud.githubusercontent.com/assets/1421861/25794111/bed3ac64-33f9-11e7-8de2-3a334ded0f73.png">


It seems everyt cases make sense. Beside...

The yellow one and red one is the one that I thought not consistent.

The red one makes sense, since we specify that we don't use the ratio/count.
The yellow one is also  fine, since we agree that we will run the post processor for raster population if there is no aggregation layer.

So, all is good.

Charlotte-Morgan:
@samnawi & I need to provide comment on this

timlinux:
@ismailsunni can you attach the original spreadsheet above here? Thanks

ismailsunni:
@timlinux here is the link https://docs.google.com/spreadsheets/d/1jARdLW6cxEBk0wiUV0MJrlYVvuLseuS2zeM8prRn1dc/edit#gid=0

samnawi:
Just try to link with #3971

felix-yew:
Hi mas @ismailsunni - I wanna ask about one of case from your investigation's result. It is about the yellow one, if I run the flood raster with population raster and aggregation vector (ratio not set), so from your investigation maybe (the yellow one means) it is not be working or there maybe can give a result? or give the result but no calculations produced in detailed report? because I run it with batch runner and get the result but there are no calculations produced in the detailed report. 
[impact-report-output.pdf](https://github.com/inasafe/inasafe/files/1009781/impact-report-output.pdf)

[...]

PR #4315, opened 2017-07-17
Change do not use to do not report. #4127

ismailsunni:
### What does it fix?
* Ticket: #4127 
* Funded by: DFAT
* Description: 
  - Per @Charlotte-Morgan 's comment, change `do not use` to `do not report`.


Gustry:
Is-it written in keywords? I think for now it's just not present in keywords, isn't it ?

ismailsunni:
@Gustry 
No, it just give more proper label from `do not use` to `do not report`. Since, it actually run the post processor although we set it to `do not use` but we don't put it in the reporting.

Issue #4179, opened 2017-05-18
Documentation plan for QGIS 4.2


timlinux:
### Problem

We need to provide technical documentation to make it easier for new developers to join the project and more importantly for power users to be able to understand the system from a higher level view. So I have identified a number of areas that should be nicely documented (with accompanying diagrams):

### Proposed Solution

**Processing workflow:** @Gustry 

* **Input Layers:** What input layers are supported, what constraints are in place before they can be used? Any other special notes users should be aware of.
* **Options:** where are options used during processing? Any other special notes beyond the descriptive docs I have already done for the options dialog?
* **Workflow planner:** - How is the decision made to run the 'special' earthquake raster on raster workflow?
* **Earthquake:** - Why do we have a separate workflow for this and what are the differences in this workflow compared to the mainstream workflow?
* **Exposure preparation:** - What happens in the  preparation phase? Any special notes? Also we should make it clear that the operation is 'file to file' and that input layers should remain unchanged.
* **Aggregation preparation:** -  What happens in the  preparation phase? Any special notes? Also we should make it clear that the operation is 'file to file' and that input layers should remain unchanged.
* **Hazard preparation:** -  What happens in the  preparation phase? Any special notes? Also we should make it clear that the operation is 'file to file' and that input layers should remain unchanged.
* **Aggregate Hazard:** - We should explain (with some diagrams) what aggregate hazard is and any special notes that power users should be aware of during this step.

* **Post processors:** - This needs descriptive documentation (what are post-processors?) but it should also provide details that a developer might use because this is likely to be an extension point for InaSAFE that organizations want to tap into. We should write a simple tutorial showing how to add a new post-processor and showing how the generated output appears in the impact tables.

* **Summary calculation:** - Descriptive documentation. Are we thinking people will likely be wanting to extend this for customizing InaSAFE outputs? If so we should also provide a little extension tutorial.

**Reporting:** @myarjunar 

* **General architecture:** - We should describe adequately the general

ismailsunni:
Do we need to document the keywords also?

timlinux:
@ismailsunni yes we should definitely document the keywords IMHO

timlinux:
TODO @timlinux:

- [ ] add keywords to the definitions help
- [ ] add IF centric wizard help to definitions help
- [ ] add kw wizard to the definitions help

Gustry:
Pre-processors

PR #4319, opened 2017-07-20
Adding definition of every keyword's property

ismailsunni:
### What does it fix?
* Ticket: #4179 
* Funded by: DFAT
* Description: 
  - Adding definition for keyword's properties

### DIFF
--- a/safe/definitions/__init__.py
+++ b/safe/definitions/__init__.py
@@ -19,6 +19,7 @@ from safe.definitions.field_groups import *
 from safe.definitions.hazard import *
 from safe.definitions.hazard_category import *
 from safe.definitions.hazard_classifications import *
+from safe.definitions.keyword_properties import *
 from safe.definitions.layer_geometry import *
 from safe.definitions.layer_modes import *
 from safe.definitions.layer_purposes import *
--- a/safe/definitions/keyword_properties.py
+++ b/safe/definitions/keyword_properties.py
@@ -1,0 +1,314 @@
+# coding=utf-8
+
+"""Definitions relating to layer keywords."""
+
+from safe.utilities.i18n import tr
+
+__copyright__ = "Copyright 2016, The InaSAFE Project"
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = '$Format:%H$'
+
+
+# Base Metadata Property
+property_organisation = {
+    'key': 'property_organisation',
+    'name': tr('Organisation'),
+    'description': tr('An organized body of people who own the layer.')
+}
+
+property_email = {
+    'key': 'property_email',
+    'name': tr('Organisation'),
+    'description': tr('The email address of the author of the layer.')
+}
+
+property_date = {
+    'key': 'property_date',
+    'name': tr('Date'),
+    'description': tr('The date when the layer is created.')
+}
+
+property_abstract = {
+    'key': 'property_abstract',
+    'name': tr('Abstract'),
+    'description': tr(
+        'A brief narrative summary of the content of the layer.')
+}
+
+property_title = {
+    'key': 'property_title',
+    'name': tr('Title'),
+    'description': tr('A name of the layer.')
+}
+
+property_license = {
+    'key': 'property_license',
+    'name': tr('License'),
+    'description': tr('A permit from an authority to use the layer.')
+}
+
+property_url = {
+    'key': 'property_url',
+    'name': tr('URL'),
+    'description': tr(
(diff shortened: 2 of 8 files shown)


safe/definitions/__init__.py
```diff
@@ -19,6 +19,7 @@
 from safe.definitions.hazard import *
 from safe.definitions.hazard_category import *
 from safe.definitions.hazard_classifications import *
+from safe.definitions.keyword_properties import *
```
ismailsunni: __[flake8]__
* [F403] 'from safe.definitions.keyword_properties import *' used; unable to detect undefined names
* [F401] 'safe.definitions.keyword_properties.*' imported but unused

Gustry:
Where do you use all these new items in `keyword_properties.py`? I saw only one use for the layer_purpose.
WIll you display a table in the inasafe help window?

ismailsunni:
Yes, this is a part of #4179, starting to write the docs.

> WIll you display a table in the inasafe help window?
Yes, that's the plan from @timlinux 

After working on this PR, I am also thinking moving the properties from metadata module to definition module. So that it's easier to manage in one  place. Then, the `keyword_properties.py` will be used in the metadata module.

Gustry:
Ok thanks. I forgot about this ticket.
Is this PR finished?
So you created these new entries in definitions without using them yet?

ismailsunni:
Yes, I will do it step by step. So, I think it's better to merge it for now, at least we can start translate it.
I will need to check how the help dialog from @timlinux works, since I will need to add this on it.

safe/definitions/keyword_properties.py
```diff
@@ -0,0 +1,314 @@
...
+    'description': tr(
+        'The address of World Web Page where we can find the layer or its '
+        'description.')
+}
+
+property_layer_purpose = {
+    'key': 'property_layer_purpose',
+    'name': tr('Layer Purpose'),
+    'description': tr(
+        'The purpose of the layer, it can be hazard layer, exposure layer, or '
```
Gustry: `it can be hazard layer, exposure layer or aggregation layer`. No need for a comma if you already have `or`.

safe/definitions/keyword_properties.py
```diff
@@ -0,0 +1,314 @@
...
+    'name': tr('Layer Mode'),
+    'description': tr(
+        'The mode of the layer, it can be continuous or classified layer.')
+}
+
+property_layer_geometry = {
+    'key': 'property_layer_geometry',
+    'name': tr('Layer Geometry'),
+    'description': tr(
+        'The geometry type of the layer, it can be point, line, polygon, '
```
Gustry: Same here.

safe/definitions/keyword_properties.py
```diff
@@ -0,0 +1,314 @@
...
+property_scale = {
+    'key': 'property_scale',
+    'name': tr('Scale'),
+    'description': tr('The default scale of the layer.')
+}
+
+property_source = {
+    'key': 'property_source',
+    'name': tr('Source'),
+    'description': tr('The location of where does the layer come from.')
```
Gustry: come**s** from?

safe/definitions/keyword_properties.py
```diff
@@ -0,0 +1,314 @@
...
+    #     }
+    # }
+}
+
+# Output Layer Metadata Property
+property_exposure_keywords = {
+    'key': 'property_exposure_keywords',
+    'name': tr('Exposure Keywords'),
+    'description': tr(
+        'A copy of original exposure keywords of the output\'s analysis.'
```
Gustry: Is it not `in` the output?
Same for below about hazard and aggregation

safe/definitions/keyword_properties.py
ismailsunni: Actually, we need it. It's oxford comma. https://english.stackexchange.com/questions/267256/oxford-comma-before-or

Gustry: Ok, I didn't know. I think learned no comma if you have `or` or `and` in French.

ismailsunni: :D we use comma in Bahasa Indonesia, if you have more than 2 objects (like in English)

safe/definitions/keyword_properties.py
ismailsunni: ah yes, thanks.

safe/definitions/keyword_properties.py
ismailsunni: Fixed.

Issue #4303, opened 2017-07-07
Feature: Templating review in InaSAFE


timlinux:
### Problem

We need a more flexible printing system in InaSAFE. I've been thinking about this a lot and here is what I propose:

1) **Support atlas prints**: Currently we can use an atlas map layout but the call to generate the atlas is different to the print call from a non atlas map.

**Outcome:** When the user has created a template that has atlas enabled, we should render the atlas to PDF.

2) **We need a system for dynamically changing the datasources in a composition**: We want to do infographics etc. from the composer, but we need to be able to reference specific tables in the map project which we will not know at design time.

**Outcome:** If a user references an InaSAFE impact layer, in their template, the reference should be maintained based on the current impact function outputs.

3_ **We need a system to reference specific templates based on the combination of hazard and exposure**: When a user is doing e.g. flood on population, he or she should be able to use a template (which e.g. includes infographic elements) specific to that hazard and exposure combination. We should have a fall back that if no specific template is defined for the hazard / exposure combination, it uses our default template (e.g. like the current A4 template we have).

**Outcome:** Reports are produced with contextual awareness of the hazard and exposure data.

4. **We need to rename the default template:** Lets use rather a generic name like "inasafe-generic-template.gpt"

**Outcome:** the template name will still be applicable regardless of color scheme or page size.

5. **We should just produce one map.** This one is more debatable, but why do we produce two maps  I think one is enough. @Charlotte-Morgan and team to comment on that.

**Outcome:** More ready to understand outputs

6. **We should support use of QGIS variables.** This can be a simple way to override aspects of the map e.g. the accent color without doing any template editing or programming.

**Outcome:** We have a way to make quick tweaks to the map.

CC @Charlotte-Morgan @ismailsunni @myarjunar

lucernae:
@timlinux I'm also interested to see where the discussion goes.
Is it going to be implemented by improving our current report package or creating another one?
If it is still using current report system, I might be able to share some comments.
It also might help us with GeoSAFE.

1. I'm still not familiar with atlas. Perhaps we only need a code branching to invoke atlas print?

[...]

PR #4325, opened 2017-07-24
Provenance data to project variables

ismailsunni:
### What does it fix?
* Ticket: #4303
* Funded by: DFAT
* Description: 
  - When user click output layer from analysis that has provenance, InaSAFE will set the value of the provenance in the project variables

screen shot 2017-07-25 at 10 31 28

cc @timlinux


safe/gui/widgets/dock.py
```diff
@@ -1396,3 +1406,29 @@ def _validate_question_area(self):
...
+
+    def set_provenance_to_project_variables(self, provenances):
+        """Helper method to update / create provenance in project variables.
+
+        :param provenances: Keys and values from provenances.
+        :type provenances: dict
+        """
+        # QgsExpressionContextUtils.setProjectVariable(provenances)
+        for key, value in provenances.items():
+            if isinstance(value, (str, unicode, int, float)):
```
ismailsunni: __[flake8]__
* [F821] undefined name 'unicode'

safe/gui/widgets/dock.py
```diff
@@ -1396,3 +1405,44 @@ def _validate_question_area(self):
...
+    def set_provenance_to_project_variables(self, provenances):
+        """Helper method to update / create provenance in project variables.
+
+        :param provenances: Keys and values from provenances.
+        :type provenances: dict
+        """
+        # Remove old provenance data first
+        self.remove_provenance_project_variables()
+        for key, value in provenances.items():
+            if isinstance(value, (str, unicode, Number)):
```
ismailsunni: __[flake8]__
* [F821] undefined name 'unicode'

safe/definitions/__init__.py
```diff
@@ -26,6 +26,7 @@
 from safe.definitions.messages import *
 from safe.definitions.minimum_needs import *
 from safe.definitions.post_processors import *
+from safe.definitions.provenances import *
```
ismailsunni: __[flake8]__
* [F403] 'from safe.definitions.provenances import *' used; unable to detect undefined names
* [F401] 'safe.definitions.provenances.*' imported but unused

safe/gui/widgets/dock.py
```diff
@@ -57,6 +60,7 @@
     add_ordered_combo_item,
     is_keyword_version_supported,
 )
+from safe.metadata.utilities import serialize_dictionary
```
ismailsunni: __[flake8]__
* [F401] 'safe.metadata.utilities.serialize_dictionary' imported but unused

safe/gui/widgets/dock.py
```diff
@@ -1396,3 +1404,65 @@ def _validate_question_area(self):
...
+            """
+            if isinstance(value, (list, tuple, set)):
+                for element_index, element_value in enumerate(value):
+                    write_project_variable(
+                        '%s__%s' % (key, element_index), element_value)
+            elif isinstance(value, dict):
+                for dict_key, dict_value in value.items():
+                    write_project_variable(
+                        '%s__%s' % (key, dict_key), dict_value)
+            elif isinstance(value, (bool, str, unicode, Number)):
```
ismailsunni: __[flake8]__
* [F821] undefined name 'unicode'

safe/gui/widgets/dock.py
```diff
@@ -1396,3 +1403,65 @@ def _validate_question_area(self):
...
```
ismailsunni: __[flake8]__
* [F821] undefined name 'unicode'

safe/gui/widgets/dock.py
Gustry: Can't you set Python 2 in this checker? Maybe it's using python3 to check this.

safe/metadata/constants.py
```diff
@@ -0,0 +1,37 @@
+# coding=utf-8
+
+"""Metadata Constants"""
```
Gustry: Not in safe/definitions?

Gustry:
When you click outside of the group in the legend, it removes theses variables?
Why exposure_keywords and hazard_keywords are empty in your screenshot?

safe/definitions/__init__.py
timlinux: > from safe.definitions.provenances import *

s/provenances/provenance

timlinux:
Some comments from me:

Entries like 'action_checklist' should be presented as plain text string or html markup, not JSON.

Impact Function Name: Should strip out the underscores

Some things like host name, python version, pyqt, qt, qgis may already be provided in the QGIS global variables - we should avoid duplicating them in project variables when already provided.

The main things I am missing here are references to the sub products:

* analysis_layer
* aggregation_impacted layer
* etc.

Having their file system paths / URI's will enable smart users to load that data dynamically and grab entries from it. I think we might provide some custom python QgsExpressions which make convenience helpers for getting content from these layers as @Gustry and I discussed in JK on our last site visit.

Basically everything else we provide here should be nicely formatted and ready to use to place into a layout (so use nice capitalization, remove 'foreign' looking stuff like underscores, use clean, logical naming etc. 

It would be also good to see if we can upstream a patch to make it easier to read and write project variables via the QGIS python API.

Lastly it will be great if you can confirm the behaviour of what happens when you click off a layer group and when you swap between analysis groups? 

Thanks for this, great work!

safe/definitions/__init__.py
```diff
@@ -26,6 +26,7 @@
 from safe.definitions.messages import *
 from safe.definitions.minimum_needs import *
 from safe.definitions.post_processors import *
+from safe.definitions.provenance import *
```
ismailsunni: __[flake8]__
* [F403] 'from safe.definitions.provenance import *' used; unable to detect undefined names
* [F401] 'safe.definitions.provenance.*' imported but unused

safe/gui/widgets/dock.py
ismailsunni: Hmm, I have checked in the SideCI. It seems we need to set the flake8 preference.
> You can customize the behaviour and some other options by adding a `tox.ini` or `setup.cfg` file to your repository.

I will do it later (not in this PR).

ismailsunni:
Hi @Gustry 

> When you click outside of the group in the legend, it removes theses variables?

Yes, it does.

> Why exposure_keywords and hazard_keywords are empty in your screenshot?

Wrong screenshot :D, I have updated it. No longer empty.

ismailsunni:
> Entries like 'action_checklist' should be presented as plain text string or html markup, not JSON.

Wrong screenshot. It's already done in this PR. See my new screenshot above. :D

> Impact Function Name: Should strip out the underscores

Fixed.

> Some things like host name, python version, pyqt, qt, qgis may already be provided in the QGIS global variables - we should avoid duplicating them in project variables when already provided.

Ok. I will use the same key in the provenance. It's a good way to review all provenance that we have. I have fixed it also.

> The main things I am missing here are references to the sub products:
> 
> analysis_layer
> aggregation_impacted layer
> etc.
> Having their file system paths / URI's will enable smart users to load that data dynamically and grab entries from it. I think we might provide some custom python QgsExpressions which make convenience helpers for getting content from these layers as @Gustry and I discussed in JK on our last site visit.

Ok, I will need to add it to the provenance. Because previously we get the URI from resolving the path.

> Basically everything else we provide here should be nicely formatted and ready to use to place into a layout (so use nice capitalization, remove 'foreign' looking stuff like underscores, use clean, logical naming etc.

Ok. I found out we don't use the `name` from `definition` often. I will fix it also. But I will still store the computer-friendly string in the provenance (so that we can get the definition).

> It would be also good to see if we can upstream a patch to make it easier to read and write project variables via the QGIS python API.

Yeah, I am planning to do it.

> Lastly it will be great if you can confirm the behaviour of what happens when you click off a layer group and when you swap between analysis groups?

For layer group, I need to set the behaviour. Currently, it only works when we select output layer.

When we swap between analysis group, it will update the project variable directly.

> Thanks for this, great work!

Thanks

safe/gui/widgets/dock.py
```diff
@@ -1396,3 +1403,73 @@ def _validate_question_area(self):
...
```
ismailsunni: __[flake8]__
* [F821] undefined name 'unicode'

safe/gui/tools/wizard/test/test_wizard_util.py
```diff
@@ -5,7 +5,7 @@
 from safe.test.utilities import get_qgis_app
 QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
 
-from safe.gui.tools.wizard.wizard_utils import get_question_text
+from safe.gui.tools.wizard.utilities import get_question_text
```
ismailsunni: __[flake8]__
* [E402] module level import not at top of file

safe/impact_function/impact_function.py
```diff
@@ -71,6 +76,38 @@
     layer_purpose_exposure_summary_table,
     layer_purpose_profiling,
 )
+from safe.definitions.provenance import (
```
ismailsunni: __[flake8]__
* [F401] 'safe.definitions.provenance.provenance_pyqt_version' imported but unused

safe/definitions/provenance.py
```diff
@@ -0,0 +1,186 @@
...
+provenance_list = [
+    provenance_action_checklist,
+    provenance_aggregation_keywords,
+    provenance_aggregation_layer,
+    provenance_aggregation_layer_id,
+    provenance_analysis_extent,
+    provenance_analysis_question,
+    provenance_data_store_uri,
+    provenance_duration,
+    provenance_end_datetime,
```
Gustry: not possible to put `start_datetime` here in this list?

safe/metadata/property/dictionary_property.py
```diff
@@ -1,12 +1,14 @@
-# -*- coding: utf-8 -*-
+# coding=utf-8
+
+"""Dictionary property"""
```
Gustry: Full stop in docstring

safe/definitions/provenance.py
```diff
@@ -0,0 +1,186 @@
...
+    'provenance_key': 'impact_function_name'
+}
+provenance_impact_function_title = {
+    'key': 'provenance_impact_function_title',
+    'name': tr('Impact Function Title'),
+    'provenance_key': 'impact_function_title'
+}
+provenance_inasafe_version = {
+    'key': 'provenance_inasafe_version',
+    'name': tr('Inasafe Version'),
```
ismailsunni: Oops, I am planning to fix the case in this file, since I was generating it using script. :D 

Thanks for the catch :D

safe/definitions/provenance.py
ismailsunni: I already put it, it's alphabetical, so it's not next to each other.

ismailsunni:
> QGIS is already providing qgis_os_name, qgis_version for instance.

I will keep it in the metadata provenance. But, I won't save it in the project variables.

Actually, we have a case where the analysis is coming from another QGIS version, and is opened in the different QGIS version.

ismailsunni:
Hi @timlinux 
I am merging this PR first since I already make a big changes, and I want to avoid conflict later. I will make another PR to address the rest tasks:
- Add behaviour when layer group is selected
- Add more content to provenance for impact layer

PR #4329, opened 2017-07-25
fix double demographic items in help window

Gustry:
### What does it fix?
* Funded by: DFAT
* Description: fix double demographic items in help window


samnawi:
LGTM
image

Issue #4307, opened 2017-07-11
Report map only prints when project CRS if 4326


timlinux:
### Problem

Currently there is no map on the report if the project CRS is set to a projection other than 4326.

Proposed solution:

Reproject the map dimensions to the project CRS when setting up the map report.

myarjunar:
closed in #4330

PR #4330, opened 2017-07-25
transform extent's crs to mapcomposer crs

myarjunar:
### What does it fix?
* Ticket: #4307 
* Funded by: DFAT
* Description: Fix map report not showing if CRS is other than EPSG:4326.


Gustry:
Please close issue if it's possible

Issue #3726, opened 2017-01-30
InaSAFE 4.0 Analysis terminology and layer names


Charlotte-Morgan:
### Problem
InaSAFE 4.0 moves away from the traditional concept of an Impact Layer and produces up to 5 data sets from the analysis. However, the terminology introduced for these data sets is not clear and users are likely to be confused especially with the use of the word aggregate to mean many things.

I find it very hard to understand what is being described, and I am an experience GIS user, understand spatial analysis and speak English.
@Gustry @timlinux - good chance I have misunderstood some things as I still do not have working examples of all analyses, but have had a go below at providing some suggestions

### Expected Behaviour
Users find it easy to understand the analysis products, the relationship between them, how and when to use them. The descriptions are simple, consistent in their construction and easy to teach, translate and reinforce.

### Proposed Solution
Terminology should be consistent and logical; and supported by trainers and training materials.
In needs to clearly separate inputs from outputs, and spatial data from tables. Where possible, it should use conventional spatial analysis  terminology to describe processes and products. 

The word aggregation is used to refer only to the aggregation layer
The word summary be used to describe analysis results that are grouped in some way; ie by hazard type, by exposure type, by aggregation area, by age , gender etc

### Current Status - according to InaSAFE help - with proposed changes
12 Layer Purposes
12.1 Hazard :)
12.2 Exposure :)
12.3 Aggregation :)
These all have the definition for each but do not explain that these are inputs to an InaSAFE analysis. 

Proposal: include more information to say these are inputs and are required or optional. How analysis runs if aggregation is not specified (more this from the analysis results area) - or a separate section on how InaSAFE works for specific use cases 

**12.4 Exposure Impacted** (Change this to Impact Layer- Exposure Summary)
Proposed wording: This impact layer contains all the results for the spatial analysis of the hazard, exposure and aggregation layers (if used) within the analysis extent. This layer is not created if the input includes a continuous raster exposure layer (eg population raster). (what is made in this case?)

**12.5 Exposure Breakdown** (Change this to Exposure Summary Table (its not a layer its a table))
Proposed wording: This impact table contains the analysis results for exposure type by hazard type, summarised by exposure type. It includes totals for affected and not affected status. It is used to generate reports and can be exported to a spreadsheet for further analysis.
Proposed change in order - move this down so the order here reflects the order displayed in the output folder for users.

**12.6 Aggregation Impacted** (Change this to Impact Layer - Aggregation Summary)
[...]

PR #4331, opened 2017-07-25
update terminology about output layers

Gustry:
### What does it fix?
* Ticket: #3726 
* Funded by: DFAT
* Description: update terminology about output layers

* [x] Check the ticket is correct 
* [x] Update filenames too

### DIFF
--- a/safe/definitions/layer_purposes.py
+++ b/safe/definitions/layer_purposes.py
@@ -63,7 +63,7 @@ layer_purpose_aggregation = {
 }
 
 layer_purpose_exposure_summary = {
-    'key': 'exposure_summary',
+    'key': 'impact_analysis',
     'name': tr('Impact Analysis'),
     'description': tr(
         'This <b>Impact Analysis</b> contains all the '
@@ -82,7 +82,7 @@ layer_purpose_exposure_summary = {
 }
 
 layer_purpose_aggregate_hazard_impacted = {
-    'key': 'aggregate_hazard_impacted',
+    'key': 'hazard_aggregation_summary',
     'name': tr('Hazard Aggregation Summary'),
     'description': tr(
         'An <b>hazard aggregation summary</b> is created during an '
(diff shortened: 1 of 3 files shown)


ismailsunni:
Let me know if you have finished this. I need it for the provenance. :)

Gustry:
@ismailsunni do you think my last commit is OK? I'm updating keys in definitions.
I'm waiting for travis to have a look.

ismailsunni:
Oops, sorry, I missed this PR because the label is `In Progress`.
LGTM. Just need to fix the travis.

Gustry:
So no problem to update the key? I'm afraid of side effect. Especially with your new variables where we will use the layer purpose to fetch the good layer. But hopefully we do this change both at the same time in 4.2
I was thinking about retro-compatbility, if you open an old analysis 4.1 in 4.2

ismailsunni:
@Gustry 
What about just updating the name, but not the key. The key is invisible from the use. We also use varialble to access the key, so I think it should be fine if we do not update the key.

Gustry:
Because the key is used for the filename:
screen shot 2017-07-27 at 10 05 25

And we wanted to update filename too, to remove the `aggregate_hazard_impacted` for instance.

ismailsunni:
Ouch. I didn't aware that it was being used to name the file. If we need a retro-compability, we need to do name migration or something like that.

What about putting the filename in the layerpurpose definition. As a list. It will have new key and old key. If the file based on the new key is not available, check the old key. Not sure though

Gustry:
Why do you want a name migration?
Did you see my latest commit in dock.py ? I know it's like a hack, but I think there is only this.

ismailsunni:
Ah I see. Sorry, just re-read the whole code in that function again.
Yes, I think it's fine, we only use the key of layer_purpose of output layer to show the impact report, right?

Gustry:
Travis should be happy now.

I would like to merge this as soon as possible today. I understood you agreed my PR if I read between lines ;-)
Still wondering if I won't have any side effects of this, but I think it should be fine. I think I will merge later today, but I will appreciate a review.

Gustry:
Sorry, I re-read the thread, I saw the LGTM already, after my key update. Let's merge this so I can translate now

Issue #4116, opened 2017-04-27
Invalid Layer Error on Debug Mode


ismailsunni:
### Problem
I got this issue first time from @samnawi in his comment https://github.com/inasafe/inasafe/pull/4114#issuecomment-297611120 

The error was like this:

```
Problem

The following problem(s) were encountered whilst running the analysis.

InvalidLayerError
Suggestion

You can try the following to resolve the issue:

The layer should not have many sublayers : /var/folders/h7/k16m33695_g2ptlzy0z8vxb00000gp/T/inasafe/2017-02-23/ismailsunni/impacts/FloodRasterOnPopulationPolygon_27April2017_23h33-21.947394/7-hazard_clipped.geojson : 0:OGRGeoJSON:1:LineString;0:OGRGeoJSON:400:Polygon
Details
These additional details were reported when the problem occurred.

The layer should not have many sublayers : /var/folders/h7/k16m33695_g2ptlzy0z8vxb00000gp/T/inasafe/2017-02-23/ismailsunni/impacts/FloodRasterOnPopulationPolygon_27April2017_23h33-21.947394/7-hazard_clipped.geojson : 0:OGRGeoJSON:1:LineString;0:OGRGeoJSON:400:Polygon
Diagnostics

In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe/safe/impact_function/impact_function.py", line 1078, in run self._run()
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe/safe/utilities/profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe/safe/impact_function/impact_function.py", line 1239, in _run self.gis_overlay_analysis()
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe/safe/utilities/profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe/safe/impact_function/impact_function.py", line 1346, in gis_overlay_analysis self.hazard_preparation()
In file "/Users/ismailsunni/.qgis2/python/plugins/inasafe/safe/utilities/profiling.py", line 93, in with_profiling ret = fn(*args, **kwargs)
[...]

PR #4333, opened 2017-07-26
allow only the same geometry type in the output layer

Gustry:
### What does it fix?
* Ticket: #4116 
* Funded by: DFAT
* Description: I think these two lines will fix this ticket. But I have more to do about invalid geometries.


Gustry:
No, we won't, because the algorithm must only output the same geometry type

PR #4338, opened 2017-07-26
Fix 4303 output layer metadata

ismailsunni:
### What does it fix?
* Ticket: #4303 
* Funded by: DFAT
* Description: 
   - Adding output layer URI in the output layer metadata (e.g. aggregate_hazard_impacted in the image below)
screen shot 2017-07-26 at 22 35 12

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -899,7 +899,6 @@ class Dock(QtGui.QDockWidget, FORM_CLASS):
             ]
 
             provenances = keywords.get('provenance_data', {})
-            LOGGER.debug('Set Keywords')
             self.set_provenance_to_project_variables(provenances)
 
             show_keywords = True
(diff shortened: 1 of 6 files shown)


safe/gui/widgets/dock.py
```diff
@@ -915,7 +914,7 @@ def layer_changed(self, layer):
             if show_keywords:
                 if inasafe_keyword_version_key not in keywords.keys():
                     show_keyword_version_message(
-                        self, 'No Version', self.inasafe_version)
```
Gustry: Is this block still valid? I think this was legacy and I was wondering if this block can be reached again today

ismailsunni:
Oh no, the travis is failed, and the value to compare is depending on datetime in the output layer path.

safe/gui/widgets/dock.py
ismailsunni: It's still possible if you have corrupted metadata xml file that doesn't have `keyword version` in it. I prefer to keep this althought the probability is very small.

Gustry: ok

Gustry:
Can't you check that the expected provenance is a subset of the real provenance? IE not checking that dict are equals, but it's a subset.

ismailsunni:
> Can't you check that the expected provenance is a subset of the real provenance? IE not checking that dict are equals, but it's a subset.

Yes, I did it. I also check if the key for output layers exist

Issue #3626, opened 2017-01-18
Update Landcover Classification and Fields


ismailsunni:
### Problem
Currently, we use a very basic classifications for land cover classes. We should have a better classification and reference.
<img width="1097" alt="screen shot 2017-01-18 at 09 21 58" src="https://cloud.githubusercontent.com/assets/1421861/22048449/aea634f6-dd5f-11e6-9e58-66190024b982.png">

We also needs more field to the land cover exposure:
- [ ] Productivity Kwt/HA
- [ ] Productivity cost Million IDR/HA
- [ ] Productivity value Million IDR/HA
- [ ] Remove the feature rate

cc @fredychandra @timlinux @Charlotte-Morgan @inasafe/issue-manager

Gustry:
Fix `industrial` -> `Industrial`

fredychandra:
Hi @ismailsunni @Gustry & @Charlotte-Morgan @ivanbusthomi 

Here I try to incorporate ticket #3626 and #3947 with Land Cover Analysis feature value concept and its report. This concept feature might be relevant with elements at risk (assets) loss assessment establishment in the future. 

![lc_featurevalue](https://cloud.githubusercontent.com/assets/6345285/24735397/f152c41a-1aac-11e7-860e-da0765358c44.JPG)

If others type feature available from the data classification, eg. Crops: Kacang, Jagung, etc. We can make feature value as well because the attributes data are available on http://aplikasi.pertanian.go.id/bdsp/newlok.asp and https://bps.go.id/linkTabelStatis/view/id/1855

![lc_featurerate_formula](https://cloud.githubusercontent.com/assets/6345285/24735923/69c3d860-1aaf-11e7-97e5-70b3f355785a.JPG)


![keyword](https://cloud.githubusercontent.com/assets/6345285/24735958/988325f2-1aaf-11e7-9e88-deacd32de6e8.JPG)

**The Result (REPORT) :** 
Add approximate of potential loss on crops below estimated number of land cover affected by flood result (report).

![anlysis result](https://cloud.githubusercontent.com/assets/6345285/24736042/036a873e-1ab0-11e7-825c-ec0f35ac898b.JPG)

The calculating result of affected area of Padi field can be compared with total harvest area in one District (Certain area). It can get percentage of productive area that affected by flood (certain relevant hazard/event)

Happy to discuss.

fredychandra:
Hi @ismailsunni  
[...]

PR #4341, opened 2017-07-27
Do not report exposure type

ismailsunni:
### What does it fix?
* Ticket: #3626 
* Funded by: DFAT
* Description: 
  - Add new class for `Do Not Report` that won't be included in the calculation. You can assign it on keyword wizard.

screen shot 2017-07-27 at 11 53 56

screen shot 2017-07-27 at 10 25 50
screen shot 2017-07-27 at 10 26 02

The number above is not added up because of the rounding.

cc @fredychandra

### DIFF
--- a/safe/definitions/exposure_classifications.py
+++ b/safe/definitions/exposure_classifications.py
@@ -671,6 +671,20 @@ generic_landcover_classes = {
                     'link': None
                 }
             ]
+        },
+        {
+            'key': 'do_not_report',
+            'name': tr('Do Not Report'),
+            'description': tr(
+                'Land use types that will be excluded in the analysis.'),
+            'osm_downloader': [],
+            'string_defaults': [],
+            'citations': [
+                {
+                    'text': None,
+                    'link': None
+                }
+            ]
         }
     ]
 }
--- a/safe/definitions/exposure_classifications.py
+++ b/safe/definitions/exposure_classifications.py
@@ -3,6 +3,7 @@
 """Definitions relating to exposure classifications."""
 
 from safe.utilities.i18n import tr
+from safe.definitions.constants import DO_NOT_REPORT
 
 __copyright__ = "Copyright 2016, The InaSAFE Project"
 __license__ = "GPL version 3"
@@ -28,6 +29,21 @@ exposure_classification_type = tr('Exposure Classification'),
 # The osm_downloader is not used in InaSAFE itself. It is used in OSM-Reporter
 # to generate the value_mapping. Please do not remove it.
 # See https://github.com/kartoza/osm-reporter/wiki how to use this key.
+
+do_not_report_class = {
+    'key': DO_NOT_REPORT,
+    'name': tr('Do Not Report'),
+    'description': tr(
+        'Land use types that will be excluded in the analysis.'),
+    'osm_downloader': [],
+    'string_defaults': [],
+    'citations': [
+        {
+            'text': None,
+            'link': None
+        }
+    ]
+}
 
 generic_structure_classes = {
     'key': 'generic_structure_classes',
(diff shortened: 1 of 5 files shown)


safe/gis/vector/prepare_vector_layer.py
```diff
@@ -358,6 +358,16 @@ def _remove_features(layer):
                 # i += 1
                 pass
 
+            # If there is exposure that mapped to do not report class
+            if layer_purpose == layer_purpose_exposure['key']:
+                is_do_not_report = layer.keywords['value_map'].get(
+                    'do_not_report')
+                if is_do_not_report:
+                    if feature[field_name] in layer.keywords['value_map'][
+                        'do_not_report']:
```
ismailsunni: __[flake8]__
* [E125] continuation line with same indent as next logical line

safe/definitions/exposure_classifications.py
```diff
@@ -671,6 +671,20 @@
                     'link': None
                 }
             ]
+        },
+        {
+            'key': 'do_not_report',
```
Gustry: can you make this key from a constant : `'key': DO_NOT_REPORT,` ?

safe/gis/vector/prepare_vector_layer.py
```diff
@@ -358,6 +358,16 @@ def _remove_features(layer):
                 # i += 1
                 pass
 
+            # If there is exposure that mapped to do not report class
+            if layer_purpose == layer_purpose_exposure['key']:
+                is_do_not_report = layer.keywords.get('value_map', {}).get(
+                    'do_not_report')
```
Gustry: And use this constant here ? `DO_NOT_REPORT` ?
Like this, all our strings are in definitions

Gustry:
Hum, on thing, there is a rule : the `Other` group is always the last one in the classification. Features without classification will go to the `Other` group. But now, if you do that, it will go to`Do Not Report`. I think you need to move it.

ismailsunni:
Yeah, I was expecting it would be difficult. But, it turns out quite easy to add. Thanks

I will move the `Other` to last, and make constant for Do No Report.

samnawi:
Hi @ismailsunni I'm worried if we use the ways like this, it will disappear one of the essences of the landcover analysis.
image
Besides the value of the crop, landcover analysis also generates the affected area for every landcover type. And DM stakeholder still need this number to do some decision.

I think we must separate the ways to set the landcover grouping and to set the field that calculated for crop.

Gustry:
@samnawi I'm not sure I got everything from your message ;-)
The meaning of this `Do Not Report` is to remove some values in your exposure dataset on-the-fly that you don't want to use. We are really removing these features from the exposure in our internal exposure dataset. Is-it not OK?

safe/definitions/exposure_classifications.py
```diff
@@ -29,6 +30,21 @@
 # to generate the value_mapping. Please do not remove it.
 # See https://github.com/kartoza/osm-reporter/wiki how to use this key.
 
+do_not_report_class = {
+    'key': DO_NOT_REPORT,
+    'name': tr('Do Not Report'),
+    'description': tr(
+        'Land use types that will be excluded in the analysis.'),
```
Gustry: Ah, now you make its own variable. Why not making it generic? (not using `land use`) and to add it to roads/buildings?

Gustry: Ok, let's keep it simple for now, not to add it other exposures

ismailsunni:
After discussion in gitter (27 July 2017). We decide not to merge this one. There is a problem like @samnawi mention above.

Since the real data of non-crop exposure class won't have productivity/productivity_cost/productivity_value, it won't be shown in the report. Then the problem's solved.

See my comment in https://github.com/inasafe/inasafe/issues/3626#issuecomment-316919070

closing this.

PR #4344, opened 2017-07-28
Use productivity, production cost, and production value.

ismailsunni:
### What does it fix?
* Ticket: No tickets
* Funded by: DFAT
* Description: 
  Make sure the definition are uniform, based on this diagram:
productivitiesdiagram


Gustry:
Can you update the graffle file?

Gustry:
Thanks. I'm merging this. Doing some translations before my new flight to London ;-)

We need to add this diagram in the main window help.

Gustry:
@ismailsunni I think there is one mistake in your diagram. There shouldn't be any "rate" in the bottom line. no?

PR #4349, opened 2017-08-02
Fix not removed variable.

ismailsunni:
### What does it fix?
* Ticket: No ticket
* Funded by: DFAT
* Description: 
  - Fix not removed project variable from impact layer.


Gustry:
As you are on leave @ismailsunni, I'm gonna merge this one for you

Issue #4252, opened 2017-06-08
We need to implement an InaSAFE expression group


timlinux:
@gustry to make first pass implementation

Start by exposing everything from the substitution list.

Make sure everything has help files too please.

Gustry:
This is done I think in the PR above. @myarjunar is working on some new expressions. We need to check that we have what we need for now.

timlinux:
@myarjunar  in the templates can you please remove any substitutions in favor of expressions. Please see https://github.com/inasafe/inasafe/wiki/Map-Composer-Templating#element-tokens - I would like to see all of this Replacement Tokens removed from the above page and implemented rather as QGIS / InSAFE Expressions which can be inserted into the template.

Also for the logos and images which can have their paths set via an expression, they should be remove from the above page and replaced with expressions in the template.

We also need to get the expressions and their help into the definitions_help please @Gustry

Gustry:
We only remove tokens only from existing templates @timlinux? But we keep them in the code, because if we remove them, then it's a API break.

I added expression help in the dialog. We may need some polishing. Like translations will be missing with the current implementation.

PR #4350, opened 2017-08-03
add expression to fetch values from the analysis layer

Gustry:
### What does it fix?
* Ticket: #4252 
* Funded by: DFAT
* Description: add expression to fetch values from the analysis layer

### DIFF
--- a/safe/definitions/provenance.py
+++ b/safe/definitions/provenance.py
@@ -195,6 +195,38 @@ provenance_layer_profiling = {
     'provenance_key': layer_purpose_profiling['key']
 }
 
+# Layers ID
+provenance_layer_exposure_summary_id = {
+    'key': 'provenance_layer_exposure_summary_id',
+    'name': layer_purpose_exposure_summary['name'] + ' ID',
+    'provenance_key': layer_purpose_exposure_summary['key'] + '_id'
+}
+provenance_layer_aggregate_hazard_impacted_id = {
+    'key': 'provenance_layer_aggregate_hazard_impacted_id',
+    'name': layer_purpose_aggregate_hazard_impacted['name'] + ' ID',
+    'provenance_key': layer_purpose_aggregate_hazard_impacted['key'] + '_id'
+}
+provenance_layer_aggregation_summary_id = {
+    'key': 'provenance_layer_aggregation_summary_id',
+    'name': layer_purpose_aggregation_summary['name'] + ' ID',
+    'provenance_key': layer_purpose_aggregation_summary['key'] + '_id'
+}
+provenance_layer_analysis_impacted_id = {
+    'key': 'provenance_layer_analysis_impacted_id',
+    'name': layer_purpose_analysis_impacted['name'] + ' ID',
+    'provenance_key': layer_purpose_analysis_impacted['key'] + '_id'
+}
+provenance_layer_exposure_summary_table_id = {
+    'key': 'provenance_layer_exposure_summary_table_id',
+    'name': layer_purpose_exposure_summary_table['name'] + ' ID',
+    'provenance_key': layer_purpose_exposure_summary_table['key'] + '_id'
+}
+provenance_layer_profiling_id = {
+    'key': 'provenance_layer_profiling_id',
+    'name': layer_purpose_profiling['name'] + ' ID',
+    'provenance_key': layer_purpose_profiling['key'] + '_id'
+}
+
 provenance_list = [
     provenance_action_checklist,
     provenance_aggregation_keywords,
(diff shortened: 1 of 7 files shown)


safe/gis/expressions.py
```diff
@@ -0,0 +1,38 @@
+# coding=utf-8
+from qgis.core import (
```
ismailsunni: __[flake8]__
* [F401] 'qgis.core.QgsVectorLayer' imported but unused

safe/gis/expressions.py
```diff
@@ -0,0 +1,37 @@
...
+
+@qgsfunction(
+    args='auto', group='InaSAFE', usesGeometry=False, referencedColumns=[])
+def inasafe_impact_analysis_layer(field, feature, parent):
+    """Retrieve a value from a field in the impact analysis layer."""
+    _ = feature, parent  # NOQA
+    project_context_scope = QgsExpressionContextUtils.projectScope()
+    registry = QgsMapLayerRegistry.instance()
+
+    key = provenance_layer_analysis_impacted_id['provenance_key']
```
myarjunar: Is it possible to make the key as a parameter?

Gustry: If you need the key, we need to support both I think.
Because we want users to create the own QGIS template. The key is very hidden from our users and we want to keep it simple, by looking in the attribute table the name field. In my opinion, I would prefer the field name

myarjunar: so the purpose is this expression alaso can be used by the user? if so, yes I agree to keep it simple, but it means we need to create another function if we want to get the data from another layer, is it correct?

safe/definitions/units.py
```diff
@@ -327,6 +329,133 @@
...
+    ],
+    'citations': [
+        {
+            'text': None,
+            'link': None
+        }
+    ]
+}
+
+unit_tens= {
```
ismailsunni: __[flake8]__
* [E225] missing whitespace around operator

safe/definitions/units.py
```diff
@@ -327,6 +329,133 @@
...
+    ],
+    'citations': [
+        {
+            'text': None,
+            'link': None
+        }
+    ]
+}
+
+unit_hundreds= {
```
ismailsunni: __[flake8]__
* [E225] missing whitespace around operator

safe/utilities/rounding.py
```diff
@@ -284,3 +286,26 @@ def html_scientific_notation_rate(rate):
...
+    :param value: The value.
+    :type value: int
+
+    :return: The new value and the denomination as a unit definition.
+    :rtype: list(int, safe.unit.definition)
+    """
+    if value == nominal_mapping.keys()[0]:
+        return 1, nominal_mapping[nominal_mapping.keys()[0]]
+
+    for min_value, max_value in zip(nominal_mapping.keys(), nominal_mapping.keys()[1:]):
```
ismailsunni: __[flake8]__
* [E501] line too long (88 > 79 characters)

safe/gis/expressions.py
Gustry: Yes, definitely, we want to encourage users to create their own map template. That's why we are doing this expression system. Then can create their own QPT by using the powerful QGIS system expression.

I just made for the analysis layer.
What do you need for you? Just let me know which other layer you need.

Gustry:
I just pushed the value name, can you check the second commit @myarjunar.

@timlinux Can you check the expression help? It's the docstring in expressions.py.

screen shot 2017-08-03 at 16 56 31

safe/utilities/test/test_rounding.py
```diff
@@ -78,5 +77,44 @@ def test_fatalities_range(self):
...
+
+        result = denomination(10000)
+        self.assertEqual(10, result[0])
+        self.assertEqual('unit_thousand', result[1]['key'])
+
+        result = denomination(101000)
+        self.assertEqual(101, result[0])
+        self.assertEqual('unit_thousand', result[1]['key'])
+
+        result = denomination(100000000000000)
```
ismailsunni: Can you put test for non positive integer? Like float, or negative number? Just in case we have a bug, then the number is not positive integer, then we are still sure that it won't throw python exception.

safe/definitions/provenance.py
```diff
@@ -195,6 +195,38 @@
     'provenance_key': layer_purpose_profiling['key']
 }
 
+# Layers ID
+provenance_layer_exposure_summary_id = {
+    'key': 'provenance_layer_exposure_summary_id',
+    'name': layer_purpose_exposure_summary['name'] + ' ID',
```
ismailsunni: Should we create a method for this? Just to make sure we have the same behaviour.

myarjunar:
LGTM :+1:

safe/utilities/test/test_rounding.py
Gustry: float was already done. I just added null and negativ values to the test.

myarjunar:
Hi @Gustry , I'm just thinking, can we use expression for all elements in the inforgraphic that we can set using expression like section header or something?

Gustry:
Yes, of course. WHy not? Can you make a quick draft of your expressions you need? Thanks

Issue #4352, opened 2017-08-07
Impact result from land cover vs flood raster with aggregation not complete polygon


felix-yew:
### Problem
When I try the analysis of land cover vs flood raster with aggregation, I get the impact result map is not in complete polygon. It is different when I run the analysis of land cover vs flood raster without aggregation, that show the complete polygon. 
![image](https://user-images.githubusercontent.com/13621886/29012552-a6463dd8-7b66-11e7-975c-87f99457ec36.png)
When I check the total number in each estimated area also have the same total number. It is strange, because in the general report, all of them show the `use caution` result in the same total number, whether supposedly if there is a missing shape area, then the `use caution' area at least have a smaller total number than the full of impact polygon result.

And when I run with debug mode, I get some error message 
```
Traceback (most recent call last):
  File "C:/Users/yanuaf/.qgis2/python/plugins\inasafe\safe\gui\widgets\dock.py", line 1215, in accept
    self.impact_function, self.iface)
  File "C:/Users/yanuaf/.qgis2/python/plugins\inasafe\safe\gui\analysis_utilities.py", line 64, in generate_impact_report
    return impact_report.process_components()
  File "C:/Users/yanuaf/.qgis2/python/plugins\inasafe\safe\report\impact_report.py", line 715, in process_components
    context = _extractor_method(self, component)
  File "C:/Users/yanuaf/.qgis2/python/plugins\inasafe\safe\report\extractors\analysis_detail.py", line 544, in analysis_detail_extractor
    total_count = int(float(feat[field_index]))
KeyError: '-1'
```

### Expected Behaviour
The analysis result IF of land cover vs flood raster with aggregation can show the best result, without missing area.

### Environment
* InaSAFE : 4.2.0b0 (Develop)
* QGIS : 2.14.14
* OS : Win-7 Pro/ 64-bit 
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : Jakarta_Landcover_WGS84.shp
  * Hazard : Jakarta_Flood_HKV_WGS84.tif
  * Aggregation : Jakarta_District_Boundary_WGS84.shp
  * Extent : 
  
Checklist (if applicable): 
* [x] Did you try to run it in debug mode ? Yes, no result and busy cursor
* [x] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? In Real World Data

### Proposed Solution
Make it happy

Gustry:
Thanks for reporting the issue.
[...]

PR #4357, opened 2017-08-10
add more debug info when an exception occurs in the analysis extractor

Gustry:
### What does it fix?
* Partial Fix Ticket: #4352
* Funded by: DFAT
* Description: Add one debug line when an exception occurs

I will try to fix it @myarjunar but I wanted to focus on the first part of the ticket (which seems related to IF itself), so I would prefer to replicate the other issue.

Can we merge this one @myarjunar and if you have time to look at this issue? Thanks

It's failing on the `Field name affected_production_cost, field index -1`

### DIFF
--- a/safe/report/extractors/analysis_detail.py
+++ b/safe/report/extractors/analysis_detail.py
@@ -1,5 +1,9 @@
 # coding=utf-8
+
 """Module used to generate context for analysis detail section."""
+
+import logging
+
 from safe.definitions.exposure import exposure_all, exposure_population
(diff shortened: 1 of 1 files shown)


myarjunar:
LGTM :+1: I will look at the issue

myarjunar:
@Gustry I think I will make a patch for this. Instead of raising error, I think we only need to continue to next field if current field is missing. What do you think?

Gustry:
Sure, do what you need. I just wanted more informations in my case this morning

PR #4358, opened 2017-08-10
fix field not found error

myarjunar:
### What does it fix?
* Ticket: #4352 (partfix)
* Funded by: DFAT
* Description:


Gustry:
Ok, I was not sure that it's was normal that the code is trying to fetch a nonexisting field.
LGTM so

myarjunar:
on this case https://github.com/inasafe/inasafe/issues/4352 , I think all of landcover's extra fields should be availbale in the exposure summary table, but it turns out the extra fields is not exist. It's @ismailsunni 's side I guess :sweat_smile:

Gustry:
So it was not really fixing the issue? But only skipping the issue, right?

PR #4359, opened 2017-08-10
make PEP 257 compulsory

Gustry:
### What does it fix?
* Funded by: DFAT
* Description: 

@inasafe/developers Since a long time, we are trying to have a good documentation in our code. It's time to add a check about our docstrings. We were trying to follow landscape.io these last few months.

Now, I'm adding PEP257 check on our docstrings. For now, a lot of tests are disabled according to the PEP.
I would like to enable more in the next few months. Like I would like to have D201, D202, D203 ...
These are very easy to fix every time you open a file.
http://pep257.readthedocs.io/en/latest/error_codes.html
* D400 First line should end with a period 😈
* D203 1 blank line required before class docstring
* D204 1 blank line required after class docstring etc ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ before_install:
 install:
   - pip install --upgrade pip
   - pip install --upgrade pep8
+  - pip install --upgrade pep257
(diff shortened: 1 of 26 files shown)


Gustry:
I just disabled the test on travis. But you should run make pep257 on your local machine before the commit sometimes.

PR #4362, opened 2017-08-10
Infographic patch

myarjunar:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: fix no data placement value

### DIFF
--- a/safe/gis/expressions.py
+++ b/safe/gis/expressions.py
@@ -39,7 +39,7 @@ def inasafe_impact_analysis_layer(field, feature, parent):
 
     index = layer.fieldNameIndex(field)
     if index < 0:
-        return index
+        return None
 
     feature = layer.getFeatures().next()
     return feature[index]
@@ -95,7 +95,7 @@ def inasafe_place_value_coefficient(number, feature, parent):
         value, unit = denomination(rounded_number, 1000)
         return str(round(value, 1))
     else:
-        return no_data_replacement['string_format']
+        return None
 
 
 @qgsfunction(
--- a/safe/gis/expressions.py
+++ b/safe/gis/expressions.py
@@ -39,7 +39,7 @@ def inasafe_impact_analysis_layer(field, feature, parent):
 
     index = layer.fieldNameIndex(field)
     if index < 0:
-        return None
+        return index
 
     feature = layer.getFeatures().next()
     return feature[index]
@@ -85,15 +85,17 @@ def inasafe_place_value_coefficient(number, feature, parent):
     It needs to be used with inasafe_number_denomination_unit.
     """
     _ = feature, parent  # NOQA
-    if number is None:
+
+    if number >= 0:
+        rounded_number = round_affected_number(
+            number,
+            enable_rounding=True,
+            use_population_rounding=True
+        )
+        value, unit = denomination(rounded_number, 1000)
+        return str(round(value, 1))
+    else:
         return no_data_replacement['string_format']
-    rounded_number = round_affected_number(
-        number,
-        enable_rounding=True,
-        use_population_rounding=True
-    )
-    value, unit = denomination(rounded_number, 1000)
-    return str(round(value, 1))
 
 
 @qgsfunction(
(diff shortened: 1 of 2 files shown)


safe/gis/expressions.py
```diff
@@ -39,7 +39,7 @@ def inasafe_impact_analysis_layer(field, feature, parent):
 
     index = layer.fieldNameIndex(field)
     if index < 0:
-        return None
+        return index
```
Gustry: Why are you returning the index?
It will be wrong value to the user to return a negative index instead of NULL

safe/gis/expressions.py
```diff
@@ -85,15 +85,17 @@ def inasafe_place_value_coefficient(number, feature, parent):
...
+    if number >= 0:
+        rounded_number = round_affected_number(
+            number,
+            enable_rounding=True,
+            use_population_rounding=True
+        )
+        value, unit = denomination(rounded_number, 1000)
+        return str(round(value, 1))
+    else:
         return no_data_replacement['string_format']
```
Gustry: BTW, what is no_data_replacement string ?

safe/gis/expressions.py
myarjunar: because in our template this function is a parameter for `inasafe_place_value_coefficient()` . If the field is not found in the layer, we want to show `no_data_replacement` string which is `dash` instead of nothing. And somehow if `inasafe_place_value_coefficient()` is being called with None. somehow it doesn't give the result we are expecting

safe/gis/expressions.py
myarjunar: it is a dash

safe/gis/expressions.py
Gustry: This change for me is a blocker. Can we fix the function `inasafe_place_value_coefficient` to take if the value is None instead?

We are going to teach users to create their own template. Some of them will use this function without `inasafe_place_value_coefficient` and returning -1 looks like a bug from the user point of view. They need to write a QGIS expression like: ` if( inasafe_impact_analysis_layer('displaced') = -1, None, inasafe_impact_analysis_layer('displaced'))`

myarjunar: The code won't works if we give Null or None as a parameter and will always returns None or Null. Seems like they don't recognize Null. Could you give it a try?

Gustry: Ok, I understood why you are updating this expression ;-)
Yes, I'm trying on my side.

So can we do it on the user side? 
`if(inasafe_impact_analysis_layer('displaced') is None, inasafe_place_value_coefficient(inasafe_impact_analysis_layer('displaced')), if wrong do something here)` ?

safe/gis/expressions.py
Gustry: Now I'm thinking that we shouldn't provide extra behaviors like this and only do what the function is supposed to do : return the coefficient or NULL. Especially not mixing integers/float AND chars like dash.
If the user wants a dash if the string is empty, then it is a `IF` statement in the QGIS expression. We might have users who doesn't want a dash. It something that can easily add in a QGIS expression. I would like to stay as more generic as possible in these expressions.

safe/gis/expressions.py
Gustry: Yes you are right, I tried:
```
def inasafe_place_value_coefficient(number, feature, parent):
    """Given a number, it will return the coefficient of the place value name.

    For instance:
    *  inasafe_place_value_coefficient(10) -> 1
    *  inasafe_place_value_coefficient(1700) -> 1.7

    It needs to be used with inasafe_number_denomination_unit.
    """
    _ = feature, parent  # NOQA
    print type(number)
    print number
```

`inasafe_place_value_coefficient(None)` is not evaluated by QGIS, it doesn't print anything. But it works well with anything else. So can we tweak the expression in the expression builder?

safe/gis/expressions.py
timlinux: > Now I'm thinking that we shouldn't provide extra behaviors like this and only do what the function is supposed to do : return the coefficient or NULL. Especially not mixing integers/float AND chars like dash.
> If the user wants a dash if the string is empty, then it is a IF statement in the QGIS expression. > We might have users who doesn't want a dash. It something that can easily add in a QGIS expression. I would like to stay as more generic as possible in these expressions.

+1 lets keep our API's clean and logical

myarjunar: done

Issue #4240, opened 2017-05-29
Could you bring back the old way to generate and style the map?


adissadis:
### Problem

Currently in InaSAFE 4.0 and 4.1, all of the reports that user get from InaSAFE analysis process are stored in temp folder. Based on users experience & feedback from several training that we had in April, they confused to find where is temp folder for all InaSAFE analysis process. Although this can be solved by changing your Location for InaSAFE analysis results in InaSAFE options (which we found this solution later).

Another thing is the features where the user can style their own map before click Print button in InaSAFE dock. There was a user who already had QGIS and InaSAFE training a long time ago and joined again in our training in April. He asked to us about the old way to style the map before click Print button and to freely choose the InaSAFE theme and orientation for their maps. This problem also can be solve by open the tmp folder to find .qpt files and try to style it in the QGIS Map Composer. But IMO, if the user is new to GIS and they not learn how to use map composer yet or they just learn how to style a map, it will be take time for them to produce a map based on what they want to show in PDF map.

Sorry to put this issue late but if this possible to you guys, could you bring back again the old way to generate and style the reports? Thanks!

### Expected Behaviour
The user can style their layers first and put what they want to show before click Print button in InaSAFE dock.

When the user click Print button in InaSAFE dock, they can choose the paper's orientation and select their own folder to put all the reports in.

### Environment
* InaSAFE : 4.0 & 4.1
* QGIS : 2.14.15
* OS :  ubuntu, windows 10
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : 
  * Hazard : 
  * Aggregation :
  * Extent :

Gustry:
Hi @adissadis,

This is still in our roadmap to bring back this feature. We are thinking about some nice things about reporting. I hope we will work on that for 4.2.

adissadis:
Hi @Gustry,
Thanks for your information! Looking forward to get this features back to InaSAFE. :smile:

Charlotte-Morgan:
hi @adissadis - agree, this is an important thing for us to address. However the work around is to share the *.qpt for users in the short term.

timlinux:
I'm thinking maybe a good workaround for this would be allow the user to mask the standard templates by placing their customized versions (with the same names as the package delivered ones) in their .qgis2/inasafe/ directory.

This way when InaSAFE scans for the template it will first look in the user directory, then in the InaSAFE plugin directory.

ismailsunni:
Hi @timlinux 
[...]

PR #4365, opened 2017-08-15
Another patch for Infographic (icons, header, functionality) 

myarjunar:
### What does it fix?
* Ticket: #4240 
* Funded by: DFAT
* Description: icons and template update, make infographic appears when user click print button from the dock

selection_129

cc @felix-yew @ivanbusthomi


safe/definitions/utilities.py
```diff
@@ -1,6 +1,6 @@
 # coding=utf-8
 """Utilities module for helping definitions retrieval."""
-
+import os
 from os.path import join, exists
```
Gustry: It's only a small comment. We were already importing `from os.path import join,exists` you could just add `splittext` on this line as it's the only function you are using. We also try having one blank line after module docstring

myarjunar: fixed

PR #4367, opened 2017-08-16
fix how we fetch items in composer, problem with bindings

Gustry:
### What does it fix?
* Ticket: #4271 
* Funded by: DFAT
* Description: @wonder-sk gave me a snippet by email. It seems to work. @myarjunar can you try to see if it works.

I still don't have any map, but at least I have the north arrow!
screen shot 2017-08-16 at 11 50 41

@myarjunar we need to remove the deprecated call, also using this help it works for you:
```
        composer_item = composition.getComposerItemById(item_id)
        try:
            html_element = composition.getComposerHtmlByItem(composer_item)
        except:
            pass
```

CC: @mbernasocchi

### DIFF
--- a/safe/report/processors/default.py
+++ b/safe/report/processors/default.py
@@ -50,6 +50,32 @@ __revision__ = '$Format:%H$'
 LOGGER = logging.getLogger('InaSAFE')
 
 
+def composition_item(composer, item_id, item_class):
+    """Fetch a specific item according to its type in a composer.
+
+    We got some problem with QgsComposition::getComposerItemById. Don't use it,
+    and use this function instead.
+    See https://github.com/inasafe/inasafe/issues/4271
+
+    :param composer: The composer to look in.
+    :type composer: QgsComposition
+
+    :param item_id: The ID of the item to look for.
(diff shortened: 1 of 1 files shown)


safe/report/processors/default.py
```diff
@@ -50,6 +50,32 @@
...
+
+    :param item_id: The ID of the item to look for.
+    :type item_id: basestring
+
+    :param item_class: The expected class name.
+    :type item_class: cls
+
+    :return: The composition item, inherited class of QgsComposerItem.
+    """
+    for item in composer.items():
```
myarjunar: before this line maybe we can check if the item class is QgsComposerMap and use `getComposerMapById` method instead to get the specific item that we want

Gustry: Good idea, but I'm facing a problem about the ID which is a string. It should be an int. We talked on gitter: https://gitter.im/inasafe/inasafe?at=59942e6e81cd9a5d7e9c9578

myarjunar:
LGTM :+1: thanks!

PR #4369, opened 2017-08-16
add special case for me to have the map working

Gustry:
### What does it fix?
* Ticket: #4271 
* Funded by: DFAT
* Description: 

@myarjunar Can you check it's still working for you?

For me, I have my two maps now, but the overview has still a problem with colors:
screen shot 2017-08-16 at 15 28 18


myarjunar:
LGTM if the purple map is not a problem :wink:

Issue #3786, opened 2017-02-07
Wizard : Improvements when we select the name field in the aggregation layer


Gustry:
### Problem

In this step, we need to select a field which should be unique.

![screen shot 2017-02-07 at 10 34 47](https://cloud.githubusercontent.com/assets/1609292/22685600/38abb312-ed21-11e6-93ac-57d10e8d0e53.png)

### Expected Behaviour

It could be helpful we can know the number of features in the layer and also the number of unique values. We will quickly notice of this field is a kind of ID.
The same would be very nice when we have to select exposure ID, aggregation ID, hazard ID.

Gustry:
Thanks Ismail for this first step.

I'm just thinking about exposure ID, hazard ID, aggr ID, can't we add the field name plus a number with brackets (or percent) and italic in the combobox?

ismailsunni:
Yes, I am thinking the same solution also for that. I think it's possible to do it. I will try to implement it.

ismailsunni:
Hi @Gustry 
Can you elaborate or give example regarding your comment:
> I'm just thinking about exposure ID, hazard ID, aggr ID, can't we add the field name plus a number with brackets (or percent) and italic in the combobox?

Is it something like : _aggregation_id (18)_
What about the `%` ?

I am sorry, I understood you comment on Aug, but not now :)

Gustry:
In the combobox, to display the field name plus the percentage of uniqueness of this field.
If we display only the number, we still don't know if it's very unique or not, so not the best.

This is my combobox and fields available in this combobox for my district aggregation layer:
```
QComboBox: Aggregation field ID
-> country_name(1%)
-> province_name(20%)
-> district_name(100%)
-> district_id(100%)
```

We would need a sentence in the help string about this percent.

Does it make sense?

ismailsunni:
Hmm, quite strange for me to use percentage to represent the uniqueness. Can we just use `number_of_unique_value / number_of_feature` ? e.g. `country_name (100/108)`. Or is it important to know the degree of uniqueness? I thought it's either unique or not.

Gustry:
Yes, go for it! It's better ;-)

ismailsunni:
Ouch, it turns out not so straight forward since we use the name of the combo box name directly. In order to change this, I need to modify SelectParameter and SelectParameterWidget to add `shown option` or something like that and store the value as data. I will move it back to backlog.

ismailsunni:
[...]

PR #4379, opened 2017-08-21
Show uniqueness in step KW field #3786.

ismailsunni:
### What does it fix?
* Ticket: #3786 
* Funded by: DFAT
* Description: 
   Show uniqueness in step KW field #3786.
screen shot 2017-08-21 at 17 54 01

Adding in the InaSAFE fields wizard is more complex.

### DIFF
--- a/safe/gui/tools/wizard/step_kw30_field.py
+++ b/safe/gui/tools/wizard/step_kw30_field.py
@@ -30,6 +30,7 @@ __license__ = "GPL version 3"
 __email__ = "info@inasafe.org"
 __revision__ = '$Format:%H$'
 
+LOGGER = logging.getLogger('InaSAFE')
 FORM_CLASS = get_wizard_step_ui_class(__file__)
 
 # Mode
@@ -124,6 +125,9 @@ class StepKwField(WizardStep, FORM_CLASS):
         if not isinstance(field_names, list):
             field_names = [field_names]
         field_descriptions = ''
+        LOGGER.debug('Test')
+        feature_count = self.parent.layer.featureCount()
+        LOGGER.debug('Feature count: %s' % feature_count)
         for field_name in field_names:
             layer_fields = self.parent.layer.dataProvider().fields()
             field_index = layer_fields.indexFromName(field_name)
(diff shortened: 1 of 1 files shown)


ismailsunni:
@Gustry if it's LGTM, go on to merge it. And update transifex for the new strings. It's just 3 simple strings.

safe/gui/tools/wizard/step_kw30_field.py
```diff
@@ -124,6 +125,9 @@ def on_lstFields_itemSelectionChanged(self):
         if not isinstance(field_names, list):
             field_names = [field_names]
         field_descriptions = ''
+        LOGGER.debug('Test')
+        feature_count = self.parent.layer.featureCount()
+        LOGGER.debug('Feature count: %s' % feature_count)
```
Gustry: Can we remove these debug noise?

safe/gui/tools/wizard/step_kw30_field.py
```diff
@@ -134,16 +138,23 @@ def on_lstFields_itemSelectionChanged(self):
...
                 i is not None and unicode(i) or 'NULL'
-                for i in unique_values]
+                for i in unique_values[0:48]]
             unique_values_str = ', '.join(unique_values_str)
             field_descriptions += tr('<b>Field name</b>: {field_name}').format(
                 field_name=field_name)
             field_descriptions += tr(
                 '<br><b>Field type</b>: {field_type}').format(
                 field_type=field_type)
+            if len(unique_values) == feature_count:
```
Gustry: Some providers won't be able to provide the feature count. For instance, the feature count on an OSM file will always return -1, because we can't count features in an OSM file efficiently. So can we skip this block if it's -1? It would not make sense to display `from -1 features`.

ismailsunni:
> Your screenshot is a hazard layer, it doesn't make any sense here, but more an aggregation layer as we need to select a unique column. So I guess, it's the same step but anyway. Thanks. There is something to fix.

Yes, it just example. I will make it to show only on aggregation layer.

safe/gui/tools/wizard/step_kw30_field.py
ismailsunni: Ouch, I never know about it. I will skip if the count == -1.

Thanks

ismailsunni:
@gustry done
screen shot 2017-08-22 at 08 22 56

safe/gui/tools/wizard/step_kw30_field.py
```diff
@@ -134,16 +135,25 @@ def on_lstFields_itemSelectionChanged(self):
...
                 '<br><b>Field type</b>: {field_type}').format(
                 field_type=field_type)
+            if (feature_count != -1 and (
+                        layer_purpose == layer_purpose_aggregation)):
+                if len(unique_values) == feature_count:
+                    unique = tr('Yes')
+                else:
+                    unique = tr('No')
+                field_descriptions += tr(
+                    '<br><b>Unique</b>: %s (%d unique values from %d '
```
Gustry: Can we switch to the new format: https://pyformat.info/#simple ?

ismailsunni: Shame, of course. I always support to use new format, but it seems I forget about it.
Thanks

Gustry:
Thanks, it's good!

Can you squash them?

ismailsunni:
@Gustry Yes, I am planning to squash them... Thanks

PR #4386, opened 2017-08-22
adjust legend column size to the template

myarjunar:
### What does it fix?
* Ticket: #4047 
* Funded by: DFAT
* Description: Adjust column size so the legend will fit the element

selection_133
selection_134

cc: @lrpower @felix-yew


ismailsunni:
The code is LGTM.

PR #4387, opened 2017-08-22
add help from docstrings for expressions

Gustry:
### What does it fix?
* Ticket: #4252 
* Funded by: DFAT
* Description: This PR brings expressions docstrings into the InaSAFE help dialog.

One limitation, on this, we can't have translations. I'm looking for a patch in QGIS 3 where we can provide our custom translated strings.

screen shot 2017-08-22 at 18 28 53

@timlinux I think we should add a 15.2 for `Element-ID` from https://github.com/inasafe/inasafe/wiki/Map-Composer-Templating, there is no help for these, except on line in this wiki. `Element Tokens` will be removed according to #4252 @myarjunar?

### DIFF
--- a/safe/gui/tools/help/definitions_help.py
+++ b/safe/gui/tools/help/definitions_help.py
@@ -3,8 +3,10 @@
 
 import copy
 import logging
+import re
 from PyQt4 import QtCore
 from os.path import exists
+from inspect import getmembers
 
 import safe.definitions as definitions
 import safe.definitions.post_processors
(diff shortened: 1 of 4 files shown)


safe/gui/tools/help/definitions_help.py
```diff
@@ -742,6 +746,52 @@ def content():
...
+
+    _create_section_header(
+        message,
+        table_of_contents,
+        'reporting',
+        tr('QGIS Expressions'),
+        heading_level=2)
+
+    qgis_expressions = {
+        fct[0]: fct[1] for fct in getmembers(expressions)
```
ismailsunni: Perhaps you can put some comment here to explain what is fct and why it can be filtered by `QgsExpressionFunction`

PR #4400, opened 2017-08-28
Update field descriptions and concept to be more informative and check the calculation

ismailsunni:
### What does it fix?
* Ticket: No ticket
* Funded by: DFAT
* Description: 
   - Update the field description to be like this:
screen shot 2017-08-28 at 21 02 27
screen shot 2017-08-28 at 21 04 03

There is a unit for productivity rate and I remove the e.g. in the below.

I also checked the calculation, all is good:
screen shot 2017-08-28 at 21 26 38

See the red boxes.
In the top table, we calculate `productivity` per polygon simply calculating `productivity rate` * `size` (are) / 10.000. We divide by 10.000 to convert the meter square to hectare.

In the middle table, we sum up the `productivity` that has the same `exposure_class`.

You can try to check the other `exposure_class` (see the value map in the dock and the third table to get the reference, if needed). 

So, my conclusion, the number is right. The size field is in meter square, but we calculate in hectare. That's why the final number is smaller since we divide by 10.000.

### Checklist:


timlinux:
Thanks @ismailsunni  - did you update the report too to show an explanation of quintals? LGTM to merge when you are ready....

ismailsunni:
Hi @timlinux 

> Thanks @ismailsunni - did you update the report too to show an explanation of quintals? LGTM to merge when you are ready....

The plan that I mentioned yesterday, I will change it to hundred kilograms, but I will translate it to quintal in bahasa Indonesia.

PR #4442, opened 2017-09-20
Fix bug step fc function

ismailsunni:
### What does it fix?
* Ticket: No Ticket
* Funded by: WB
* Description: 
Fixing this bug:
```
AttributeError: 'WizardDialog' object has no attribute 'step_fc_function'
  File "C:/Users/aless/.qgis2/python/plugins\inasafe\safe\gui\tools\wizard\wizard_dialog.py", line 678, in on_pbnNext_released
    new_step.set_widgets()
  File "C:/Users/aless/.qgis2/python/plugins\inasafe\safe\gui\tools\wizard\step_kw65_summary.py", line 130, in set_widgets
    im_func = self.parent.step_fc_function.selected_function()
AttributeError: 'WizardDialog' object has no attribute 'step_fc_function'
```

To replicate, assign keyword in a layer when working on IFCW.

The fix is just merely removing old code that's not longer relevant. Previously we need to set the IF that we want to use, but now, all IF is possible (no need to choose IF anymore).

### Checklist:

### DIFF
--- a/metadata.txt
+++ b/metadata.txt
@@ -18,7 +18,7 @@ about=Developed for the Indonesian Government - BNPB, Australian Government - AI
 # If you change version and status here, please also change in safe/definitions/versions.py
 version=4.3.0
 # alpha, beta, rc or final
-status=final
+status=alpha
 
 
 # end of mandatory metadata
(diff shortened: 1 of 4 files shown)


ismailsunni:
Thanks @Gustry 
I agree with your suggestion, I create a ticket here : https://github.com/inasafe/inasafe/issues/4446

I will merge this one first, since we need to at least make the IFCW works again.

Issue #4428, opened 2017-09-12
Tsunami and Structure IF Report Bugs


lrpower:
### Problem
A few issues I have noticed in the Tsunami Raster on Structure Point IF Report

1. High, Medium, Low and Dry does not equal the Total. While I don't want the Dry value to be rounded, doesn't InaSAFE logic mean it should be?
![image](https://user-images.githubusercontent.com/10228025/30310603-95ae9044-97d4-11e7-9382-2f5646300d1f.png)

2. There should be more definition between these two results because the affected, not affected, not exposed counts and totals have nothing to do with the hazard zones. 
![image](https://user-images.githubusercontent.com/10228025/30311408-6d0ff354-97d8-11e7-86b5-63b207c9f64d.png)

3. This table is a mishmash of rounding and not rounding. Residential and Other are rounded, but the others appear not to be.  
![image](https://user-images.githubusercontent.com/10228025/30310901-2c07ab10-97d6-11e7-8d0e-e2f1206cc0fc.png)

4. This should be titled _Estimated number of buildings **affected** by aggregated area_.
Since an aggregation has not been used, this probably doesn't need to be displayed because it is just a simplified version of the by Structure Type table.
![image](https://user-images.githubusercontent.com/10228025/30310761-6b68d7e4-97d5-11e7-9407-ab18db20e9ba.png)

### Environment
* InaSAFE : 4.3.0
* QGIS : 2.18.10
* OS : Windows-7-6.1.7601-SP1
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : Structure points
  * Hazard : Tsunami Raster
  * Aggregation :n/a
  * Extent : Nukalofa
  
### Proposed Solution
1. Either add Not Exposed to equal the total or calculate the total of the figures shown
2. Add a line to separate the two tables.
3. Be consistent with rounding values
4. Remove aggregation table when not applying an aggregation; add **affected** if the values are going to be the affected values. 
5. Make a decision regarding the term Structures or Buildings. They are both used throughout the report.

Gustry:
Thanks a lot @lrpower for this ticket.
I will try to run analysis on my side to check your issues.

1. In the Analysis layer summary in your legend, is there a field  `Not exposed` ? I guess you are right, the not exposed class is not shown.


2. > There should be more definition between these two results because the affected, not affected, not exposed counts and totals have nothing to do with the hazard zones.

These definitions are based on the hazard class, so not really true.
[...]

Issue #4412, opened 2017-08-31
InaSAFE reporting bugs


Charlotte-Morgan:
A few bugs in the reports - Jakarta flood analysis. 

### problems ###
- [x] the report mentions aggregation area when no aggregation was used
![image](https://user-images.githubusercontent.com/5743135/29904163-dcbc6b52-8e4a-11e7-8350-0c0d8743c946.png)
- [x] map reference is not computing
![image](https://user-images.githubusercontent.com/5743135/29904198-ff69dd9c-8e4a-11e7-81b3-3c0042305a75.png)
-  [ ] action check list has 2 blank pages
- [x] infographic report - unit should be thousand not thousands; million not millions etc
![image](https://user-images.githubusercontent.com/5743135/29905911-ac6806dc-8e54-11e7-808d-24168b925e89.png)

### proposed solution ##
- [x] update title on tables - drop the "per aggregation area" if aggregation data set is not used
- [x] make expression for map reference happy
- [ ] remove blank pages
- [x] tweak the unit word

### Data ###
Petabencana flood and vector population with no aggregation.

Gustry:
I'm only gonna fix the CRS element @myarjunar, it's in my next PR

myarjunar:
closed in #4448

PR #4448, opened 2017-09-20
aggregation specific report

myarjunar:
### What does it fix?
* Ticket: #4428 #4412 
* Funded by: DFAT
* Description: 

### Checklist:

### DIFF
--- a/safe/report/extractors/aggregate_result.py
+++ b/safe/report/extractors/aggregate_result.py
@@ -57,6 +57,10 @@ def aggregation_result_extractor(impact_report, component_metadata):
     aggregation_summary_fields = aggregation_summary.keywords[
         'inasafe_fields']
     debug_mode = impact_report.impact_function.debug_mode
+    use_aggregation = bool(impact_report.impact_function.provenance[
+                               'aggregation_layer'])
+    if not use_aggregation:
+        return context
 
     """Filtering report sections"""
 
--- a/safe/report/extractors/general_report.py
+++ b/safe/report/extractors/general_report.py
@@ -4,7 +4,8 @@ from safe.definitions.exposure import exposure_population
 from safe.definitions.fields import (
     hazard_count_field,
     total_field,
-    fatalities_field)
+    fatalities_field,
+    total_not_exposed_field)
 from safe.report.extractors.util import (
     layer_definition_type,
     resolve_from_dictionary,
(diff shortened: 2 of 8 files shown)


safe/report/extractors/aggregate_result.py
```diff
@@ -57,6 +57,15 @@ def aggregation_result_extractor(impact_report, component_metadata):
     aggregation_summary_fields = aggregation_summary.keywords[
         'inasafe_fields']
     debug_mode = impact_report.impact_function.debug_mode
+    use_aggregation = bool(impact_report.impact_function.provenance[
```
ismailsunni: __[flake8]__
* [F841] local variable 'use_aggregation' is assigned to but never used

safe/report/extractors/aggregate_result.py
```diff
@@ -57,10 +57,15 @@ def aggregation_result_extractor(impact_report, component_metadata):
...
-    use_aggregation = bool(impact_report.impact_function.provenance[
-                               'aggregation_layer'])
-    if not use_aggregation:
-        return context
+    # use_aggregation = bool(impact_report.impact_function.provenance[
+    #                            'aggregation_layer'])
+
+    # Before removing the aggregation table if there is no aggregation used,
+    # make sure that no aggregation is treated differently with selecting
+    # only one area.
```
myarjunar: @Gustry I need to ask you here, does `entire area` treated equally with ie choosing only `one selection area`?

safe/report/extractors/general_report.py
```diff
@@ -108,11 +109,25 @@ def general_report_extractor(impact_report, component_metadata):
...
+
+        # calculate the total exposed number by subtract total with
+        # total not exposed number
+        value_to_count = {}
+        for field in fields:
+            field_name = analysis_inasafe_fields[field['key']]
             total = value_from_field_name(field_name, analysis_layer)
+            value_to_count[field['key']] = total
+
+        total_exposed = abs(
```
ismailsunni: Why do you need to use `abs` here? Should it be always positive if total > total not exposed?. I suggest to remove the `abs`, so that if we got minus number, we can see directly if there is an error.

safe/report/extractors/aggregate_result.py
Gustry: Yes, if you have one row only, the total will be the same. Let's check together.

Gustry:
Thanks, it's better with the new field.
Do you need something about the one feature in aggregation layer?

safe/report/extractors/general_report.py
myarjunar: thanks, but we remove the calculation process in the reporting side now, we are using new `total_exposed` table from the analysis layer

safe/report/extractors/aggregate_result.py
myarjunar: ah yes about this, I mean if we are selecting only one feature from the aggregation area, does it count as an aggregation area or does it count as no aggregation area? if it's stll count as an aggregation area than we are safe to remove the aggregation report

Gustry: Even with one feature in the aggregation layer, we have an aggregation layer, so yes, it counts as an aggregation area.

PR #4456, opened 2017-09-22
fix #4435

### LINKED ISSUE #4435: Incorrect Definition for "Caution" in Flood Classes
### Problem
In the InaSAFE Realtime for Situational User, we got an explanation from @emirhartato about the definition of hazard classification for flood from PetaBencana.

Currently, we define the "caution" class is _no water encountered above ground height but there are may be floods in surrounding areas and you should use caution when moving around in this area_.
image

Apparently, the right definition of "Caution" is flooded area, but no information about the depth of water.

### Expected Behaviour
It should same with the definition in the original source
  
### Proposed Solution
I think we have to change our definition based on the reference that we use (petabencana)

cc : 
@Charlotte-Morgan @timlinux @felix-yew @ivanbusthomi

Charlotte-Morgan:
### What does it fix?
* Ticket: #4435
* Funded by: 
* Description:  update definition for Petabencana flood class "use caution"
### Checklist:

### DIFF
--- a/safe/definitions/hazard_classifications.py
+++ b/safe/definitions/hazard_classifications.py
@@ -638,9 +638,7 @@ flood_petabencana_hazard_classes = {
             'name': tr('Use caution'),
             'affected': False,
             'description': tr(
-                'No water encountered above ground height but there are may '
-                'be floods in surrounding areas and you should use caution '
-                'when moving around in this area.'),
+                'Water is present but depth is not confirmed.'),
             'fatality_rate': None,
             # displacement rate estimated from DMI analysis of historical
             # flood data and IDP numbers
--- a/safe/definitions/hazard_classifications.py
+++ b/safe/definitions/hazard_classifications.py
@@ -579,7 +579,7 @@ flood_petabencana_hazard_classes = {
             'displacement_rate': 0.05,
             'numeric_default_min': 1.5,
             'numeric_default_max': big_number,
-            'string_defaults': ['high, severe'],
+            'string_defaults': ['high', 'severe'],
             'citations': [
                 {
                     'text': None,
@@ -601,7 +601,7 @@ flood_petabencana_hazard_classes = {
             'displacement_rate': 0.03,
             'numeric_default_min': 0.7,
             'numeric_default_max': 1.5,
-            'string_defaults': ['medium, moderate'],
+            'string_defaults': ['medium', 'moderate'],
             'citations': [
                 {
                     'text': None,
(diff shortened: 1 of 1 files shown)


safe/definitions/hazard_classifications.py
```diff
@@ -638,16 +638,14 @@
...
-                'when moving around in this area.'),
+                'An unknown level of flooding - use caution - '),
             'fatality_rate': None,
             # displacement rate estimated from DMI analysis of historical
             # flood data and IDP numbers
             'displacement_rate': 0.0,
             'numeric_default_min': 0,
             'numeric_default_max': 0.1,
-            'string_defaults': ['caution'],
+            'string_defaults': ['caution, unknown'],
```
ismailsunni: If you want to register two strings, you should put `'` in between the string, before adding comma. So, it should be `['caution', 'unknown']`

safe/definitions/hazard_classifications.py
```diff
@@ -611,19 +611,19 @@
...
-                'Water encountered up to knee deep.'),
+                'Flooding of between 10 and 70 centimetres.'),
             'fatality_rate': None,
             # displacement rate estimated from DMI analysis of historical
             # flood data and IDP numbers
             'displacement_rate': 0.01,
             'numeric_default_min': 0.1,
             'numeric_default_max': 0.7,
-            'string_defaults': ['low'],
+            'string_defaults': ['low, minor'],
```
ismailsunni: Same as comment below.

safe/definitions/hazard_classifications.py
```diff
@@ -589,19 +589,19 @@
...
-                'Water is above knee height, below waist height.'),
+                'Flooding between 71 and 150 centimetres.'),
             'fatality_rate': None,
             # displacement rate estimated from DMI analysis of historical
             # flood data and IDP numbers
             'displacement_rate': 0.03,
             'numeric_default_min': 0.7,
             'numeric_default_max': 1.5,
-            'string_defaults': ['medium'],
+            'string_defaults': ['medium, moderate'],
```
ismailsunni: Same as comment below.

safe/definitions/hazard_classifications.py
```diff
@@ -568,18 +568,18 @@
...
-            'description': tr('Water is above waist height.'),
+            'description': tr('Flooding is over 150 centimetres.'),
             'fatality_rate': None,
             # displacement rate estimated from DMI analysis of historical
             # flood data and IDP numbers
             'displacement_rate': 0.05,
             'numeric_default_min': 1.5,
             'numeric_default_max': big_number,
-            'string_defaults': ['high'],
+            'string_defaults': ['high, severe'],
```
ismailsunni: Same as comment below

ismailsunni:
Hi @Charlotte-Morgan 
I will merge this and back port it to master for bug fix release. Thanks

PR #4472, opened 2017-09-27
display warning about conflicting plugin

Gustry:
### What does it fix?
* Ticket: None
* Funded by: DFAT
* Description: 
screen shot 2017-09-27 at 10 43 44

I could it better, but it's not my priority ;-)

The problem is if the plugin is enabled, it will even raise an exception in InaSAFE and FloodMapper will be loaded itself, so users think that the problem comes from InaSAFE:

screen shot 2017-09-27 at 10 46 05

No more InaSAFE toolbar, exception from InaSAFE, and FloodMapper is loaded.

### Checklist:

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -97,6 +97,8 @@ from safe.gui.widgets.message import (
     show_no_keywords_message,
     show_keyword_version_message,
     getting_started_message,
+    conflicting_plugin_message,
+    conflicting_plugin_string,
     no_overlap_message,
     ready_message,
     enable_messaging)
@@ -108,6 +110,7 @@ from safe.gui.analysis_utilities import (
     generate_infographic_report,
     add_layer_to_canvas,
     remove_layer_from_canvas)
+from safe.utilities.utilities import is_plugin_installed
 
 __copyright__ = "Copyright 2016, The InaSAFE Project"
 __license__ = "GPL version 3"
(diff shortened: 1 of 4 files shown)


safe/utilities/utilities.py
```diff
@@ -412,3 +414,13 @@ def readable_os_version():
         return ' {version}'.format(version=platform.mac_ver()[0])
     elif platform.system() == 'Windows':
         return platform.platform()
+
+def is_plugin_installed(name):
```
ismailsunni: __[flake8]__
* [E302] expected 2 blank lines, found 1

safe/utilities/utilities.py
```diff
@@ -412,3 +414,13 @@ def readable_os_version():
...
+    """Check if a plugin is installed, even if it's not enabled.
+
+    :param name: Name of the plugin to check.
+    :type name: string
+    """
+    directory = QgsApplication.qgisSettingsDirPath()
+    path = join(directory, 'python', 'plugins', name)
+    return isdir(path)
```
ismailsunni: __[flake8]__
* [W292] no newline at end of file

safe/utilities/utilities.py
```diff
@@ -412,3 +414,13 @@ def readable_os_version():
         return ' {version}'.format(version=platform.mac_ver()[0])
     elif platform.system() == 'Windows':
         return platform.platform()
+
+
+def is_plugin_installed(name):
+    """Check if a plugin is installed, even if it's not enabled.
+
+    :param name: Name of the plugin to check.
+    :type name: string
```
ismailsunni: `returns` is missing

safe/gui/widgets/dock.py
```diff
@@ -841,7 +848,11 @@ def layer_changed(self, layer):
 
         # Do nothing if there is no active layer - see #1861
         if not self._has_active_layer():
-            send_static_message(self, getting_started_message())
+            if is_plugin_installed('EmergencyMapper'):
```
ismailsunni: Perhaps you can also put this if statement as a method. So that you don't need to change the plugin name later if you want to update it. And make a list of the coflicting plugins (if there is more than one).

But, the current one is working great also. My above comment is not mandatory

Gustry: Yes, it what I thought, I wanted to make something clean, but it's very very low priority for me. I doubt we will have another plugin again. And maybe in a few days or weeks, they will fix their plugin so we will have to remove these functions.

PR #4490, opened 2017-10-03
Remove too complex provenance from project variable. Fix #4436

### LINKED ISSUE #4436: Store Complicated Dictionary as HTML String in Project Variable
### Problem
Currently we store complicated dictionary like action list or notes as as normal string with generated key in the project variable. Example of the keys are: `dictionary_key__dictionary_key: value` or `list_key__1`, `list_key__2`. Turns out, this is not easy to be retrieved in the report.

### Proposed Solution
Change to store them as HTML String in the project variable so that it's easy to retrieved in the report.

cc @timlinux @myarjunar

ismailsunni:
### What does it fix?
* Ticket: #4436 
* Funded by: DFAT
* Description: 
   - Remove too complex provenance from project variable. Fix #4436
### Checklist:

### DIFF
--- a/safe/definitions/provenance.py
+++ b/safe/definitions/provenance.py
@@ -278,3 +278,8 @@ provenance_list = [
     provenance_layer_exposure_summary_table_id,
     provenance_layer_profiling_id,
 ]
+
+# Mapping to global variable in QGIS
+duplicated_global_variables = {
+    provenance_os['provenance_key']: 'qgis_os_name'
+}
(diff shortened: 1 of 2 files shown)


Gustry:
Can we also remove duplicate? I think OS, etc

ismailsunni:
@Gustry Done. Currently, there is only `os` that is duplicate. We need to add some variables in the QGIS (perhaps I will do it for QGIS 3.0).

Gustry:
Yes, it could be nice to reduce InaSAFE variables which belong to QGIS.

I don't know if the QPT is using the OS variable already.

ismailsunni:
Thanks @Gustry 
Can you give me comment regarding @Gustry 's comment above @myarjunar ?

ismailsunni:
I have checked with @myarjunar, OS is not used from variable, but from provenance. So, my changes are safe. Will merge this after travis says ok.

Issue #4262, opened 2017-06-12
RFC: Direction and distance tool


Charlotte-Morgan:
We need a new analysis tool that calculates distance and direction from a point hazard. This tool will be based on the functionality in Earthquake  realtime analysis and will allow us to migrate realtime analyses to safe core.

## Requirements:
This tool will calculate the distance and direction from point hazards such as volcano vent and earthquake epicentre to a point exposure such as populated places or critical infrastructure such as airports.

## Limitations:
The aim is to replicate existing realtime analysis not build a mega do everything tool.

timlinux:
@ivanbusthomi some useful links for you:

https://github.com/inasafe/inasafe/blob/realtime/realtime/shake_event.py#L442

This function calls out to various interesting bits - including the cardinality and distances:

        The following fields will be created for each city feature:
            QgsField('id', QVariant.Int),
            QgsField('name', QVariant.String),
            QgsField('population', QVariant.Int),
            QgsField('mmi', QVariant.Double),
            QgsField('dist_to', QVariant.Double),
            QgsField('dir_to', QVariant.Double),
            QgsField('dir_from', QVariant.Double),
            QgsField('roman', QVariant.String),
            QgsField('colour', QVariant.String),
        The 'name' and 'population' fields will be obtained from our geonames
        dataset.
        A raster lookup for each city will be done to set the mmi field
        in the city feature with the value on the raster. The raster should be
        one generated using :func:`mmiDatToRaster`. The raster will be created
        first if needed.
        The distance to and direction to/from fields will be set using QGIS
        geometry API.
        It is a requirement that there will always be at least one city
        on the map for context so we will iteratively do a city selection,
        starting with the extents of the MMI dataset and then zooming
        out by self.zoom_factor until we have some cities selected.
        After making a selection the extents used (taking into account the
        iterative scaling mentioned above) will be stored in the class
        attributes so that when producing a map it can be used to ensure
        the cities and the shake area are visible on the map. See
        :samp:`self.extent_with_cities` in :func:`__init__`.
        .. note:: We separate the logic of creating features from writing a
          layer so that we can write to any format we like whilst reusing the
          core logic.
        .. note:: The original dataset will be modified in place.

ivanbusthomi:
Hi @Charlotte-Morgan and @timlinux ,
this is my to do list for this ticket

- [ ] design the workflow of the tools
- [ ] design the user interface
- [ ] write the function

ivanbusthomi:
[...]

PR #4507, opened 2017-10-06
Update Shakemap Converter

ivanbusthomi:
### What does it fix?
* Ticket: #4262
* Funded by: 
* Description: Replicate realtime functionality in calculating nearby cities affected by adding new feature in the shakemap converter tools.

### Checklist:


ismailsunni:
@ivanbusthomi 
There are pep8 and flake8 violations. Can you fix them? Just run make pep8 or make flake8 to check it locally.
Can you also make a gif / screencast how it works also?

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
```diff
@@ -111,6 +111,25 @@ def on_input_path_textChanged(self):
...
         self.update_warning()
 
+    # noinspection PyPep8Naming
+    def on_input_place_textChanged(self):
+        """Action when input place layer name is changed."""
+        if os.path.exists(self.input_place.text()):
+            try:
+                self.place_layer = QgsVectorLayer(
+                    self.input_place.text(),
+                    'Nearby Cities',
```
Gustry: Use the `tr` function if we keep this layer in the map canvas for the user.

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+    find_distance = QgsDistanceArea()
+    find_distance.setEllipsoidalMode(True)
+    find_distance.setEllipsoid('WGS84')
+
+    fields = places_layer.dataProvider().fields()
+    # list to store cities information
+    cities = []
+    # start calculating distance and get cardinality
+    for feature in places_layer.getFeatures():
+        city_name = str(feature[fields.indexFromName(name)])
```
Gustry: I would avoid the `str` function. What about if the city is called Besançon, the city where I was born ;-)
You will get a python exception ;-)

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+        distance = find_distance.measureLine(hazard_point, city_point)
+        # hazard angle
+        bearing_to = hazard_point.azimuth(city_point)
+        bearing_from = city_point.azimuth(hazard_point)
+        # get direction of the bearing angle
+        direction_to = bearing_to_cardinal(bearing_to)
+        direction_from = bearing_to_cardinal(bearing_from)
+        # store information in a dictionary
+        city = {
+            'name': str(city_name),
```
Gustry: Same here about `str()`

Gustry:
@ismailsunni I wonder if we shouldn't discourage new auto-connect slot in the code? I'm personally not a fan and I prefer an explicit connect. I'm not against in this PR as we never discussed that before.

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
```diff
@@ -111,6 +111,25 @@ def on_input_path_textChanged(self):
...
+                    'ogr'
+                )
+                if self.place_layer.isValid():
+                    LOGGER.debug('Get field information')
+                    self.name_field.setLayer(self.place_layer)
+                    self.population_field.setLayer(self.place_layer)
+                else:
+                    LOGGER.debug('failed to set name field')
+            except:
```
ismailsunni: Can you give comment here why do you pass the exception? Also it's a good practice to use specific exception class, so that we know what exception that we catch.

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+from qgis.core import QgsDistanceArea
+
+LOGGER = logging.getLogger('InaSAFE')
+
+
+def bearing_to_cardinal(angle):
+    """Get cardinality of an angle.
+
+    :param angle: Bearing angle
+    :return:
```
ismailsunni: Give return value, and the type. Also type for the angle. Is it in radian, degree, or gradient ?

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    direction_list = [
```
ismailsunni: Perhaps it's also good to make the list translatable. You don't want to use N in Bahasa Indonesia. Although, I am not sure what's the translation will be.

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+        'N', 'NNE', 'NE', 'ENE', 'E', 'ESE', 'SE',
+        'SSE', 'S', 'SSW', 'SW', 'WSW', 'W',
+        'WNW', 'NW', 'NNW'
+    ]
+
+    bearing = float(angle)
+    direction_count = len(direction_list)
+    direction_interval = 360. / direction_count
+    index = int(round(bearing / direction_interval))
```
ismailsunni: Perhaps you can put reason why do you use round here. Not using `ceiling` or `floor`. And make sure you have unit test for any function that use `round`. It will help someone in the future who will migrate to python 3, because the behaviour is little bit different.
```python
# python 2.x
In [7]: [round(x) for x in [0.1, 0.5, 0.7, 1.1, 1.5, 1.7, 2, 2.1, 2.5]]
Out[7]: [0.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 3.0]

# python 3
>>> [round(x) for x in [0.1, 0.5, 0.7, 1.1, 1.5, 1.7, 2, 2.1, 2.5]]
[0, 0, 1, 1, 2, 2, 2, 2, 2]
```

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+    # list to store cities information
+    cities = []
+    # start calculating distance and get cardinality
+    for feature in places_layer.getFeatures():
+        city_name = str(feature[fields.indexFromName(name)])
+        if not population:
+            # give zero value if there is no population fields given
+            population_number = 0
+        else:
+            population_number = str(feature[fields.indexFromName(population)])
```
ismailsunni: avoid `str`

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+        # hazard angle
+        bearing_to = hazard_point.azimuth(city_point)
+        bearing_from = city_point.azimuth(hazard_point)
+        # get direction of the bearing angle
+        direction_to = bearing_to_cardinal(bearing_to)
+        direction_from = bearing_to_cardinal(bearing_from)
+        # store information in a dictionary
+        city = {
+            'name': str(city_name),
+            'population': int(population_number),
```
ismailsunni: May be you can use the rounding function that we have already for population.

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,84 @@
...
+            'distance': float(distance),
```
ismailsunni: Any reason why using float here?

ismailsunni:
> @ismailsunni I wonder if we shouldn't discourage new auto-connect slot in the code? I'm personally not a fan and I prefer an explicit connect. I'm not against in this PR as we never discussed that before.

Me too. I prefer the explicit way.

ivanbusthomi:
@ismailsunni 
> There are pep8 and flake8 violations. Can you fix them? Just run make pep8 or make flake8 to check it locally.
Can you also make a gif / screencast how it works also?

OK, will do it

> I wonder if we shouldn't discourage new auto-connect slot in the code? I'm personally not a fan and I prefer an explicit connect. I'm not against in this PR as we never discussed that before.

By autoconnect slot, do you mean like this?
```
def on_input_path_textChanged(self):
    ...
```
I can change it to use the explicit connect one , I also think it will also make it easier to see what are the connections that have been created in the code.

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
ivanbusthomi: sorry I forgot to use the `tr`, will include it in my next commit

safe/utilities/direction_distance.py
ivanbusthomi: Hmmm, I am not aware if that will give me an exception. Thanks for the information.

safe/utilities/direction_distance.py
ivanbusthomi: ok

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
ivanbusthomi: I pass it because I don't have any idea yet on what I will do :sweat_smile: 
I will include it in my next commit. do you have any suggestion for me?

safe/utilities/direction_distance.py
ivanbusthomi: Sure, I will provide more complete information in the docstring

safe/utilities/direction_distance.py
ivanbusthomi: I think we only have 8 well-known directions, while in the list we have 18. Or maybe for Indonesia, we limit it to just 8 directions? I am not sure whether it is doable though

safe/utilities/direction_distance.py
ivanbusthomi: That code is the same code that realtime use in shake_event.py
https://github.com/inasafe/inasafe-realtime/blob/10f34187a3ace62817173537288ff675a6806766/src/realtime/earthquake/shake_event.py#L1635

I am not particularly sure if there is any particular reason in using rounding. I will try changing it

safe/utilities/direction_distance.py
ivanbusthomi: ok

safe/utilities/direction_distance.py
ivanbusthomi: I was expecting of having non integer result from the distance, and round it later when calling it in the keywords creation. Maybe I could just change it to string here so I won't have to modify it outside.

Gustry:
> Me too. I prefer the explicit way.

Ok, cool. Let's make it deprecated today ;-) Even QGIS is doing it https://github.com/qgis/QGIS/pull/5326

> By autoconnect slot, do you mean like this?

yes @ivanbusthomi this kind of connect is better in the constructor:
`self.field_list.itemSelectionChanged.connect(self.update_footer)`

Do you think you can provide some tests for `safe/utilities/direction_distance.py` ? The first function is easy ;-)

ivanbusthomi:
> Do you think you can provide some tests for safe/utilities/direction_distance.py ? The first function is easy ;-)

sure, I will include it once I understand what unittest and how it works :)

Gustry:
You can just have a look to a easy test like this one: https://github.com/inasafe/inasafe/blob/develop/safe/utilities/test/test_utilities.py#L44

We are just calling the function and test the result.

ivanbusthomi:
Hi @ismailsunni,
> There are pep8 and flake8 violations. Can you fix them? Just run make pep8 or make flake8 to check it locally.

any idea how to do that on windows? this is the gif of how it works by the way 
demo

Gustry:
Quite a long GIF, but it's very nice! Great.

You can see your error on travis: https://travis-ci.org/inasafe/inasafe/builds/287029985#L1516
It's only one. And I see you already fixed in your commit ;-)

ivanbusthomi:
@Gustry thanks, sorry it was quite long. I just recorded the whole process :D

ismailsunni:
> any idea how to do that on windows?

Not an expert, but I think you can just install pep8 and flake8 by using pop. May be something like this https://stackoverflow.com/questions/29371951/installing-pep8-on-windows

ivanbusthomi:
@ismailsunni 
> Not an expert, but I think you can just install pep8 and flake8 by using pip.

I already installed them using pip, but I am not really familiar with running `make` in windows

Gustry:
If you have installed Pep8, you might skip the `make` command.
"Make" is just a way to save the command in a file, a kind of alias. By memory, you can go to safe/test/test_python_pep to see inside for windows, made by Tim. You can the run pep8 directly.
You can also launch this test file in Pycharm.

safe/gui/tools/shake_grid/shake_grid.py
```diff
@@ -821,6 +835,85 @@ def set_contour_properties(self, input_file):
 
         layer.commitChanges()
 
+    def mmi_point_sampling(self, place_layer, raster_mmi_path):
+        """Get MMI information from ShakeMap Tiff on place location.
+
+        Get the MMI value on each cities/places from the generated tiff files.
```
ismailsunni: Add new line to make it more readable.

safe/gui/tools/shake_grid/shake_grid.py
```diff
@@ -821,6 +835,85 @@ def set_contour_properties(self, input_file):
...
+        :param place_layer: Point layer generated from distance calculation
+        :type place_layer: QgsVectorLayer
```
ismailsunni: New line between params

safe/gui/tools/shake_grid/shake_grid.py
```diff
@@ -821,6 +835,85 @@ def set_contour_properties(self, input_file):
...
+        :param raster_mmi_path: Path to converted mmi raster.
+        :type raster_mmi_path: str
```
ismailsunni: New line between param and returns

safe/gui/tools/shake_grid/shake_grid.py
```diff
@@ -821,6 +835,85 @@ def set_contour_properties(self, input_file):
...
+            key=itemgetter('mmi_values', 'population'),
+            reverse=True
+        )
+        nearest_place = sorted_places[0]
+        # combine locality text
+        city_name = nearest_place['name']
+        city_distance = nearest_place['distance']
+        city_bearing = nearest_place['bearing_to']
+        city_direction = nearest_place['dir_to']
+        self.locality = 'Located ' + str(math.floor(city_distance / 1000)) + \
```
ismailsunni: Put it in `tr`. And use new string formatter to make it easy to translate.
e.g. 
```python
tr(
            'In this wizard step: {step_name}, you will be able to '
            'set a field that corresponded with a InaSAFE field '
            'concept.').format(step_name=self.step_name)

```

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,103 @@
...
+    feature_list = []
+    for feature in places_layer.getFeatures():
+        feature_list.append(feature)
+    output_provider.addFeatures(feature_list)
+    # create new fields to store the calculation result
+    output_provider.addAttributes([
+        QgsField('distance', QVariant.Double),
+        QgsField('bearing_to', QVariant.Double),
+        QgsField('dir_to', QVariant.String),
+        QgsField('bearing_fr', QVariant.Double),
```
ismailsunni: Use full word. Or is it must a shape file?

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,103 @@
...
+        QgsField('dir_from', QVariant.String),
+        QgsField('mmi', QVariant.Double)
+    ])
+    output_layer.updateFields()
+    # get field index
+    distance_index = output_provider.fields().indexFromName('distance')
```
ismailsunni: Even better if you use variable name to make it easy to refactor or avoid mis-typing

safe/utilities/test/test_direction_distance.py
```diff
@@ -0,0 +1,31 @@
+import unittest
```
ismailsunni: Docstring

safe/utilities/direction_distance.py
ivanbusthomi: if I remember correctly, field name is limited to only 10 characters. It is a memory layer, and later I save it as geojson

safe/gui/tools/shake_grid/shake_grid.py
ivanbusthomi: OK

safe/gui/tools/shake_grid/shake_grid.py
ivanbusthomi: OK

safe/utilities/direction_distance.py
ismailsunni: I think it's only limitation if we use shapefile. In GeoJSON you can have more than 10 chars field name.

safe/gui/tools/shake_grid/shake_grid.py
```diff
@@ -821,6 +837,99 @@ def set_contour_properties(self, input_file):
 
         layer.commitChanges()
 
+    def mmi_point_sampling(self, place_layer, raster_mmi_path):
+        """Get MMI information from ShakeMap Tiff on place location.
+
+        Get the MMI value on each cities/places from the generated tiff files.
+
+        :param place_layer: Point layer generated from distance calculation
```
Gustry: We try to add a full stop at the end of each `:param` and `:return:`. Better if it can be done, but no worries ;-)

safe/gui/tools/shake_grid/shake_grid.py
```diff
@@ -821,6 +837,99 @@ def set_contour_properties(self, input_file):
...
+        :param raster_mmi_path: Path to converted mmi raster.
+        :type raster_mmi_path: str
+
+        :return: Modified place layer containing MMI values
+        :rtype: QgsVectorLayer
+        """
+        shake_raster = QgsRasterLayer(raster_mmi_path, 'shake_raster')
+        shake_provider = shake_raster.dataProvider()
+
+        place_fields = place_layer.dataProvider().fields()
```
Gustry: `place_fields = place_layer.fields()` No need for to fetch the dataprovider

safe/gui/tools/shake_grid/test/test_shakemap_converter_dialog.py
```diff
@@ -60,13 +60,22 @@ def test_converting(self):
             '20131105060809',
             'output',
             'grid.xml')
+        place_path = standard_data_path(
+            'exposure',
+            'places_shakemap.shp'
```
Gustry: We also try to move away from shapefile and go for geojson. It's one file, easier to edit in PyCharm, and we can see them in GitHub easily: https://github.com/inasafe/inasafe/blob/develop/safe/test/data/gisv4/exposure/building-points.geojson

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,118 @@
...
+        direction_to = bearing_to_cardinal(bearing_to)
+        direction_from = bearing_to_cardinal(bearing_from)
+        # store the value to attribute table
+        output_layer.changeAttributeValue(fid, distance_index, distance)
+        output_layer.changeAttributeValue(fid, bearing_to_index, bearing_to)
+        output_layer.changeAttributeValue(fid, dir_to_index, direction_to)
+        output_layer.changeAttributeValue(fid, bearing_fr_index, bearing_from)
+        output_layer.changeAttributeValue(fid, dir_from_index, direction_from)
+    output_layer.commitChanges()
+    return output_layer
```
Gustry: Just a question, maybe I didn't follow enough, we don't need to create keywords with definitions with these new fields? @ismailsunni ?

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,118 @@
...
+    find_distance.setEllipsoidalMode(True)
+    find_distance.setEllipsoid('WGS84')
+
+    fields = places_layer.dataProvider().fields()
+
+    # memory layer to save new place layer
+    output_layer = create_memory_layer(
+        'Nearby Places', QGis.Point, places_layer.crs(), fields
+    )
+    output_provider = output_layer.dataProvider()
```
Gustry: Can we skip using the provider? Just work on the QgsVectorLayer

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,118 @@
...
+    # define distance
+    find_distance = QgsDistanceArea()
+    find_distance.setEllipsoidalMode(True)
+    find_distance.setEllipsoid('WGS84')
+
+    fields = places_layer.dataProvider().fields()
+
+    # memory layer to save new place layer
+    output_layer = create_memory_layer(
+        'Nearby Places', QGis.Point, places_layer.crs(), fields
```
Gustry: No need for `tr`?

safe/utilities/direction_distance.py
ismailsunni: Ouch, thanks @Gustry . We do need to create the definiton. Can you check the `safe/definitions/fields.py` @ivanbusthomi ? Just follow the other definition there, and put the created fields in the `inasafe_fields` keyword.
It's quite important actually.

safe/gui/tools/shake_grid/shake_grid.py
ivanbusthomi: OK

safe/gui/tools/shake_grid/shake_grid.py
ivanbusthomi: Thanks for reminding me, I must've missed that

safe/gui/tools/shake_grid/test/test_shakemap_converter_dialog.py
ivanbusthomi: OK, I will change the place shp to the geojson one

ivanbusthomi:
@ismailsunni and @Gustry 
after doing a rebase to develop, I run some tests and received this error. looking at the error log, it did not redirect to any changes that I've made. Any idea on how to fix it?
```
  File "C:\Program Files\JetBrains\PyCharm 2017.2.3\helpers\pycharm\_jb_unittest_runner.py", line 35, in <module>
    main(argv=args, module=None, testRunner=unittestpy.TeamcityTestRunner, buffer=not JB_DISABLE_BUFFERING)
  File "C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\unittest\main.py", line 94, in __init__
    self.parseArgs(argv)
  File "C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\unittest\main.py", line 149, in parseArgs
    self.createTests()
  File "C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\unittest\main.py", line 158, in createTests
    self.module)
  File "C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\unittest\loader.py", line 128, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "C:\PROGRA~1\QGIS2~1.14\apps\Python27\lib\unittest\loader.py", line 91, in loadTestsFromName
    module = __import__('.'.join(parts_copy))
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\gui\tools\shake_grid\test\test_shake_grid.py", line 9, in <module>
    from safe.common.utilities import unique_filename, temp_dir
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\__init__.py", line 29, in <module>
    from safe.common.custom_logging import setup_logger
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\common\custom_logging.py", line 34, in <module>
    from safe.common.utilities import log_file_path  # NOQA
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\common\utilities.py", line 25, in <module>
    from safe.utilities.rounding import (
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\utilities\rounding.py", line 10, in <module>
    from safe.definitions.units import unit_mapping, nominal_mapping
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\definitions\__init__.py", line 19, in <module>
    from safe.definitions.hazard import *
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\definitions\hazard.py", line 13, in <module>
    from safe.definitions.hazard_classifications import (
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\definitions\hazard_classifications.py", line 157, in <module>
    'Options.' % current_earthquake_model_name()),
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\definitions\earthquake.py", line 364, in current_earthquake_model_name
    'earthquake_function', expected_type=str)
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\utilities\settings.py", line 124, in setting
    return general_setting(full_key, default, expected_type, qsettings)
  File "C:\Users\Ivan Busthomi\.qgis2\python\plugins\inasafe\safe\utilities\settings.py", line 63, in general_setting
    return qsettings.value(key, default, expected_type)
TypeError: unable to convert a QVariant of type 0 to a QMetaType of type 10
```

Gustry:
Which test are you running?
On windows?

ivanbusthomi:
I am running the test_shake_grid one, and yes on windows.

Gustry:
Which test specifically?
The file `test_shake_grid` is running fine for me.

In QGIS, Settings, Advanced and then InaSAFE folder, can you tell us everything about the value `earthquake_function`?

Can you go in InaSAFE settings and change your EQ function to another one ?

safe/definitions/earthquake.py
```diff
@@ -361,7 +361,7 @@ def current_earthquake_model_name():
         settings.
     """
     default_earthquake_function = setting(
-        'earthquake_function', expected_type=str)
+        'earthquake_function', EARTHQUAKE_FUNCTIONS[0]['key'],str)
```
ismailsunni: space after comma

safe/definitions/fields.py
```diff
@@ -2505,6 +2505,109 @@
...
+    'type': QVariant.Double,
+    'length': default_field_length,
+    'precision': default_field_precision,
+    'absolute': True,
+    'description': tr(
+        'The distance value between place and hazard point.'
+    ),
+    'help_text': tr(
+        'Distance value between place feature to the epicenter of the hazard. '
+        'The distance is calculated using WGS84 as ellipsoid model. The unit of'
```
ismailsunni: Make sure to add `space` character before the quote to avoid missing `space`.

safe/definitions/fields.py
```diff
@@ -2505,6 +2505,109 @@
...
+    'header_name': tr('Bearing'),
+    'type': QVariant.Double,
+    'length': default_field_length,
+    'precision': default_field_precision,
+    'absolute': False,
+    'description': tr(
+        'An azimuth angle at a place to a hazard point.'),
+    'help_text': tr(
+        'A bearing angle is an angle measured to a point as observed in current'
+        ' location using north as a reference direction. In this case, "bearing'
```
ismailsunni: Always put `space` character in the end of string, not in the begining of string. I see there are some more.

safe/definitions/fields.py
```diff
@@ -2505,6 +2505,109 @@
...
+            'link': None
+        }
+    ]
+}
+
+# TODO: Ivan to update
+direction_field = {
+    'key': 'direction_field',
+    'name': tr('Direction'),
+    'field_name': 'dir_from',
```
ismailsunni: Why not use a full name? Like `direction_from`. And why not only `direction`?

safe/utilities/direction_distance.py
```diff
@@ -0,0 +1,120 @@
...
+)
+from safe.utilities.i18n import tr
+
+LOGGER = logging.getLogger('InaSAFE')
+
+
+def bearing_to_cardinal(angle):
+    """Get cardinality of an angle.
+
+    :param angle: Bearing angle
```
ismailsunni: Dot in the end of sentence.

safe/definitions/fields.py
ivanbusthomi: OK, I will update it.

safe/utilities/direction_distance.py
ivanbusthomi: OK, I will update it.

safe/definitions/fields.py
ivanbusthomi: OK, I will update it.

safe/definitions/fields.py
ivanbusthomi: AFAIK, there is a limitation on how many numbers of character a field name can have, which is around 10 character IIRC. 

The reason I also add `from` and not only leave it as `direction` because it is a relative value which depends on where our base point is. For example, if the azimuth from A to B is 90 degree, then the `direction_from` A to B would be East, and the `direction_to` A from B would be West.

**But**, since at the moment we only use one of them (`direction from`, to get the direction value from a hazard point to places points), I think I will change it to use `direction` and add some more information in the help text.

Gustry:
Do you want to add more commits or do you want a review?

Thanks for all your commits ;-)

ivanbusthomi:
@Gustry 
> Do you want to add more commits or do you want a review?

I am adding one more commit to address @ismailsunni reviews above.

safe/definitions/fields.py
```diff
@@ -2505,6 +2505,108 @@
     'units': currencies,
 }
 
+# # # # # # # # # #
+# Direction and Distance Field
+# # # # # # # # # #
+
+# Count for each exposure type
```
Gustry: unrelated comment

ivanbusthomi:
Thanks, I am happy if you want to merge it.

ivanbusthomi:
Thanks,

Gustry:
I will let @ismailsunni comment and/or merge ;-)

ismailsunni:
Hi @ivanbusthomi 
There are still 4 comments that haven't been addressed (easy one). Can you give update about it?

safe/gui/tools/shake_grid/shake_grid.py
ismailsunni: Ah, this is already done.

safe/utilities/test/test_direction_distance.py
ismailsunni: Done also.

Gustry:
If you talk about docstrings, new line, they have been fixed I think

ismailsunni:
Ah sorry, the diff didn't show full code (since the line that I commented didn't get any change :D)

Gustry:
Yes, because the docstring or new line don't affect the line where you put your comment :D 

We try to move away from shapefile, but we didn't make it hard requirements.

Issue #4486, opened 2017-09-29
Need to check that every client of an IF have the same behaviour


Gustry:
### Problem

We have many clients of an Impact Function:
* dock
* wizard
* CLI
* batch runner
* multi-exposure

There is too much logic today in the dock about what to do after an Impact Function has run.
It brings some discrepancies. For instance, infographics are working only if we run the IF from the dock.

We should refactor the code by using either the IF or having a function to call in all of these clients to be sure we have the same behavior.

Maybe the first step is to fix the IFCW in the master branch to get back infographics. @myarjunar cam you have a look here?

Gustry:
Have a look to this ticket: https://github.com/inasafe/inasafe/issues/3001

timlinux:
@Gustry can take the lead in tasking @ismailsunni and @myarjunar with subtasks and help to get this done for 4.3

Issue #3001, opened 2016-06-30
Duplicate code between the dock and the wizard


Gustry:
### Problem
Thanks to @NyakudyaA, I noticed that when an analysis is finished using IFCW, the behaviour is not the same than when we are using the dock.
In the IFCW, I found that a lot of code has been copied from the dock. However, for example I brought some fix on the dock these last weeks, but I wasn't aware about that duplication in the wizard.

### Proposed solution
We should avoid duplication in the code between the dock and the analysis_handler

Quick thoughts about some inheritance ? 
        inasafe_gui
        /        \
dock      wizard

Of course, the dock will still have a lot of code, it's just to avoid duplication with the analysis_handler.

timlinux:
Hmm that would be interesting...I don't know if it makes sense to inherit everything from dock. What about a shared base class that is used by both? That would also help to make dock smaller....

ismailsunni:
Hi @timlinux and @Gustry 
I have another suggestion for this issue:
In the wizard, we use AnalysisHandler class to run the impact function. It only needs to :
1. Create the analysis handler
2. Setup and run the Impact Function
3. Set action for print and report composer button.

I think, we can do it also in Dock. So, Dock, will not call the Impact Function directly, but it will do the same  as the Wizard. This means, AnalysisHandler is an interface to run Impact Function.]By doing this, we will have these benefits:

1. No more code duplication, since all code to set up and run IF will be in the AnalysisHandler
2. Dock will have less code, since we move out the IF setup and run.
3. Better concern separation. Dock and Wizard will only manage the UI things (choose layer, click button, show result from the AnalysisHandler, etc)
4. We can add more user interface in the future more easily. 

The cons so far:
1. We need to send signal from the AnalysisHandler to the Wizard or the Dock so that they will aware the state of the analysis. But it's already done in the Wizard - AnalysisHandler.

Any comment for this idea?

timlinux:
@ismailsunni sorry if its a bomb question - but why do we need analysis handler at all? Cant IF just handle all the activities of model execution + print production? I think it will be much cleaner and easier to explain / train people with...

ismailsunni:
No worries, I like to discuss to prevent future problem :D 

- model execution, yes the IF can do it. The AnalysisHandler will help to get the parameter to setup from dock and wizard. It's not a big benefit of using AnalysisHandler.
[...]

Issue #4504, opened 2017-10-05
Infographic report can not produce with the batch runner


felix-yew:
### Problem
I can not find the infographic report result when I run the InaSAFE population analysis with Batch Runner.

### Expected Behaviour
The infographic report can produce with the batch runner. It very helpful if we want to run a lot of population analysis.

### Environment
* InaSAFE : 4.3.0 dev-master
* QGIS : 2.14.14
* OS : Win 7 Pro/64bit

### Proposed Solution
Bring back the infographic report.

Cc: @Charlotte-Morgan @samnawi @ivanbusthomi @ismailsunni @myarjunar

Gustry:
This is a duplicate of https://github.com/inasafe/inasafe/issues/4486

@myarjunar Can you have a look to this one? It's a bug on master.

PR #4516, opened 2017-10-11
Change logic for generating report to one function

myarjunar:
### What does it fix?
* Ticket: #4486 
* Funded by: DFAT
* Description: Previously, we have a logic of generating report written in each IF client with the same code just like in #3001, now we want to create only one method and write all logic to generate a report there and use only that function in each client of IF to generate report.

### Checklist:

### DIFF
--- a/safe/gui/analysis_utilities.py
+++ b/safe/gui/analysis_utilities.py
@@ -46,13 +46,10 @@ def generate_report(impact_function, iface):
 
     # don't generate infographic if exposure is not population
     exposure_type = layer_definition_type(impact_function.exposure)
+    map_overview_layer = None
     if exposure_type != exposure_population:
         report_components.remove(infographic_report)
-
-    # if exposure is population, we need to add
-    # map overview layer
-    exposure_type = layer_definition_type(impact_function.exposure)
-    if exposure_type == exposure_population:
+    else:
         map_overview_layer = QgsRasterLayer(
             map_overview['path'], 'Overview')
         add_layer_to_canvas(
@@ -118,7 +115,8 @@ def generate_report(impact_function, iface):
         if error_code == ImpactReport.REPORT_GENERATION_FAILED:
             break
 
-    remove_layer_from_canvas(map_overview_layer, impact_function)
+    if map_overview_layer:
+        remove_layer_from_canvas(map_overview_layer, impact_function)
 
     return error_code, message
 
--- a/safe/gui/analysis_utilities.py
+++ b/safe/gui/analysis_utilities.py
@@ -4,17 +4,21 @@
 
 import os
 from collections import OrderedDict
+from copy import deepcopy
 from PyQt4.QtCore import QDir, Qt
 from qgis.core import (
     QgsMapLayerRegistry,
     QgsProject,
-    QgsRasterLayer)
+    QgsRasterLayer,
+    QgsMapLayer,
+    QGis)
 
 from safe.definitions.exposure import exposure_population
 from safe.definitions.reports.infographic import map_overview
 from safe.definitions.utilities import definition, update_template_component
 from safe.definitions.fields import hazard_class_field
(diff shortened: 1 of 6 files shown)


Gustry:
I guess it's also linked to issue https://github.com/inasafe/inasafe/issues/3001 and https://github.com/inasafe/inasafe/issues/4504

Thanks @myarjunar I'm going to have a look

safe/gui/tools/batch/batch_dialog.py
```diff
@@ -506,18 +510,25 @@ def run_task(self, task_item, status_item, count=0, index=''):
...
                         try:
-                            # this line is to save the impact report in default
-                            # InaSAFE directory.
-                            generate_impact_report(impact_function, self.iface)
-                            generate_impact_map_report(
-                                impact_function,
-                                self.iface)
+                            # generate report from component definition and
+                            # save it in the default output directory
+                            report_components = deepcopy(
```
Gustry: Can you explain why we need a deep copy here?

safe/gui/tools/batch/batch_dialog.py
```diff
@@ -506,18 +510,25 @@ def run_task(self, task_item, status_item, count=0, index=''):
...
+                                all_default_report_components)
+
+                            # don't generate infographic if exposure is not
+                            # a population
+                            exposure_type = layer_definition_type(
+                                impact_function.exposure)
+                            if exposure_type != exposure_population:
```
Gustry: Ok, I understand about the deep copy now. But can't we move this logic into the function? It's still code that need to be duplicated between dock, IFCW, batch runner ...

safe/gui/analysis_utilities.py
```diff
@@ -26,151 +32,95 @@
...
+
+    # don't generate infographic if exposure is not population
+    exposure_type = layer_definition_type(impact_function.exposure)
+    if exposure_type != exposure_population:
+        report_components.remove(infographic_report)
+
+    # if exposure is population, we need to add
+    # map overview layer
+    exposure_type = layer_definition_type(impact_function.exposure)
+    if exposure_type == exposure_population:
```
Gustry: * Remove duplicated line L48 and L55.
* Use an `else` statement instead two IF doing the reverse

safe/gui/analysis_utilities.py
```diff
@@ -26,151 +32,95 @@
...
-        # No other way for now
-        return
+        # We will generate it on the fly without storing it after datastore
+        # supports
+        impact_report.output_folder = os.path.join(layer_dir, 'output')
+        error_code, message = impact_report.process_components()
+        if error_code == ImpactReport.REPORT_GENERATION_FAILED:
+            break
+
+    remove_layer_from_canvas(map_overview_layer, impact_function)
```
Gustry: `map_overview_layer` will raise an exception if it's not a population analysis because it seems uninitialized!
Correct me if I'm wrong, it's difficult to read code on github

safe/gui/analysis_utilities.py
myarjunar: upsy

safe/gui/tools/batch/batch_dialog.py
```diff
@@ -506,18 +504,13 @@ def run_task(self, task_item, status_item, count=0, index=''):
...
+                                impact_function, self.iface)
+
                             # this line is to save the report in user specified
                             # directory.
                             self.generate_pdf_report(
-                                impact_function,
-                                self.iface,
-                                group_name)
+                                impact_function, self.iface, group_name)
                         except:
```
ismailsunni: May be set the expected exception here

Gustry: I guess it's `Exception` because we need to catch all exceptions. I'm fine for that.

Gustry:
Will you change inasafe_cli too? I'm not sure about this repo

Issue #4509, opened 2017-10-09
We need to serialise the impact function


timlinux:
We need to be able to serialise and deserialise impact function. Then these steps:

- [ ] Make sure the impact layer xml includes all needed properties of the IF
- [ ] We need a second IF associated with teh dock 'current_impact_function' which is the state of the current impact layer group
- [ ] We need to ba able to call the reporting engine, first tweaking the apt if needed on the deserialised IF

ismailsunni:
Some more detail tasks for myself:
- [ ] Make a comparison method / operator to check if two IF is the same
- [ ] Unit test
- [ ] Loading IF from output layer's metadata

timlinux:
This should fix #4240

timlinux:
Once the deserialisation is done, @myarjunar to implement print as a separate step from the analysis with the ability for user to first select a QPT before running print:

![screen shot 2017-10-16 at 09 33 41](https://user-images.githubusercontent.com/178003/31600360-6973ddf8-b256-11e7-9154-14b80d77de76.png)

PR #4521, opened 2017-10-16
[WIP] IF serialization

ismailsunni:
### What does it fix?
* Ticket: #4509 
* Funded by: DFAT
* Description: 
   - Enable serialization of IF.

### Checklist:


Gustry:
This looks a lot of code ! Awesome

Can you provide a small python snippet how it looks like from the client?
The is the output of the serialization? and what do you need explicitly to make it runnable again?

Do you need these analysis log.csv in the PR?

ismailsunni:
Hi @Gustry 
> Can you provide a small python snippet how it looks like from the client?

There is a unit test on how to use it in `test_impact_function.py`. Basically, you use it like this:

```python
loaded_impact_function = ImpactFunction.load_from_output_metadata(output_metadata)
```

Where `output_metadata` is any output layer's full keywords/metadata. After that, you will get the impact_function.

The serialization is basically the `provenance_data`.

> Do you need these analysis log.csv in the PR?

It's almost done, so I want to include it also (rather than visiting it again later in the future).

ismailsunni:
I think this PR is almost finished. But I need @myarjunar to merge his work first regarding using provenance to generate the report (currently it is using the IF.hazard or IF.exposure, which I changed).

safe/definitions/provenance.py
```diff
@@ -13,221 +13,226 @@
...
 }
 provenance_layer_exposure_summary_table_id = {
-    'key': 'provenance_layer_exposure_summary_table_id',
+    'key': u'provenance_layer_exposure_summary_table_id',
     'name': layer_purpose_exposure_summary_table['name'] + ' ID',
     'provenance_key': layer_purpose_exposure_summary_table['key'] + '_id'
 }
 provenance_layer_profiling_id = {
-    'key': 'provenance_layer_profiling_id',
+    'key': u'provenance_layer_profiling_id',
```
Gustry: Just curious, why do we need the `u`? It's only plain text.

safe/definitions/test/test_utilities.py
```diff
@@ -492,6 +494,22 @@ def test_set_provenance(self):
         }
         self.assertDictEqual(provenance_collection, expected)
 
+    def test_get_provenance(self):
+        """Test fot get_provenance"""
```
Gustry: Full stop is better ;-)
But I don't mind for test file

safe/gis/utilities.py
```diff
@@ -0,0 +1,64 @@
+# coding=utf-8
+"""Common utilities function for GIS."""
```
Gustry: Any difference between 'safe/gis/tools.py' ? Why not in this one?

safe/gis/utilities.py
```diff
@@ -0,0 +1,64 @@
...
+
+def load_layer(layer_path):
+    """Helper to load and return a single QGIS layer
+
+    :param layer_path: Path name to raster or vector file.
+    :type layer_path: str
+
+    :returns: tuple containing layer and its layer_purpose.
+    :rtype: (QgsMapLayer, str)
+
```
Gustry: No need blank line

safe/gis/utilities.py
```diff
@@ -0,0 +1,64 @@
...
+__copyright__ = "Copyright 2017, The InaSAFE Project"
+__license__ = "GPL version 3"
+__email__ = "info@inasafe.org"
+__revision__ = '$Format:%H$'
+
+LOGGER = logging.getLogger('InaSAFE')
+
+
+def load_layer(layer_path):
+    """Helper to load and return a single QGIS layer
```
Gustry: Full stop

safe/gis/utilities.py
```diff
@@ -0,0 +1,64 @@
...
+        layer = QgsVectorLayer(layer_path, base_name, 'ogr')
+    elif extension in ['.csv']:
+        uri = 'file:///%s?delimiter=%s' % (layer_path, ',')
+        layer = QgsVectorLayer(uri, base_name, 'delimitedtext')
+    else:
+        message = 'File %s had illegal extension' % layer_path
+        raise Exception(message)
+
+    # noinspection PyUnresolvedReferences
+    message = 'Layer "%s" is not valid' % layer.source()
```
Gustry: Can you move this var in the IF?

safe/impact_function/impact_function.py
```diff
@@ -18,7 +18,8 @@
     QgsCoordinateReferenceSystem,
     QgsRectangle,
     QgsVectorLayer,
-    QGis
+    QGis,
+    QgsMapLayer
```
Gustry: Small trick, but I always add the comma, even if it's the last one. You have one line less in your github PR and in git blame for the next time.

safe/impact_function/impact_function.py
```diff
@@ -675,7 +756,7 @@ def callback(self):
 
             self.progress_callback(current, maximum, message=None)
 
-        :rtype: function
+        :rtype: function()
```
Gustry: hum, if I read this docstring, I might understand that the function is evaluated in the return.

and according to Python, it's a function:
```
def foo(): return '1'
type(foo)
<type 'function'>
type(1)
<type 'int'>
```

safe/impact_function/impact_function.py
```diff
@@ -2356,3 +2466,118 @@ def append_ISO19115_keywords(keywords):
...
+        :type output_metadata: OutputLayerMetadata
+
+        :returns: Impact Function based on the metadata.
+        :rtype: ImpactFunction
+        """
+        impact_function = ImpactFunction()
+        provenance = output_metadata['provenance_data']
+        # Set exposure layer
+        exposure_path = get_provenance(provenance, provenance_exposure_layer)
+        if exposure_path:
```
Gustry: Do we need these 'IF' ? AFAIK, this path will be always defined, or we can't create the IF and we should stop the process. No?

safe/impact_function/test/test_impact_function.py
```diff
@@ -1149,6 +1153,61 @@ def _check_minimum_fields_exists(self, impact_function):
                     inasafe_fields,
                     message.format(field_key=field['key']))
 
+    def test_equality(self):
+        """Testing IF equal operator."""
+        new_impact_function = ImpactFunction()
```
Gustry: Hum, where this variable come from?

safe/impact_function/test/test_impact_function.py
```diff
@@ -1149,6 +1153,61 @@ def _check_minimum_fields_exists(self, impact_function):
...
+        self.assertEqual(new_impact_function, new_impact_function)
+
+    @unittest.skipIf(
+        not os.path.exists('/Users/ismailsunni/dev/python/inasafe-dev'),
```
Gustry: best condition ever ;-)

safe/utilities/test/test_styling.py
```diff
@@ -26,9 +26,9 @@
 from safe.test.utilities import (
```
Gustry: Ahah, thanks for spotting this file. This file can easily removed from InaSAFE and the linked functions. I will do a separate PR to see

safe/definitions/provenance.py
ismailsunni: Ah, it is because I compare dictionary. This one was using string, the other one is using unicode. I forget that you have created that nice bytefy function, so I used it. But I forget to revert this changes. I will change it back like before.

safe/definitions/test/test_utilities.py
ismailsunni: Thanks :D

safe/impact_function/impact_function.py
ismailsunni: Done.

safe/impact_function/impact_function.py
ismailsunni: I am not sure also for this. But PyCharm shows a warning if I only use `function` only. It says it's not callable. Somehow, I put the brackets, and the warnings are gone. What's the situtation in your PyCharm?

safe/impact_function/test/test_impact_function.py
ismailsunni: what do you mean? It's a declaration.

safe/impact_function/test/test_impact_function.py
ismailsunni: :D I am the only one who can run the test :D

safe/utilities/test/test_styling.py
ismailsunni: Thanks :)

safe/gis/utilities.py
ismailsunni: Done

safe/gis/utilities.py
ismailsunni: Done

safe/gis/utilities.py
ismailsunni: I took it from `safe/test/utilities.py`, since I wanted to move it to non-test, and the best one is `gis`, so I just naturally put it in `safe/gis/utilities` *same file name. But, i think it's better to put it in `tools.py`, less file :). I wasn't aware that tools.py is similar 

I will move it.

safe/gis/utilities.py
ismailsunni: Ouch, I thought IF = Impact Function. I thought you asked me to move it to the `impact_function.py`:D

Done also.

safe/impact_function/impact_function.py
ismailsunni: I am not sure either actually.
If the user clean or move the original layers of input layers, we can skip this. Actually, we only need the output's layer to generate the report.
But, honestly, I am not sure about what's the best behaviour.

ismailsunni:
@myarjunar see, all is good :D

safe/impact_function/impact_function.py
```diff
@@ -312,6 +315,10 @@ def __eq__(self, other):
             'profiling',
         ]
         for if_property in properties:
+            # Skip if it's debug mode for profiling
+            if self.debug_mode:
+                if if_property == 'profiling':
```
Gustry: Use the layer purpose key from definition?

safe/impact_function/test/test_impact_function.py
```diff
@@ -682,6 +683,19 @@ def test_scenario(self, scenario_path=None, use_debug=True):
         # counted in the JSON file.
         self.assertEqual(len(outputs) - 1, expected_outputs['count'])
         self.assertEqual(len(outputs), computed_nb_outputs)
+
+        # Test deserialization
+        if test_loader:
+            LOGGER.debug('Test deserialization. Lontong')
```
Gustry: Why not "lemper ayam"?

safe/impact_function/test/test_impact_function.py
```diff
@@ -682,6 +683,19 @@ def test_scenario(self, scenario_path=None, use_debug=True):
...
+            output_metadata = impact_function.impact.keywords
+            new_impact_function = ImpactFunction. \
+                load_from_output_metadata(output_metadata)
+            try:
+                self.assertEquals(impact_function, new_impact_function)
+            except AssertionError as e:
+                raise AssertionError(
```
Gustry: Exceptions are very important if we know where they come from. By doing this, we lose the orignal traceback and we create a new one. Then we can't know where it has been raised before. Look for `Another case when people often clobber the traceback is when they want to add information to it, e.g.:` in http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html

Or maybe you can do:
```
except AssertionError as e:
     LOGGER.debug("Exception found in" +  scenario_path)
     raise
```
 +                raise AssertionError(

safe/impact_function/test/test_impact_function.py
```diff
@@ -713,10 +727,12 @@ def test_scenarios(self):
         }
 
         # If we want to invert the selection.
-        invert = False
+        invert = True
```
Gustry: Can we keep it to False?

safe/impact_function/impact_function.py
Gustry: Hum, I think it's still a function, but I guess PyCharm is confused because the same `callback` function has the `@property` decorator. And it's true a property can't be called. But in this case, the property is a function.

I understand what you mean, I got the same issue, but I let it as I didn't know how to fix it.
Do what you want ;-)
IMHO, we should fix this `property` which isn't a real property but a function maybe. But I'm lazzy to do that, very low priority.

safe/impact_function/test/test_impact_function.py
Gustry: I think I was tired ...

safe/impact_function/impact_function.py
ismailsunni: Hmm, actually it's not a layer purpose key, but IF's attribute. So, I think I will keep using it.

safe/impact_function/test/test_impact_function.py
ismailsunni: Ahaha, you got me :D

safe/impact_function/test/test_impact_function.py
ismailsunni: Ouch, I forget it again. Thanks

safe/impact_function/test/test_impact_function.py
ismailsunni: Hmm, my bad. I just want to add the scenario file. Ok, I will  just use your suggestion. It seems I need to read the article in clear mind. Thanks.

ismailsunni:
Yeay, green, I am merging this.

PR #4531, opened 2017-10-19
improve the make command to run tests with docker

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: improve the make command to run tests with docker

* One make command
* No more generated files
* Time at the end

### Checklist:

### DIFF
--- a/safe/gui/tools/test/test_multi_buffer.py
+++ b/safe/gui/tools/test/test_multi_buffer.py
@@ -22,11 +22,6 @@ class MultiBufferTest(unittest.TestCase):
         """Test initialisation run before each test."""
         self.output_path = standard_data_path(
             'hazard', 'volcano_point_multi_buffer.geojson')
-        pass
-
-    def tearDown(self):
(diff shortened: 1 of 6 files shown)


safe/gui/tools/test/test_multi_buffer.py
```diff
@@ -77,6 +72,14 @@ def multi_buffer_test(self, output_path):
 
         self.assertEqual(expected_fields_name, new_field_names)
 
+        # We need to clean generated files
+        try:
+            os.remove(self.output_path)
+            os.remove(self.output_path.replace('.geojson', '.xml'))
```
ismailsunni: if the first one raise exception, the second one wouldn't be run. Is that ok?

Gustry: yes it's ok, because the XML is linked to the geojson ;-)

Gustry:
Yes I think so, but very low priority for me too

PR #4539, opened 2017-10-20
pycharm cleaning imports and script to rollback

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: The script I started yesterday

I already run it today @ismailsunni. It's sorting alphabetically your imports from this morning ;-)

We can use PyCharm from command line to run these process sometimes: https://www.jetbrains.com/help/pycharm/command-line-code-inspector.html

### Checklist:

### DIFF
--- a/safe/common/utilities.py
+++ b/safe/common/utilities.py
@@ -25,7 +25,6 @@ from safe.common.exceptions import VerificationError
 from safe.utilities.rounding import (
     thousand_separator, decimal_separator, add_separators)
 
-
 LOGGER = logging.getLogger('InaSAFE')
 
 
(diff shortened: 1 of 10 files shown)


ismailsunni:
Sorry @Gustry 
Why do we need to roll back?

Gustry:
Because I don't want PyCharm to touch these files:
* I don't care about tests files for now. We have some rules like 'get_qgis_app` on top before imports
* I don't want __init__ because it might break for now the order of import, I will check later
* I don't want safe/plugin.py because import qgis needs to be first for the SIP api version 2 and PyCharm is removing the logger.

ismailsunni:
ah I understand now, thanks :)

Issue #4482, opened 2017-09-28
InaSAFE Profile


ismailsunni:
### Problem
Create a tool to support editing the displaced and affected people for each hazard classification. This should be stored somewhere in the `.qgis2/python/plugins/inasafe` directory, so that we can share it to another people.

### Expected Behaviour
User can edit the displaced and affected ratio people.

### Proposed Solution
- [x] Create the UI
  - [x] With logo, organisation name, contact in the first tab) --> save and load
  - [x] Second tab is the setting affected and displaced fields
  - [x] Run it on the first running InaSAFE
- [x]  Make it update the hazard classification definition value

cc @timlinux

ismailsunni:
Another good thing to add here is currency. Currently, we put it in InaSAFE setting, it would be good to put it in the InaSAFE profile.

ismailsunni:
Hi @timlinux @Charlotte-Morgan 
I have made a simple mock up for the UI. Do you have any comment about it?

<img width="656" alt="inasafeprofile" src="https://user-images.githubusercontent.com/1421861/31212784-c41976b0-a9cc-11e7-8ee9-6cd0618e3ba9.png">

Charlotte-Morgan:
hi @ismailsunni thanks for the mock up. 
The organisation tab looks fine. 
* Will currency be a drop down?
* Might you want to add units too?

A few suggestions for the profile tab: - 
* I think it needs a different name but not sure what; maybe preferences / settings / ... / customisation
*Maybe  "Please fill in the custom value for your organisation profile"  could say "Please customise the population exposure preferences for your organisation profile:
* only relevant classifications should show for each hazard :) and only those that apply to population exposure
* swap affected to the left and displaced to the right
* please use the word rate not ratio 
* affected should be binary True or False
* displaced should only be editable if affected = true
* profile preview => customised profile

I think it would be good to have definitions for concepts; especially affected and displaced and how these are used by InaSAFE (analysis & reports)

ismailsunni:
Hi @Charlotte-Morgan thanks for the review
> Will currency be a drop down?

I was thinking to make it drop down, but the user also can set it manually (i.e. write in the dropdown). I think there is a widget for it. But for the first step, I will make it as drop down like in the current InaSAFE Setting.

> Might you want to add units too?

What units do you mean? I am not sure we can add unit since every layers use different unit. Or perhaps you mean another unit?

A few suggestions for the profile tab: -

[...]

PR #4543, opened 2017-10-24
Update Option Tabs

ismailsunni:
### What does it fix?
* Ticket: #4482 
* Funded by: DFAT
* Description: 
   - First step to have InaSAFE profile
   - Reorder tab in the InaSAFE options and make it prettier

update_logo

### Checklist:

### DIFF
--- a/safe/gui/tools/options_dialog.py
+++ b/safe/gui/tools/options_dialog.py
@@ -2,14 +2,15 @@
 """InaSAFE Options Dialog"""
 
 import logging
+import os
 
 # This import is to enable SIP API V2
 # noinspection PyUnresolvedReferences
 import qgis  # NOQA pylint: disable=unused-import
-from PyQt4.QtCore import pyqtSignature, pyqtSlot, QVariant, QSettings
+from PyQt4.QtCore import pyqtSignature, pyqtSlot, QVariant, QSettings, Qt
 from PyQt4.QtGui import (
     QDialog, QFileDialog, QDialogButtonBox, QGroupBox, QVBoxLayout,
-    QScrollArea, QWidget)
+    QScrollArea, QWidget, QPixmap,)
 from parameters.float_parameter import FloatParameter
 from parameters.integer_parameter import IntegerParameter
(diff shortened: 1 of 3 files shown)


safe/gui/tools/options_dialog.py
```diff
@@ -160,6 +162,7 @@ def __init__(self, iface, parent=None, qsetting=''):
         # hide custom template dir toggle
         self.custom_templates_dir_checkbox.hide()
         self.splitter_custom_report.hide()
+        # self.set_organisation_logo()
```
Gustry: You keep it for legacy or is-it really old code?

safe/gui/tools/options_dialog.py
```diff
@@ -412,7 +418,7 @@ def on_toolNorthArrowPath_clicked(self):
 
     # noinspection PyPep8Naming
     @pyqtSignature('')  # prevents actions being handled twice
```
Gustry: Can'we remove these decorator too? I guess if we use explicit connect, it's fine, no? It's only a question.

safe/gui/tools/options_dialog.py
```diff
@@ -412,7 +418,7 @@ def on_toolNorthArrowPath_clicked(self):
...
-    def on_toolOrganisationLogoPath_clicked(self):
+    def on_open_organisation_logo_path_button_clicked(self):
         """Auto-connect slot activated when logo file tool button is clicked.
```
Gustry: Can we remove these autoconnects while you are working on this file?

safe/gui/tools/options_dialog.py
ismailsunni: Just forget to delete :)

safe/gui/tools/options_dialog.py
ismailsunni: I think we can. I never use this decorator actually.

safe/gui/tools/options_dialog.py
ismailsunni: Ok, I will. Thanks for the review.

PR #4546, opened 2017-10-24
remove some legacy code

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: I got a QGIS crash while I was doing some testing on my other branch, maybe because of my new dev environment.

So I was looking a little bit in the test package. @timlinux what would block us to use qgis.utils.iface instead of the fake InaSAFE one?

Happy to read comments

### Checklist:


timlinux:
That looks fine for me @Gustry - can you just check that the QGIS Supplied one has implementations for message bar and the few other methods that had a line or two of logic in them. But basically if the tests can pass in travis I am happy to remove this stuff...

Gustry:
* Travis is using docker, so it's not using these files anyway.
* The iface that we got in PyCharm using the 'get_qgis_app' will come from qgis.utils.iface. I remember during my first month in the internship, you didn't agree to use this qgis.utils.iface because we couldn't do some testing somehow (you got an error about the messageBar or something). What should I check?

Gustry:
Ok, let's try and we can still revert if needed

PR #4548, opened 2017-10-26
use our internal settings class instead of QSettings

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: One step forward to QGIS 3 as we will have to switch to QgsSettings. So we can do this change in our class to switch them all ;-)

@ismailsunni There is still a few. And some are using the QSettings for testing. I will let them or you can help. I was doing this as a break

### Checklist:


Gustry:
Don't worry, the list is not finished. You can still do some ;-)

PR #4551, opened 2017-10-26
switch from pep8 to pycodestyle

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: We got a warning since a few days saying that pep8 is going to be removed and need to be replaced by pycodestyle.

I saw that already one year ago. Time to switch

### Checklist:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,4 +4,4 @@ max-line-length = 79
 
 # E12x continuation line indentation
 
-ignore = E121
+ignore = E121,E722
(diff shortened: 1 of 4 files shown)


.flake8
```diff
@@ -4,4 +4,4 @@ max-line-length = 79
 
 # E12x continuation line indentation
 
-ignore = E121
+ignore = E121,E722
```
ismailsunni: Can you put what's E722 about in the comment? So that we don't need to look up in the internet.

Gustry: I plan to fix it in the next few days :-) So this is only temporary

PR #4558, opened 2017-10-27
trying to remove "message" from exception

Gustry:
https://stackoverflow.com/questions/1272138/baseexception-message-deprecated-in-python-2-6

### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: A little step forward to Python 3 maybe

While I'm debugging, I have some python warnings.

We still have some `exception.message` where we write on our `Message` object. I guess it's not so good but I'm not sure how to remove these.

it seems the message is already included in the constructor.

I'm not sure about this.

### Checklist:

### DIFF
--- a/safe/common/exceptions.py
+++ b/safe/common/exceptions.py
@@ -31,36 +31,28 @@ class InaSAFEError(RuntimeError):
     suggestion = 'An unspecified error occurred.'
 
     def __init__(self, message=None):
-        """"General constructor.
+        """General constructor.
 
         :param message: The optional error message.
         :type message: str, unicode, MessageElement
-        """""
+        """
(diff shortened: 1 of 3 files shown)


Gustry:
travis is happy. The check is about `unicode`, even it's available on python 2

ismailsunni:
> While I'm debugging, I have some python warnings.
What's the warning about?

ismailsunni:
Btw, can we disable that sideci check for unicode?

Gustry:
The same as in the stackoverflow link.

Ok Let's merge to give it a try.

About unicode, I'm not sure, I already set python 2 I think.

PR #4562, opened 2017-10-28
Welcome message

ismailsunni:
### What does it fix?
* Ticket: Part of #4482 
* Funded by: DFAT
* Description: 
   - Adding welcome message when you run InaSAFE for the first time.
   - It will check the setting for `inasafe/first_run`. If it's True, or not exist, it will show the welcome message. Currently, you can only get the message once. If you want to see it again, set `inasafe/first_run` setting in QGIS, and set it to True.

welcome_message

Of course I need you @timlinux to make a better message :D  

### Checklist:

### DIFF
--- a/safe/gui/tools/help/welcome_message.py
+++ b/safe/gui/tools/help/welcome_message.py
@@ -1,0 +1,78 @@
+# coding=utf-8
+"""Welcome text for InaSAFE."""
+
+import logging
+
+from safe import messaging as m
+from safe.messaging import styles
+from safe.utilities.i18n import tr
+
+LOGGER = logging.getLogger('InaSAFE')
+# For chapter sections
+# Items marked as numbered below will show section numbering in HTML render
+TITLE_STYLE = styles.TITLE_LEVEL_1_STYLE  # h1 Not numbered
+SECTION_STYLE = styles.SECTION_LEVEL_2_STYLE  # h2 numbered
+SUBSECTION_STYLE = styles.SUBSECTION_LEVEL_3_STYLE  # h3 numbered
+BLUE_CHAPTER_STYLE = styles.BLUE_LEVEL_4_STYLE  # h4 numbered
+RED_CHAPTER_STYLE = styles.RED_LEVEL_4_STYLE  # h4 numbered
+DETAILS_STYLE = styles.ORANGE_LEVEL_5_STYLE  # h5 numbered
+DETAILS_SUBGROUP_STYLE = styles.GREY_LEVEL_6_STYLE  # h5 numbered
+# For images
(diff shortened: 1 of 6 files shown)


safe/gui/tools/help/welcome_message.py
```diff
@@ -0,0 +1,78 @@
...
+    return message
+
+
+def heading():
+    """Helper method that returns just the header.
+
+    This method was added so that the text could be reused in the
+    other contexts.
+
+    .. versionadded:: 4.1.0
```
Gustry: 4.3 ?

safe/gui/tools/help/welcome_message.py
```diff
@@ -0,0 +1,78 @@
...
+    return message
+
+
+def content():
+    """Helper method that returns just the content.
+
+    This method was added so that the text could be reused in the
+    dock_help module.
+
+    .. versionadded:: 4.1.0
```
Gustry: 4.3 ?

safe/gui/tools/help/welcome_message.py
ismailsunni: Ah yes, forget to update it. Thanks.

safe/gui/tools/help/welcome_message.py
ismailsunni: Same as above.

ismailsunni:
After checking again, I am still not sure if my code is right. There are some case:
1. Show it after the plugin loaded / installed (call the function in initGui method)
2. Show after the QGIS is loaded (make a slot for `initializationCompleted` completed)

But, the first one will make the welcome message shown before the initializationCompleted (if we want to show it every time we open in QGIS).

Gustry:
What is wrong with the initGUI ? Is-it not enough?

But please, we mustn't be too intrusive on UX. I hate plugins doing too much display or popup on launch time if I don't explicitly use them or call them

ismailsunni:
A little update from the UI. Adding check box to always show it.
screen shot 2017-10-30 at 21 02 47
screen shot 2017-10-30 at 21 03 19

ismailsunni:
> What is wrong with the initGUI ? Is-it not enough?

initGUI is finished before QGIS initialization is completed. So, it will be shown before QGIS (which is not good, I think). But all is good now, since we decide to show it only after open QGIS (not after load InaSAFE)

> But please, we mustn't be too intrusive on UX. I hate plugins doing too much display or popup on launch time if I don't explicitly use them or call them

Yeah, me too. We only show this if:
1. The user set always show the welcome message check box
2. If the user use newer version of InaSAFE (compared the previous one). And this only once per new version installed.

PR #4565, opened 2017-10-30
Use is_affected for affected post processor.

ismailsunni:
### What does it fix?
* Ticket: #4482 
* Funded by: DFAT
* Description: 
   - Use is_affected function for the affected post processor so that it will use the value from InaSAFE profile.

### Checklist:

### DIFF
--- a/safe/report/extractors/action_notes.py
+++ b/safe/report/extractors/action_notes.py
@@ -14,7 +14,7 @@ from safe.utilities.resources import (
     resource_url,
     resources_path)
 from safe.utilities.rounding import html_scientific_notation_rate
-from safe.definitions.utilities import get_displacement_rate
+from safe.definitions.utilities import get_displacement_rate, is_affected
 
 __copyright__ = "Copyright 2016, The InaSAFE Project"
 __license__ = "GPL version 3"
@@ -86,8 +86,18 @@ def notes_assumptions_extractor(impact_report, component_metadata):
     # Check hazard affected class
     affected_classes = []
     for hazard_class in hazard_classification['classes']:
-        if hazard_class.get('affected', False):
-            affected_classes.append(hazard_class)
+        if exposure_keywords['exposure'] == exposure_population['key']:
+            # Taking from profile
+            is_affected_class = is_affected(
+                hazard=hazard_keywords['hazard'],
+                classification=hazard_classification['key'],
+                hazard_class=hazard_class['key'],
+            )
+            if is_affected_class:
+                affected_classes.append(hazard_class)
+        else:
+            if hazard_class.get('affected', False):
+                affected_classes.append(hazard_class)
 
     if affected_classes:
         affected_note_dict = resolve_from_dictionary(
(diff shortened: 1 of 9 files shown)


Gustry:
Some tests layers have been generated before the 4.0 freeze. So these layers are missing hazard_keywords. Feel free to monkey patch them in the test or add the missing keywords in the XML

ismailsunni:
Thanks for the hint @Gustry

safe/report/test/test_impact_report.py
```diff
@@ -102,13 +102,15 @@ def setUp(self):
         # Change displacement rate so the result is easily distinguished
         self.default_displacement_rate = flood_hazard_classes['classes'][0][
             'displacement_rate']
-        flood_hazard_classes['classes'][0][
-            'displacement_rate'] = self.custom_displacement_rate
+        # flood_hazard_classes['classes'][0][
+        #     'displacement_rate'] = self.custom_displacement_rate
```
Gustry: Any reason to keep oldcode? It's in a test file, it's ok, but I usually remove code which will be dead forever

safe/report/test/test_impact_report.py
```diff
@@ -1063,6 +1065,152 @@ def test_minimum_needs_outputs(self):
 
         shutil.rmtree(output_folder, ignore_errors=True)
 
+    def test_minimum_needs_outputs_modified(self):
+        """Test generate minimum needs section with updated profile
```
Gustry: The regex is not able to read multiline comments to add the full stop. but it's test file, I don't mind ;-)

safe/report/test/test_impact_report.py
```diff
@@ -1063,6 +1065,152 @@ def test_minimum_needs_outputs(self):
...
+        # Change affected
+        profile = setting(key='profile')
+        profile[hazard_flood['key']][flood_hazard_classes['key']][
+            'wet']['affected'] = False
+        profile[hazard_flood['key']][flood_hazard_classes['key']][
+            'dry']['affected'] = True
+        profile[hazard_flood['key']][flood_hazard_classes['key']][
+            'dry']['displacement_rate'] = 0.5
+        set_setting(key='profile', value=profile)
+        is_affected_wet = is_affected(
```
Gustry: I took me more seconds to read, `wet_is_affected` looks better. Same for the one. But I also don't care, it's a test file. For next time

safe/report/extractors/action_notes.py
```diff
@@ -255,7 +255,6 @@ def action_checklist_report_pdf_extractor(impact_report, component_metadata):
     # - Substitution maps
     # - Element settings, such as icon for picture file or image source
 
-    _ = component_metadata
```
Gustry: If you still need that, to mark that the var is not used you need:
`_ = component_metadata  # NOQA` and then everything is fine. I agree it could be nice automatically a `_` variable.

ismailsunni: Yeah. I will let it since I just pick it up when I open this file.

I forget to add NOQA previously :)

Thanks

Gustry: yep, because pycharm will now complain too ;-)

PR #4573, opened 2017-10-31
Export import setting

ismailsunni:
### What does it fix?
* Ticket: #4482 
* Funded by: DFAT
* Description: 
   - Add export and import functionality in the options

### Checklist:


Gustry:
Just thinking is some settings are needed in the file. Like the last extent used by the IF, prefered extent selector without aggregation, ... but should we worry about these settings?

ismailsunni:
Hmm, not sure also. There are many things that probably not works like reporting path, north arrow logo, and so on. But @timlinux said to make it work first with all settings. We can address that later if needed.

PR #4574, opened 2017-10-31
 add first implementation preprocessors API

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: This PR brings the API to have some pre-processors in InaSAFE.

The API is limited to our use cases:
* produce a side product about nearby places
* produce contours for realtime

But I think we can easily improve it add more use cases if needed.

These postprocessors have not been implemented in this PR! Only the API.
I think I will merge like this, even with these temporary `fake layer strings`. Only if on the release day we don't have any real preprocessor implemented, I will comment these lines and disable tests related to preprocessors. So @ivanbusthomi can just come and update the existing code. Is-it ok?

Preprocessors and Postprocessors are quite different.
* Postprocessors are editing the single impact layer only by adding new fields based on a formula. The postprocessors works feature per feature.
* Preprocessors are creating new layers (with a unique layer_purpose to find them in the datastore)
* Because preprocessors don't work with a single layer like in postprocessors, a preprocessors works with an impact function object. It makes the creation of the condition very easy to write (contrary to the postprocessor API). 

### Checklist:


safe/processors/pre_processors.py
```diff
@@ -0,0 +1,81 @@
...
+    :rtype: QgsMapLayer
+    """
+    _ = impact_function  # NOQA
+    from safe.test.utilities import load_test_vector_layer
+    fake_layer = load_test_vector_layer(
+        'gisv4',
+        'impacts',
+        'building-points-classified-vector.geojson',
+        clone_to_memory=True)
+    return fake_layer
```
Gustry: From line 57 to 64, these lines are considered temporary.
Basically, it should be something like:
```
exposure = impact_function.exposure
hazard = impact_function.hazard
new_layer = awesome_function_to_create_side_product(exposure, hazard)
return new_layer
```
@ivanbusthomi @ismailsunni, do you think it can be OK?

ismailsunni: Yeah, I think so.

Issue #4495, opened 2017-10-04
Reporting Option


ismailsunni:
### Problem
From the last day discussion at October 2017 Site Visit, we decide to create a dialog to change / adjust  how the reporting will look like. Some of the features:
- [ ] Ordering/Show-Hide layers in the map from the current loaded layers, result layers.
- [ ] Choosing which template that the user want to use.
- [ ] Put some standard options to set the layers

This dialog will be put in the dock (like the old impact function option button with 3 dots.)

cc @timlinux @Charlotte-Morgan @Gustry @myarjunar @samnawi @felix-yew @ivanbusthomi 

PS: Let me know or just update the ticket if I missed something.

Gustry:
Looks good.
I'm working on how to provide output layer names from the IF before the process.

Gustry:
First PR here: https://github.com/inasafe/inasafe/pull/4488

Then my part should be more or less done. @myarjunar will you take these options in the reporting engine?

Charlotte-Morgan:
@ismailsunni will this also address #4477 ?

ismailsunni:
@Charlotte-Morgan 
I think yes, since the user can specifiy which layers that they want to see (and the order of the layer), it will solve the issue in #4477. Bug I agree, we need to put and option to use a standard option.

ismailsunni:
Based on last discussion, we agree to use layer panel to set the layer order and visibility. So, we don't need this reporting option dialog. I am closing this. cc @Gustry @timlinux @myarjunar

Gustry:
Don't we need it for the multiexposure? Or did I forgot something already?

ismailsunni:
Can we do it in the layer canvas also for multi exposure?

timlinux:
Just a note that

>  Ordering/Show-Hide layers in the map from the current loaded layers, result layers.

is only for multi exposure workflow

timlinux:
> Choosing which template that the user want to use.

This should be for dock and multi-exposure etc.

Gustry:
All implemented according to PR above

PR #4579, opened 2017-11-02
Reporting options for the multi exposure

Gustry:
### What does it fix?
* Ticket: #4495 for the multiexposure
* Funded by: DFAT
* Description: Add a new panel in the multiexposure dialog to be able to set layer order in the report

screen shot 2017-11-02 at 20 21 42

Not the best UI, comments are welcome

@ismailsunni A review would be more than welcome ;-)

### Checklist:

### DIFF
--- a/safe/definitions/font.py
+++ b/safe/definitions/font.py
@@ -1,7 +1,12 @@
 # coding=utf-8
 """Fonts which are used in InaSAFE."""
 
-from PyQt4 import QtGui
+from PyQt4.QtGui import QFont
 
-big_font = QtGui.QFont()
+big_font = QFont()
 big_font.setPointSize(80)
+
+bold_font = QFont()
+bold_font.setItalic(True)
+bold_font.setBold(True)
+bold_font.setWeight(75)
--- a/safe/gui/analysis_utilities.py
+++ b/safe/gui/analysis_utilities.py
@@ -252,11 +252,12 @@ def add_layers_to_canvas_with_custom_orders(order, impact_function):
     :param impact_function: The multi exposure impact function used.
     :type impact_function: MultiExposureImpactFunction
     """
-    # Need to remove all groups  and layers
+    # Need to remove all groups and layers
     QgsMapLayerRegistry.instance().removeAllMapLayers()
     root = QgsProject.instance().layerTreeRoot()
     root.removeAllChildren()
 
+    # Insert layers in the good order.
     for layer in order:
         if layer[0] == FROM_CANVAS['key']:
             style = QDomDocument()
--- a/safe/gui/analysis_utilities.py
+++ b/safe/gui/analysis_utilities.py
@@ -7,12 +7,15 @@ from collections import OrderedDict
 from copy import deepcopy
 
 from PyQt4.QtCore import QDir, Qt
+from PyQt4.QtXml import QDomDocument
 from qgis.core import (
+    QGis,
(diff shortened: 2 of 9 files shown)


safe/definitions/font.py
```diff
@@ -1,7 +1,12 @@
...
 """Fonts which are used in InaSAFE."""
 
-from PyQt4 import QtGui
+from PyQt4.QtGui import QFont
 
-big_font = QtGui.QFont()
+big_font = QFont()
 big_font.setPointSize(80)
+
+bold_font = QFont()
```
ismailsunni: Ah nice, I should use it also in the profile widget

safe/gui/analysis_utilities.py
```diff
@@ -229,6 +235,44 @@ def add_debug_layers_to_canvas(impact_function):
...
+            ...
+        ]
+
+    :param order: Special structure the list of layers to add.
+    :type order: list
+
+    :param impact_function: The multi exposure impact function used.
+    :type impact_function: MultiExposureImpactFunction
+    """
+    # Need to remove all groups  and layers
```
ismailsunni: Two space before `and` ? :)

safe/gui/tools/multi_exposure_dialog.py
```diff
@@ -87,6 +108,219 @@ def __init__(self, parent=None):
...
+        canvas_branch = QTreeWidgetItem(
+            self.tree.invisibleRootItem(), [FROM_CANVAS['name']])
+        canvas_branch.setFont(0, bold_font)
+        canvas_branch.setExpanded(True)
+        canvas_branch.setFlags(Qt.ItemIsEnabled)
+
+        # List layers from the canvas
+        loaded_layers = QgsMapLayerRegistry.instance().mapLayers().values()
+        canvas_layers = iface.mapCanvas().layers()
+        flag = setting('visibleLayersOnlyFlag', True, bool)
```
ismailsunni: You don't need to set the default value, since it's already set in the `default_settings.py`. Anyway, not mandatory, but changing in one place is better.

safe/gui/tools/multi_exposure_dialog.py
```diff
@@ -215,6 +442,54 @@ def progress_callback(self, current_value, maximum_value, message=None):
         self.progress_bar.setValue(current_value)
         QApplication.processEvents()
 
+    def validate_impact_function(self):
+        """Check validity of the current impact function."""
+        # Always set it to False
+        self.button_box.button(QDialogButtonBox.Ok).setEnabled(False)
+
+        use_selected_only = setting('useSelectedFeaturesOnly', True, bool)
```
ismailsunni: Same as previous comment

safe/gui/tools/multi_exposure_dialog.py
Gustry: Does it means it's checking that the default value is set automatically if it doesn't exist in the QSettings?

Should we add a lookup in default_settings.py when we use the `settings` function too?

ismailsunni: > Does it means it's checking that the default value is set automatically if it doesn't exist in the QSettings?

Yes

> Should we add a lookup in default_settings.py when we use the settings function too?

Ideally, yes. I have a wish list to make each setting has more verbose description/definition. So that we can understand, what is what in the setting.

Issue #4440, opened 2017-09-19
[Multi Exposure] Integrate Report and Infographic for Exposures to be One


ismailsunni:
### Problem

After finishing #4439 we want to have one report / infographic for the whole analys (more than one exposures). We need to discuss how the report should be first of course.

cc @timlinux @myarjunar @Gustry

Gustry:
@myarjunar You can start having a look. The analysis layer is ready. The multi exposure is still like a dockyard, so feel free to update to make it work for you

timlinux:
For layer ordering we need to establish how the layer ordering will be passed to the api

![multiexposurelayerordering](https://user-images.githubusercontent.com/178003/32265806-50bcebc6-beed-11e7-88ae-b69cec8ad993.png)

Gustry:
Linked to ticket https://github.com/inasafe/inasafe/issues/4495

timlinux:
This has been implemented as part of the new multi-exposure API here  https://github.com/inasafe/inasafe/blob/develop/safe/impact_function/multi_exposure_wrapper.py#L414

See also #4495

Note that we only support Infographics for population single exposure and currently do not have plans to create generic QGIS composer Templates for multi-exposure.

PR #4582, opened 2017-11-05
Reporting for multi exposure

myarjunar:
### What does it fix?
* Ticket: #4440 
* Funded by: DFAT
* Description: 

a snippet of the general report
selection_146

### Checklist:

### DIFF
--- a/safe/definitions/provenance.py
+++ b/safe/definitions/provenance.py
@@ -108,6 +108,11 @@ provenance_impact_function_name = {
     'name': tr('Impact Function Name'),
     'provenance_key': 'impact_function_name'
 }
+provenance_multi_exposure_impact_function_name = {
+    'key': 'provenance_multi_exposure_impact_function_name',
+    'name': tr('Multi Exposure Impact Function Name'),
+    'provenance_key': 'multi_exposure_impact_function_name'
+}
 provenance_impact_function_title = {
     'key': 'provenance_impact_function_title',
     'name': tr('Impact Function Title'),
@@ -249,6 +254,8 @@ provenance_layer_profiling_id = {
     'provenance_key': layer_purpose_profiling['key'] + '_id'
 }
 
+
+# Mapping to global variable in QGIS
 provenance_list = [
     provenance_action_checklist,
     provenance_aggregation_keywords,
@@ -269,6 +276,7 @@ provenance_list = [
     provenance_hazard_layer_id,
     provenance_host_name,
     provenance_impact_function_name,
+    provenance_multi_exposure_impact_function_name,
     provenance_impact_function_title,
     provenance_inasafe_version,
     provenance_map_legend_title,
@@ -296,8 +304,6 @@ provenance_list = [
     provenance_layer_exposure_summary_table_id,
     provenance_layer_profiling_id,
 ]
-
-# Mapping to global variable in QGIS
 duplicated_global_variables = {
     provenance_os['provenance_key']: 'qgis_os_name'
 }
--- a/safe/definitions/provenance.py
+++ b/safe/definitions/provenance.py
@@ -254,8 +254,6 @@ provenance_layer_profiling_id = {
     'provenance_key': layer_purpose_profiling['key'] + '_id'
 }
 
-
-# Mapping to global variable in QGIS
 provenance_list = [
     provenance_action_checklist,
     provenance_aggregation_keywords,
@@ -304,6 +302,8 @@ provenance_list = [
     provenance_layer_exposure_summary_table_id,
     provenance_layer_profiling_id,
 ]
+
+# Mapping to global variable in QGIS
 duplicated_global_variables = {
     provenance_os['provenance_key']: 'qgis_os_name'
 }
--- a/safe/definitions/provenance.py
+++ b/safe/definitions/provenance.py
@@ -108,11 +108,6 @@ provenance_impact_function_name = {
     'name': tr('Impact Function Name'),
     'provenance_key': 'impact_function_name'
 }
-provenance_multi_exposure_impact_function_name = {
-    'key': 'provenance_multi_exposure_impact_function_name',
-    'name': tr('Multi Exposure Impact Function Name'),
-    'provenance_key': 'multi_exposure_impact_function_name'
-}
 provenance_impact_function_title = {
     'key': 'provenance_impact_function_title',
     'name': tr('Impact Function Title'),
@@ -274,7 +269,6 @@ provenance_list = [
     provenance_hazard_layer_id,
     provenance_host_name,
     provenance_impact_function_name,
-    provenance_multi_exposure_impact_function_name,
     provenance_impact_function_title,
     provenance_inasafe_version,
     provenance_map_legend_title,
--- a/safe/gis/vector/summary_5_multi_exposure.py
+++ b/safe/gis/vector/summary_5_multi_exposure.py
@@ -158,7 +158,14 @@ def multi_exposure_analysis_summary(analysis, intermediate_analysis):
     analysis.commitChanges()
     analysis.keywords['title'] = layer_purpose_analysis_impacted['name']
     analysis.keywords['layer_purpose'] = layer_purpose_analysis_impacted['key']
-    analysis.keywords['exposure_keywords'] = MULTI_EXPOSURE_ANALYSIS_FLAG
+
+    # Set up the extra keywords so everyone knows it's a
+    # multi exposure analysis result.
+    extra_keywords = {
+        'analysis_type': MULTI_EXPOSURE_ANALYSIS_FLAG
+    }
+    analysis.keywords['extra_keywords'] = extra_keywords
+
     if qgis_version() >= 21600:
         analysis.setName(analysis.keywords['title'])
     else:
@@ -233,7 +240,14 @@ def multi_exposure_aggregation_summary(aggregation, intermediate_layers):
     aggregation.keywords['title'] = layer_purpose_aggregation_summary['name']
     aggregation.keywords['layer_purpose'] = (
         layer_purpose_aggregation_summary['key'])
-    aggregation.keywords['exposure_keywords'] = MULTI_EXPOSURE_ANALYSIS_FLAG
+
+    # Set up the extra keywords so everyone knows it's a
+    # multi exposure analysis result.
+    extra_keywords = {
+        'analysis_type': MULTI_EXPOSURE_ANALYSIS_FLAG
+    }
+    aggregation.keywords['extra_keywords'] = extra_keywords
+
     if qgis_version() >= 21600:
         aggregation.setName(aggregation.keywords['title'])
     else:
--- a/safe/gis/vector/summary_5_multi_exposure.py
+++ b/safe/gis/vector/summary_5_multi_exposure.py
@@ -4,6 +4,7 @@
 
 from qgis.core import QgsFeatureRequest
 
+from safe.definitions.constants import MULTI_EXPOSURE_ANALYSIS_FLAG
 from safe.definitions.fields import (
     affected_exposure_count_field,
     exposure_affected_exposure_type_count_field,
@@ -157,6 +158,7 @@ def multi_exposure_analysis_summary(analysis, intermediate_analysis):
     analysis.commitChanges()
     analysis.keywords['title'] = layer_purpose_analysis_impacted['name']
     analysis.keywords['layer_purpose'] = layer_purpose_analysis_impacted['key']
+    analysis.keywords['exposure_keywords'] = MULTI_EXPOSURE_ANALYSIS_FLAG
     if qgis_version() >= 21600:
         analysis.setName(analysis.keywords['title'])
     else:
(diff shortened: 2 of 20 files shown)


safe/impact_function/impact_function.py
```diff
@@ -252,6 +252,7 @@ def __init__(self):
         self._name = None  # e.g. Flood Raster on Building Polygon
         self._title = None  # be affected
         self._unique_name = None  # EXP + On + Haz + DDMMMMYYYY + HHhMM.SS.ms
+        self._multi_exposure_name = None
```
Gustry: What is the good reason to have this name here?
impact_function.py is not aware of the multiexposure. That's why we have the multi_exposure_wrapper to manage all impact_function individually.
I would prefer very much if we can remove this. If it's for the provenance, can't you write it in the wrapper? I really tried to not touch the impact_function.

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -443,3 +467,55 @@ def run(self):
             self._analysis_summary, analysis_width, analysis_color)
 
         return ANALYSIS_SUCCESS, None
+
+    def generate_report(self, components, output_folder=None, IFACE=None):
```
Gustry: A variable must be lowercase. See here where I asked your review: https://github.com/inasafe/inasafe/pull/4581

safe/report/expressions/html_report.py
```diff
@@ -0,0 +1,193 @@
...
+    if not project_context_scope.hasVariable(key):
+        return None
+
+    analysis_dir = dirname(project_context_scope.variable(key))
+    return get_impact_report_as_string(analysis_dir)
+
+
+@qgsfunction(
+    args='auto', group=group, usesGeometry=False, referencedColumns=[],
+    help_text=help_message.to_html(), helpText=help_message.to_html())
```
Gustry: You need to redefine help_message for each expression with custom help and examples.

safe/report/expressions/html_report.py
```diff
@@ -0,0 +1,193 @@
...
+    _ = feature, parent  # NOQA
+    analysis_dir = get_analysis_dir(exposure_land_cover['key'])
+    if analysis_dir:
+        return get_impact_report_as_string(analysis_dir)
+    return None
+
+
+def get_analysis_dir(exposure_key):
+    """Retrieve an output directory of an analysis/ImpactFunction from a
+    multi exposure analysis/ImpactFunction based on exposure type..
```
Gustry: Double full stop

safe/impact_function/impact_function.py
myarjunar: ah this is my old approach, for provenance. I was trying to find a way to identify whether an IF is a member of Multi-exposure IF, but I ended up using the group registry in the legend to identify that case. we can safely remove this, thanks anyway.

safe/report/expressions/html_report.py
myarjunar: ups I wasn't aware about that, I thought it was taken from the docstring. thanks

Gustry: Please read lines between 34 to 41 in the same file, everything is explained, you copy/pasted these comments ;-)

safe/definitions/provenance.py
```diff
@@ -249,6 +254,8 @@
     'provenance_key': layer_purpose_profiling['key'] + '_id'
 }
 
+
+# Mapping to global variable in QGIS
```
ismailsunni: Why do you move this?

safe/gui/tools/multi_exposure_dialog.py
```diff
@@ -245,6 +249,15 @@ def accept(self):
                 code, message = self._multi_exposure_if.run()
                 # self.assertEqual(code, ANALYSIS_MULTI_SUCCESS, message)
 
+                error_code, message = self._multi_exposure_if.generate_report(
+                    [standard_multi_exposure_impact_report_metadata_pdf])
+
+                if error_code == ImpactReport.REPORT_GENERATION_FAILED:
+                    LOGGER.info(tr(
```
ismailsunni: No need to use `tr` for logging. Less translation work

safe/gui/tools/multi_exposure_dialog.py
```diff
@@ -260,6 +273,16 @@ def accept(self):
                         0, analysis.name)
                     detailed_group.setVisible(Qt.Checked)
                     add_impact_layers_to_canvas(analysis, group=detailed_group)
+
+                    error_code, message = analysis.generate_report(
+                        all_default_report_components)
+
+                    if error_code == ImpactReport.REPORT_GENERATION_FAILED:
+                        LOGGER.info(tr(
```
ismailsunni: Same here

safe/gui/widgets/dock.py
```diff
@@ -864,17 +865,24 @@ def layer_changed(self, layer):
...
             show_keywords = True
             if keywords.get('layer_purpose') in impacted_layer:
-                report_path = os.path.dirname(layer.source())
+                report_dir_path = os.path.dirname(layer.source())
                 report_path = os.path.join(
-                    report_path, 'output/impact-report-output.html')
+                    report_dir_path, 'output/impact-report-output.html')
+                multi_exposure_report_path = os.path.join(
+                    report_dir_path,
+                    'output/multi-exposure-impact-report-output.html')
```
ismailsunni: Don't use `/`, vunerable on Windows

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -298,6 +321,21 @@ def prepare(self):
             if status != PREPARE_SUCCESS:
                 return status, message
 
+        hazard_name = get_name(self.hazard.keywords.get('hazard'))
+        hazard_geometry_name = get_name(geometry_type(self.hazard))
+        self._name = (
+            u'Multi exposure {hazard_type} {hazard_geometry} On '.format(
```
ismailsunni: Use title case for `exposure` ? Since `On` is also in title case (although, `on` is preposition). Do we need a translation here?

safe/definitions/provenance.py
myarjunar: ups must be my magic

safe/gis/vector/summary_5_multi_exposure.py
```diff
@@ -157,6 +158,7 @@ def multi_exposure_analysis_summary(analysis, intermediate_analysis):
     analysis.commitChanges()
     analysis.keywords['title'] = layer_purpose_analysis_impacted['name']
     analysis.keywords['layer_purpose'] = layer_purpose_analysis_impacted['key']
+    analysis.keywords['exposure_keywords'] = MULTI_EXPOSURE_ANALYSIS_FLAG
```
Gustry: Why do you set this flag here?
`exposure_keywords` must be keywords from the exposure layer.

safe/gis/vector/summary_5_multi_exposure.py
```diff
@@ -231,6 +233,7 @@ def multi_exposure_aggregation_summary(aggregation, intermediate_layers):
     aggregation.keywords['title'] = layer_purpose_aggregation_summary['name']
     aggregation.keywords['layer_purpose'] = (
         layer_purpose_aggregation_summary['key'])
+    aggregation.keywords['exposure_keywords'] = MULTI_EXPOSURE_ANALYSIS_FLAG
```
Gustry: Same question here

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -446,3 +470,55 @@ def run(self):
...
+        # iface set up, in case IF run from test
+        if not iface:
+            iface = IFACE
+
+        for component in components:
+
+            report_metadata = ReportMetadata(
+                metadata_dict=update_template_component(component))
+
+            self.impact_report = ImpactReport(
```
Gustry: PyCharm is warning you about this variable not in __init__. Add it there
Question, should it be private?

Gustry:
It's a very big PR. Hard to review carefully. it would be nice to do smaller PR next time ;-)

myarjunar:
yes I'm sorry for that but it surprised me as well, I didn't expect to be this big @Gustry

safe/gis/vector/summary_5_multi_exposure.py
myarjunar: I'm about to ask this thing. because it has multiple exposures, could we set the exposure keywords with list of exposure key or better to create new keywords which tell us whether that layer is coming from multi exposure or not?

safe/impact_function/multi_exposure_wrapper.py
myarjunar: yes it should be private

safe/gis/vector/summary_5_multi_exposure.py
Gustry: Yes I understand.
Hum, it would be better to have a dictionary, pseudocode:
```
{
    roads exposure key: roads keywords list,
    population exposure key: population keywords list,
}
```
No ?

myarjunar: so @ismailsunni suggest me to put that flag as an `extra_keywords`. I think I will go with that

Gustry: You will go with the flag you mean?
Sorry, I'm confused, where is the suggestion? Why `extra_keywords` ? So you drop `exposure_keywords` ?

myarjunar: If we go with flag in `extra_keywords`, `exposure_keywords` will be just empty as it is. I prefer with flag in `extra_keywords` because I think we will introduce new standard in filling the `exposure_keywords` if we use the `extra_keywords` because single IF already has theirs.

Gustry: exposure_keywords must stay the same across IF. So it's either not present or with keywords from exposures. So yes, don't put that flag here I think.

ismailsunni: > Sorry, I'm confused, where is the suggestion?

in private chat

Gustry: it would be nice to be on gitter ;-)

ismailsunni:
Another notes, I saw you added html_report.py file. Can you add unit test for it also? It will make our future easier to have unit test per file / part.

ismailsunni:
I just check develop branch, it seems we have more files for code than the test file for safe/reporting.
screen shot 2017-11-09 at 06 17 27

I am not sure that we have all code tested (or at least most of it) since I almost never touch that module (beside fixing unit test).

Anyway, having too much unit test is better than not enough one.

/me is missing the coverage

Gustry:
> Another notes, I saw you added html_report.py file. Can you add unit test for it also? It will make our future easier to have unit test per file / part.

Agree. Discussed with @myarjunar we need to setup the legend for this test, so not so easy.

> it seems we have more files for code than the test file for safe/reporting.

Some test are covering one or many files, so it also fine sometimes.

> /me is missing the coverage

Something like this from PyCharm ;-)

screen shot 2017-11-09 at 10 00 15

I got a lot failing tests so it's very representative.

safe/gis/vector/summary_5_multi_exposure.py
```diff
@@ -157,6 +158,14 @@ def multi_exposure_analysis_summary(analysis, intermediate_analysis):
     analysis.commitChanges()
     analysis.keywords['title'] = layer_purpose_analysis_impacted['name']
     analysis.keywords['layer_purpose'] = layer_purpose_analysis_impacted['key']
+
+    # Set up the extra keywords so everyone knows it's a
+    # multi exposure analysis result.
+    extra_keywords = {
+        'analysis_type': MULTI_EXPOSURE_ANALYSIS_FLAG
```
Gustry: Can you create the entry for `analysis_type` in the definition instead and use it where you need to read this entry.

myarjunar:
> Another notes, I saw you added html_report.py file. Can you add unit test for it also? It will make our future easier to have unit test per file / part.

@ismailsunni @Gustry the test will be coming in this PR #4610 for the simple one first, and later after I figure out how to setup all aspects that I need would be coming in another PR, I hope its ok.

safe/report/impact_report.py
```diff
@@ -440,6 +443,14 @@ def impact_function(self):
         """
         return self._impact_function
 
+    @property
+    def multi_exposure_impact_function(self):
```
Gustry: Is it needed? Is this empty when it's not a multi_exposure dialog?

myarjunar: yes it's needed, and it's empty when it's not a multi_exposure dialog

timlinux:
See https://github.com/inasafe/inasafe/issues/4612

safe/report/impact_report.py
Gustry: Ok, not a fan of specific properties which belong to the parent caller. But if you say it's needed ...

myarjunar: so what I did here is to avoid the calling of `ImpactReport.impact_function` (it was being used a lot in the extractors), because if we put the `MultiExposureImpactFunction()` as a value of `impact_function` parameter in the `ImpactReport` it will give error because `MultiExposureImpactFunction()` doesn't have a complete attribute as an 'ImpactFunction()', unless `MultiExposureImpactFunction()` is inherited from `ImpactFunction()`. any thoughts?

Gustry: No, let's go for it. It's not inheriting. The multiexposure is on top of the impact function now

myarjunar: thanks @Gustry I also wish for a better solution for this. Maybe I will just put a note why we do it for now

PR #4584, opened 2017-11-06
add custom order for layers provided by the multiexposure analysis

Gustry:
### What does it fix?
* Ticket: #4495 for the multiexposure
* Funded by: DFAT
* Description: add custom order for layers provided by the multiexposure analysis

### Checklist:

### DIFF
--- a/safe/gis/vector/summary_5_multi_exposure.py
+++ b/safe/gis/vector/summary_5_multi_exposure.py
@@ -156,6 +156,7 @@ def multi_exposure_analysis_summary(analysis, intermediate_analysis):
 
     analysis.commitChanges()
     analysis.keywords['title'] = layer_purpose_analysis_impacted['name']
+    analysis.keywords['layer_purpose'] = layer_purpose_analysis_impacted['key']
     if qgis_version() >= 21600:
         analysis.setName(analysis.keywords['title'])
     else:
@@ -228,6 +229,8 @@ def multi_exposure_aggregation_summary(aggregation, intermediate_layers):
 
     aggregation.commitChanges()
     aggregation.keywords['title'] = layer_purpose_aggregation_summary['name']
+    aggregation.keywords['layer_purpose'] = (
+        layer_purpose_aggregation_summary['key'])
     if qgis_version() >= 21600:
         aggregation.setName(aggregation.keywords['title'])
     else:
--- a/safe/gui/analysis_utilities.py
+++ b/safe/gui/analysis_utilities.py
@@ -17,6 +17,7 @@ from qgis.core import (
     QgsVectorLayer,
 )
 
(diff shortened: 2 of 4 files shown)


Gustry:
I will add a unittest in my next PR

timlinux:
@Gustry thanks

I tested your latest work - looking good. I still get this issue for some layers - but I guess it is just some dodgy data in our test data set?

screen shot 2017-11-07 at 09 11 38

Also when the result layers are added in the legend, would it be possible to keep all the other layers in the project too, but push them below the selected layers and hide them? Put another way: take the layers from the right hand box of the layer ordering, place them at the top of the layers list in a new group, making duplicates of layers if they already exist, make them visible and hide all other layers, leaving the original layers structure untouched.

Gustry:
Thanks @timlinux It's done:
screen shot 2017-11-07 at 09 15 57

I agree, it's much much better.

safe/gis/vector/summary_5_multi_exposure.py
```diff
@@ -228,6 +229,8 @@ def multi_exposure_aggregation_summary(aggregation, intermediate_layers):
 
     aggregation.commitChanges()
     aggregation.keywords['title'] = layer_purpose_aggregation_summary['name']
+    aggregation.keywords['layer_purpose'] = (
+        layer_purpose_aggregation_summary['key'])
```
myarjunar: I had issue with this as well, it says `aggregation` in the keywords before

safe/gui/analysis_utilities.py
```diff
@@ -252,26 +253,54 @@ def add_layers_to_canvas_with_custom_orders(order, impact_function):
...
-    # Need to remove all groups and layers
-    QgsMapLayerRegistry.instance().removeAllMapLayers()
     root = QgsProject.instance().layerTreeRoot()
-    root.removeAllChildren()
+    root.setVisible(False)  # Make all layers hidden.
 
-    # Insert layers in the good order.
-    for layer in order:
-        if layer[0] == FROM_CANVAS['key']:
+    group_analysis = root.insertGroup(0, impact_function.name)
```
myarjunar: do we have a definition for multi exposure string format? because I use raw string comparison in my PR for identifying a group whether it's multi exposure group or not

safe/gis/vector/summary_5_multi_exposure.py
Gustry: Yes I noticed. And I needed this fix too. Thanks for the hint. Git will merge automatically this for you because it's identical in your PR

safe/gui/analysis_utilities.py
Gustry: ouch, doing string comparison is very dangerous. I'm adding a flag on the group for you. Wait 5 minutes

Issue #4571, opened 2017-10-31
Incorrect report when use extent of exposure layer


samnawi:
### Problem
The report of landcover analysis using extent of exposure layer, generate an incorrect number in the report.

This example shows attribute of the result and the number that appear in the report, when I set the analysis using extent of exposure layer.
![image](https://user-images.githubusercontent.com/8871958/32207506-f4ee1a40-be2e-11e7-91db-4625c973e7a7.png)

But, when I set the analysis using extent of intersection hazard and exposure layers, show the right number in the report.
![image](https://user-images.githubusercontent.com/8871958/32207564-5fc4a50a-be2f-11e7-9e9b-23153d08c64a.png)

### Expected Behaviour
Information in the report has to show the right number.

### Environment
* InaSAFE : Develop
* QGIS : 2.14
* OS : Win7 64 Bit
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : 
  * Hazard : 
  * Aggregation :
  * Extent : 
  
[Test_Data2.zip](https://github.com/inasafe/inasafe/files/1429615/Test_Data2.zip)

Checklist (if applicable): 
* [ ] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? 

### Proposed Solution
Make it happy

cc : @timlinux @ismailsunni @Gustry

myarjunar:
Thanks for including the data! I will try to replicate @samnawi @lrpower

Gustry:
This is a problem in the impact function, not in reporting.

Gustry:
The problem is also in qgis layers, not the impact analysis summary, but in one level up (the aggregate hazard for instance).

I'm afraid that we have this issue since inasafe 4.0. The first hazard polygon (with id=0) has always been counted as `not exposed` regardless of the hazard class column. I need to check deeper. I made a PR, it should fix the issue, but I need to check for any side effect.

ismailsunni:
@Gustry do you think is it good idea to replicate this issue as a unit test? @samnawi 's data is a small data also.

Gustry:
I will keep this ticket, I will try something a little bit later. The PR has been merged so it should be fixed anyway.

PR #4596, opened 2017-11-08
fix if statement if value is 0

Gustry:
### What does it fix?
* Ticket: #4571
* Funded by: DFAT
* Description: This is a scary PR. We have this bug since 4.0. The thing is I was using a `if not value`. But in the ticket #4571, the value was `0` so then it was converted to `Not exposed`, but 0 was right.

I'm just waiting for travis first.

### Checklist:


timlinux:
Esh good catch @Gustry

timlinux:
@Gustry is still tweaking here ... let us know when it is ready for merge.

PR #4599, opened 2017-11-09
Bearing and distance post processors

ivanbusthomi:
### What does it fix?
* Ticket: #
* Funded by: 
* Description: 

### Checklist:


safe/gis/vector/tools.py
```diff
@@ -324,6 +324,47 @@ def read_dynamic_inasafe_field(inasafe_fields, dynamic_field, black_list=None):
...
+    :param point_a: First Point.
+    :type point_a: QgsPoint
+
+    :param point_b: Second Point.
+    :type point_b: QgsPoint
+
+    :return: The distance between input points.
+    :rtype: float
+    """
+    return point_b.azimuth(point_a)
```
Gustry: Maybe you can remove this function and call the azimuth straight in the parent function? It like more docstring than code :)

safe/processors/post_processor_functions.py
```diff
@@ -56,6 +58,82 @@ def size(size_calculator, geometry):
...
+    :type angle: float
+
+    :return: Cardinality text.
+    :rtype: str
+    """
+    # this method could still be improved later, since the acquisition interval
+    # is a bit strange, i.e the input angle of 22.499° will return `N` even
+    # though 22.5° is the direction for `NNE`
+    direction_list = [
+        'N', 'NNE', 'NE', 'ENE', 'E', 'ESE', 'SE',
```
Gustry: Can you add the `tr` function? It French, I need to replace `W` by `O` : https://upload.wikimedia.org/wikipedia/commons/thumb/2/26/Compass_Rose_French_North.svg/220px-Compass_Rose_French_North.svg.png

safe/processors/post_processors.py
```diff
@@ -79,6 +85,87 @@
     }
 }
 
+post_processor_distance = {
+    'key': 'post_processor_distance',
+    'name': tr('Distance Post Processor'),
+    'description': tr(
+        u'A post processor to calculate the distance between two points. The '
+        u'unit is defined in the exposure definition.'),
```
Gustry: Hum, does place has a unit? Because we use a place exposure, right?

safe/processors/__init__.py
```diff
@@ -40,6 +40,9 @@
     post_processor_fatalities,
     post_processor_displaced_ratio,
     post_processor_displaced,
+    post_processor_distance,
+    post_processor_bearing,
+    post_processor_cardinality
```
Gustry: Can you update the kind of tree a few lines before? Cardinality is a child of bearing because it's using the output of the bearing output.

safe/utilities/test/test_direction_distance.py
```diff
@@ -1,49 +0,0 @@
...
-        self.assertEqual(bearing_to_cardinal(157.5), 'SSE')
-        self.assertEqual(bearing_to_cardinal(180), 'S')
-        self.assertEqual(bearing_to_cardinal(0), 'N')
-        self.assertEqual(bearing_to_cardinal(-22.5), 'NNW')
-        self.assertEqual(bearing_to_cardinal(-45), 'NW')
-        self.assertEqual(bearing_to_cardinal(-67.5), 'WNW')
-        self.assertEqual(bearing_to_cardinal(-90), 'W')
-        self.assertEqual(bearing_to_cardinal(-112.5), 'WSW')
-        self.assertEqual(bearing_to_cardinal(-135), 'SW')
-        self.assertEqual(bearing_to_cardinal(-157.5), 'SSW')
```
Gustry: I guess we could keep these tests. Can you create the package `test` in the `processors` folder? (with __init__.py to make it a valid package)

Gustry:
Is Travis failing on another scenario? That might be affected because it's running now on every EQ impact function. But we want this postprocessor to run only for EQ on places, not EQ on buildings for instance. You need to add a new condition to the post processor.

Good anyway !

safe/processors/post_processor_functions.py
Gustry: maybe for a context in translations, it's ok to do 
`tr('N,NNE,NE,ENE,E, ESE,SE,all_of_them').split(',')` so we have only one string to translate in transifex and the translator can understand what it means and then the code will split each of them.

ivanbusthomi:
> You need to add a new condition to the post processor.

Yes, I kinda have that idea on the back of my mind, but not quite sure whether I should implement it or not. I will add new condition that it only run only on earthquake on places.

safe/utilities/test/test_direction_distance.py
ivanbusthomi: OK, I will update also the test for the post processors.

safe/processors/post_processors.py
ivanbusthomi: Leftover from copy and paste :) forget to clean it up.

safe/processors/__init__.py
ivanbusthomi: Thanks, for reminding me

safe/processors/post_processor_functions.py
ivanbusthomi: OK, I will add it.

safe/processors/post_processor_functions.py
```diff
@@ -56,6 +59,116 @@ def size(size_calculator, geometry):
...
+    :return: Cardinality text.
+    :rtype: str
+    """
+    # this method could still be improved later, since the acquisition interval
+    # is a bit strange, i.e the input angle of 22.499° will return `N` even
+    # though 22.5° is the direction for `NNE`
+    _ = earthquake_hazard, place_exposure # NOQA
+
+    direction_list = tr(
+        'N, NNE, NE, ENE, E, ESE, SE, SSE, S, SSW, SW, WSW, W, WNW, NW, NNW'
```
Gustry: You must remove spaces. It's only one string.

Gustry:
Good job!

Let's try the test in another PR

Issue #4441, opened 2017-09-19
[Realtime Porting] 


ismailsunni:
### Problem
We need to move functionality from InaSAFE realtime to InaSAFE core. E.g.
- [x] Evaluate cardinality : city, distance, and direction from epicentre (related to #4262 )
- [x] Nearest city (related to #4262 )
- [x] Grid converter to tiff file (implemented in InaSAFE)
- [x] Grid to mmi contour with smoothing function using scipy option #4615 
- [x] Multi exposure reporting and the calculation (separated into #4437 #4438 #4439 #4440 )

cc @timlinux @Gustry @myarjunar and @lucernae

akbargumbira:
I have the smoothing work on inasafe project without using scipy unfinished
(haven't tested it). Let me know if it could worth something. I'll clean it
up a bit and commit it.

ismailsunni:
@akbargumbira can you share your work somewhere? and is it fast enough?

akbargumbira:
@ismailsunni check it here (not even tested and separated the convolution function). The gaussian filter is taken from here https://github.com/nicjhan/gaussian-filter/blob/master/gaussian_filter.py (need to give the author credit if it's used at the end)

akbargumbira:
Please compare it to develop branch (I branched it from develop)

ismailsunni:
Thanks @akbargumbira I will check it.

ismailsunni:
All is done, I think we can close this.

PR #4600, opened 2017-11-10
bearing/direction/distance/name of the nearest place in QGIS expressions

Gustry:
### What does it fix?
* Ticket: #4441
* Funded by: DFAT
* Description: After the great work from @ivanbusthomi ;-)

The ordering is done by distance only on the layer itself before the end of the impact function.
It's like a quick hack, but I guess, as we don't have any other use case, I don't want to spend too much time on something cleaner.

Then some new expressions are introduced:
```
'Located ' || to_int(distance_to_nearest_place()) || 'm, ' ||  to_int(bearing_to_nearest_place()) || ' ˚ ' ||  direction_to_nearest_place() || ' of ' ||  name_of_the_nearest_place() 
```
it becomes
screen shot 2017-11-10 at 15 05 39

Some questions/todo:
* shouldn't we prefix all inasafe expressions by `inasafe_` to make a kind of namespace?
* The distance is always in meters for now. We need to improve this
* I guess we should remove layer_id from the project variables and use a hidden flag, like the multiexposure one, no @ismailsunni? It will clean the project variables.
* I noticed @ismailsunni it's using publicSource from the provenance to serialize etc. This won't work with postgis layers.

### Checklist:

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -18,6 +18,7 @@ from qgis.core import (
     QgsGeometry,
     QgsCoordinateTransform,
     QgsCoordinateReferenceSystem,
+    QgsFeatureRequest,
     QgsRectangle,
     QgsVectorLayer,
     QGis,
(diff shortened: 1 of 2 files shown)


myarjunar:
> shouldn't we prefix all inasafe expressions by inasafe_ to make a kind of namespace?

I was thinking about this too, but I thought it will be much harder for user to read so I went with a solution to name the group with inasafe, not the expression.

PR #4602, opened 2017-11-13
Smoothing

ismailsunni:
### What does it fix?
* Ticket: #4441 
* Funded by: DFAT
* Description: 
   - Adding smoothing with scipy (from realtime) and with numpy
   - Add smoothing option in the shake map converter dialog
      screen shot 2017-11-13 at 12 02 12
   - Scipy option will be disable if there is no scipy installed
   - The result using numpy and scipy is the same 
     screen shot 2017-11-13 at 12 04 55
   - Small improvement in shake map converter dialog (open file behaviour)

### Checklist:

### DIFF
--- a/safe/definitions/constants.py
+++ b/safe/definitions/constants.py
@@ -77,3 +77,9 @@ APPLICATION_NAME = 'inasafe'
 
 # Layer properties
 MULTI_EXPOSURE_ANALYSIS_FLAG = 'multi_exposure_analysis'
+
+
+# Smoothing mode
+NONE_SMOOTHING = 'none_smoothing'
+NUMPY_SMOOTHING = 'numpy_smoothing'
+SCIPY_SMOOTHING = 'scipy_smoothing'
--- a/safe/gui/tools/shake_grid/shakemap_converter_dialog.py
+++ b/safe/gui/tools/shake_grid/shakemap_converter_dialog.py
@@ -268,25 +268,37 @@ class ShakemapConverterDialog(QDialog, FORM_CLASS):
     def on_open_input_tool_clicked(self):
         """Autoconnect slot activated when open input tool button is clicked.
         """
+        input_path = self.input_path.text()
+        if not input_path:
+            input_path = os.path.expanduser('~')
         # noinspection PyCallByClass,PyTypeChecker
         filename = QFileDialog.getOpenFileName(
-            self, tr('Input file'), 'grid.xml', tr('Raw grid file (*.xml)'))
-        self.input_path.setText(filename)
+            self, tr('Input file'), input_path, tr('Raw grid file (*.xml)'))
+        if filename:
+            self.input_path.setText(filename)
 
     @pyqtSignature('')  # prevents actions being handled twice
     def on_open_output_tool_clicked(self):
         """Autoconnect slot activated when open output tool button is clicked.
         """
+        output_path = self.output_path.text()
+        if not output_path:
+            output_path = os.path.expanduser('~')
         # noinspection PyCallByClass,PyTypeChecker
         filename = QFileDialog.getSaveFileName(
-            self, tr('Output file'), 'grid.tif', tr('Raster file (*.tif)'))
-        self.output_path.setText(filename)
+            self, tr('Output file'), output_path, tr('Raster file (*.tif)'))
+        if filename:
+            self.output_path.setText(filename)
 
     @pyqtSignature('')
     def on_open_place_tool_clicked(self):
+        input_place = self.input_place.text()
+        if not input_place:
+            input_place = os.path.expanduser('~')
         filename = QFileDialog.getOpenFileName(
-            self, tr('Input place layer'), '', tr('All Files (*.*)'))
-        self.input_place.setText(filename)
+            self, tr('Input place layer'), input_place, tr('All Files (*.*)'))
+        if filename:
+            self.input_place.setText(filename)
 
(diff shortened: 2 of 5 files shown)


safe/gui/tools/shake_grid/shakemap_converter_dialog.py
```diff
@@ -238,28 +268,37 @@ def accept(self):
     def on_open_input_tool_clicked(self):
         """Autoconnect slot activated when open input tool button is clicked.
         """
+        input_path = self.input_path.text()
+        if not input_path:
+            input_path = os.path.expanduser('~')
```
Gustry: Something that would be very nice to fix for the user experience, it's to follow the last input directory. We should use the QSettings in the `UI` section.
screen shot 2017-11-13 at 07 43 28

I know a lot of users hate when a software doesn't remember your last used folder and bring you to your home directory everytime. Something we should think of when we create new dialog.

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
```diff
@@ -99,7 +120,7 @@ def on_output_path_textChanged(self):
     def on_input_path_textChanged(self):
         """Action when input file name is changed."""
         input_path = self.input_path.text()
-        input_not_grid_msg = self.tr('input file is not .xml')
+        input_not_grid_msg = tr('input file is not .xml')
```
Gustry: Should we do a search and replace in the code for `self.tr(` to `tr(` ?

Gustry:
I'm pretty surprised that shasum is giving the same result. One pixel different and the shasum must be different. But using two different python functions, with rounding etc ... Whaou.

So does it mean we don't need the pre-processor thing I created last week? :D

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
ismailsunni: Yeah, I think we can do that. It should be easy I think.

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
ismailsunni: Yes, I also hate it actually.
Is there a method or widget that do it automatically? Like checking `last` + `widget name or something` in QGIS?

ismailsunni:
> I'm pretty surprised that shasum is giving the same result. One pixel different and the shasum must be different. But using two different python functions, with rounding etc ... Whaou.

Yeah, me too. It's exactly the same. Actually, I just wanted to compare the result by checking the looks, but it turns out returning the same file.

> So does it mean we don't need the pre-processor thing I created last week? :D

Why? I thought we need it to make run automatically, don't we?

safe/definitions/constants.py
```diff
@@ -77,3 +77,9 @@
 
 # Layer properties
 MULTI_EXPOSURE_ANALYSIS_FLAG = 'multi_exposure_analysis'
+
+
+# Smoothing mode
+NONE_SMOOTHING = 'none_smoothing'
+NUMPY_SMOOTHING = 'numpy_smoothing'
+SCIPY_SMOOTHING = 'scipy_smoothing'
```
Gustry: Quick note, just noticed this weekend that Python as a Enum function: https://docs.python.org/3/library/enum.html
It has been backported to Python 2.7.

ismailsunni:
Btw, the one that use numpy is much slower than the scipy one.
```
2017-11-13T15:50:08	0	We are using NUMPY smoothing
2017-11-13T15:50:08	0	Duration : 0.325406
2017-11-13T15:50:51	0	We are using SCIPY smoothing
2017-11-13T15:50:51	0	Duration : 0.01058
```

It's like 30x slower. Is it fast enough @lucernae ?

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
Gustry: QgsFileWidget, but we still need to load/save the last used path.

safe/definitions/constants.py
ismailsunni: ah nice, we should use it in InaSAFE

Gustry:
> Why? I thought we need it to make run automatically, don't we?

Is it creating the contour file already?
We don't want to create it twice

ismailsunni:
> Is it creating the contour file already?
We don't want to create it twice

Yes. In the realtime, we create two shake map. One without smoothing, and one with smoothing. The one with smoothing is used for displaying the map only.
from realtime code:
```
        # Convert grid.xml (we'll give the title with event_id)
        # RM: convert event_id to str too. This avoid the layer name is
        # falsely read as int
        self.shake_grid = ShakeGrid(
            str(self.event_id), get_grid_source(), self.grid_file_path())
        # RM: add smoothed grid version. Only used for display in the map
        self.shake_grid_smooth = ShakeGrid(
            str(self.event_id), get_grid_source(), self.grid_file_path(),
            output_basename='mmi-smooth',
            smoothed=True)
```

Gustry:
Hum, ok.
So shouldn't we move the code to anoher file?

ismailsunni:
@Gustry I think so. We need to see if the parts the we created (multi exposure, direction, bearing, smoothing) are connected properly so that InaSAFE desktop can do what InaSAFE realtime currently does.

Can we merge this first, then check what is missing to replicate realtime?

PR #4608, opened 2017-11-13
use provenance keys and hide password

### LINKED ISSUE #4605: Python exception if we use a layer from PostGIS or WFS
### Problem

We can set keywords for a PostGIS layers. We could do that for many releases already:
screen shot 2017-11-13 at 07 12 59

* I don't want my password to be displayed in the dock. We should use `publicSource()`.

Then, I try to run the impact function, I have a Python exception:
```
Problem

Exception

Details

File dbname='stdm' host=localhost port=5433 user='etienne' sslmode=disable key='id' srid=4326 type=MultiPolygon table="public"."buildings" (geom) sql= had illegal extension
Diagnostics

In file "/Users/etienne/.qgis2/python/plugins/inasafe/safe/impact_function/impact_function.py", line 1411, in run get_provenance(self.provenance, provenance_exposure_layer))[0]
In file "/Users/etienne/.qgis2/python/plugins/inasafe/safe/gis/tools.py", line 126, in load_layer raise Exception(message)
```

I just went to the code where the exception is raised:
```
    if extension in ['.asc', '.tif', '.tiff']:
        layer = QgsRasterLayer(layer_path, base_name)
    elif extension in ['.shp', '.geojson', '.gpkg']:
        layer = QgsVectorLayer(layer_path, base_name, 'ogr')
    elif extension in ['.csv']:
        uri = 'file:///%s?delimiter=%s' % (layer_path, ',')
        layer = QgsVectorLayer(uri, base_name, 'delimitedtext')
    else:
        message = 'File %s had illegal extension' % layer_path
        raise Exception(message)
```

We should never check for extensions. QGIS supports many many extensions which are not written here: GML, KML, MapInfo etc ... And it doesn't work with services such as WFS, PostGIS, MySQL, ... (while we have the metadata.db to store keywords when it's not file based)

### Expected Behaviour

* Use `source()` if we need to programmatically fetch the real source of the layer
* Use `publicSource()` to display the layer source in the dock or in reports to not display passwords.
* Never check for extensions.
* Don't use hardcoded `OGR` provider such as `QgsVectorLayer(layer_path, base_name, 'ogr')`. The driver might be postgres, WFS ...

Gustry:
### What does it fix?
* Ticket: Partial fix #4605 
* Funded by: DFAT
* Description: use provenance keys and hide password

### Checklist:

### DIFF
--- a/safe/report/extractors/analysis_provenance_details.py
+++ b/safe/report/extractors/analysis_provenance_details.py
@@ -3,7 +3,13 @@
 from __future__ import absolute_import
 
 from collections import OrderedDict
+from qgis.core import QgsDataSourceURI
 
+from safe.definitions.provenance import (
+    provenance_exposure_layer,
+    provenance_hazard_layer,
+    provenance_aggregation_layer,
+)
 from safe.report.extractors.composer import QGISComposerContext
 from safe.report.extractors.util import resolve_from_dictionary, \
     jinja2_output_as_string
@@ -68,7 +74,9 @@ def analysis_provenance_details_extractor(impact_report, component_metadata):
         'hazard_keywords'])
 
     # hazard_keywords doesn't have hazard_layer path information
-    hazard_layer = impact_report.impact_function.provenance.get('hazard_layer')
+    hazard_layer = QgsDataSourceURI.removePassword(
+        impact_report.impact_function.provenance.get(
+            provenance_hazard_layer['provenance_key']))
     hazard_keywords['hazard_layer'] = hazard_layer
 
     # keep only value maps with IF exposure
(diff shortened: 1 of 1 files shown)


safe/report/extractors/analysis_provenance_details.py
```diff
@@ -68,7 +74,9 @@ def analysis_provenance_details_extractor(impact_report, component_metadata):
         'hazard_keywords'])
 
     # hazard_keywords doesn't have hazard_layer path information
-    hazard_layer = impact_report.impact_function.provenance.get('hazard_layer')
+    hazard_layer = QgsDataSourceURI.removePassword(
```
Gustry: This seems to be the first use of the source

safe/report/extractors/analysis_provenance_details.py
```diff
@@ -244,7 +254,8 @@ def analysis_provenance_details_simplified_extractor(
         'header': header,
         'provenance': provenance_format.format(
             layer_name=hazard_keywords.get('title'),
-            source=hazard_keywords.get('source') or default_source)
+            source=QgsDataSourceURI.removePassword(
+                hazard_keywords.get('source') or default_source))
```
Gustry: This is seems to be the second one

PR #4609, opened 2017-11-13
improve provider management

Gustry:
### What does it fix?
* Ticket: #4605 
* Funded by: DFAT
* Description: improve provider management

### Checklist:

### DIFF
--- a/safe/gis/tools.py
+++ b/safe/gis/tools.py
@@ -6,8 +6,19 @@ import logging
 import os
 
 from PyQt4.QtCore import QPyNullVariant
-from qgis.core import QgsRasterLayer, QgsVectorLayer, QgsMapLayerRegistry
+from qgis.core import (
+    QgsRasterLayer,
+    QgsVectorLayer,
+    QgsMapLayerRegistry,
+    QgsDataSourceURI,
+)
 
+from safe.definitions.constants import (
+    VECTOR_DRIVERS,
+    RASTER_DRIVERS,
+    OGR_EXTENSIONS,
+    GDAL_EXTENSIONS,
+)
 from safe.common.exceptions import NoKeywordsFoundError
 from safe.definitions.layer_geometry import (
     layer_geometry_raster,
@@ -16,7 +27,12 @@ from safe.definitions.layer_geometry import (
     layer_geometry_polygon,
 )
 from safe.utilities.gis import (
-    is_raster_layer, is_point_layer, is_line_layer, is_polygon_layer)
+    is_raster_layer,
+    is_point_layer,
+    is_line_layer,
+    is_polygon_layer,
+    qgis_version,
+)
 from safe.utilities.metadata import read_iso19115_metadata
 from safe.utilities.utilities import monkey_patch_keywords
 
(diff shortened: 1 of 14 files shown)


Gustry:
Gentle ping from 2 days ago ;-)
Thanks, I need this commit for serialization

safe/gis/tools.py
```diff
@@ -113,45 +129,194 @@ def reclassify_value(one_value, ranges):
...
+    :param layer_name: Layer name to use. Default to 'tmp'.
+    :type layer_name: basestring
+
+    :return: The layer or None if it's failed.
+    :rtype: QgsMapLayer
+    """
+    if provider in RASTER_DRIVERS:
+        return QgsRasterLayer(layer_uri, layer_name, provider)
+    elif provider in VECTOR_DRIVERS:
+        return QgsVectorLayer(layer_uri, layer_name, provider)
```
ismailsunni: What about delimitedtext? I saw in the other part of code, we need to use it explicitly.
```
            # Explicitly use URI with delimiter or tests fail in Windows. TS.
            layer = QgsVectorLayer(
                'file:///%s?delimiter=,' % layer_uri, layer_name, driver)
```

safe/gis/tools.py
```diff
@@ -113,45 +129,194 @@ def reclassify_value(one_value, ranges):
...
+            name = os.path.splitext(file_name)[0]
+        else:
+            # Might be a DB, take the DB name
+            source = QgsDataSourceURI(source)
+            name = source.table()
+
+    if not name:
+        name = 'default'
+
+    if qgis_version() >= 21800:
```
ismailsunni: IIRC, you told me that we don't need to do this? Or is it the other API?

safe/gis/tools.py
```diff
@@ -113,45 +129,194 @@ def reclassify_value(one_value, ranges):
...
+    if len(split) == 1:
+        return split[0], None
+    else:
+        return split
+
+
+def load_layer(full_layer_uri_string, name=None, provider=None):
+    """Helper to load and return a single QGIS layer based on our layer URI.
+
+    :param provider:
```
ismailsunni: Docstring.

safe/gis/tools.py
Gustry: The getter is the same `name()` whatever the QGIS version, but the setter is different between QGIS 2.14 and 2.18. So it's fine.

safe/gis/tools.py
Gustry: Yes, I was thinking about this one. But this function is used when the caller of this function known the provider name. So if the user known the provider name, I expected the caller to have a well formatted  URI

safe/gui/tools/wizard/test/test_keyword_wizard.py
```diff
@@ -2632,6 +2632,8 @@ def test_continuous_vector(self):
 
         self.assertDictEqual(real_keywords, expected_keyword)
 
+    @unittest.skip(
```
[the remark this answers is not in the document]
Gustry: I forgot this one. I wanted to fix it.

safe/gis/tools.py
ismailsunni: ah I see. But perhaps put the expected URI for this special  case (delimited text) to make the future usage of this method easier.

PR #4610, opened 2017-11-13
add unittest for map report and infographic expressions

myarjunar:
### What does it fix?
* Ticket: #4582 
* Funded by: DFAT
* Description: Here I try to add a simple unittest for map_report and infographic expressions first. html_report and other expressions that require parameters and legend in the app will come next because I still have no idea how to do it.

### Checklist:

### DIFF
--- a/safe/report/test/test_expressions.py
+++ b/safe/report/test/test_expressions.py
@@ -1,0 +1,275 @@
+# coding=utf-8
+"""Unittest for qgis expressions."""
+
+import unittest
+
+from qgis.core import QgsExpression
+
+from safe.definitions.reports.infographic import (
+    map_overview_header,
+    population_chart_header,
+    people_section_header,
+    age_gender_section_header,
+    age_gender_section_notes,
+    vulnerability_section_header,
+    female_vulnerability_section_header,
+    minimum_needs_section_header,
+    additional_minimum_needs_section_header,
+    minimum_needs_section_notes)
+from safe.definitions.reports.map_report import (
+    black_inasafe_logo_path,
+    white_inasafe_logo_path,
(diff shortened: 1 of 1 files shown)


safe/report/test/test_expressions.py
```diff
@@ -0,0 +1,262 @@
...
+        self.evaluate(disclaimer_title_header_element, expected_result)
+
+        # disclaimer_text_element
+        expected_result = (
+            u'InaSAFE has been jointly developed by the Indonesian '
+            u'Government-BNPB, the Australian Government, '
+            u'the World Bank-GFDRR and independent contributors. '
+            u'These agencies and the individual software developers of '
+            u'InaSAFE take no responsibility for the correctness of outputs '
+            u'from InaSAFE or decisions derived as a consequence.')
```
Gustry: Can we take these strings from definitions?
If some not python developers come to update definitions, it doesn't break tests which are only strings.
Same for other tests

myarjunar: so we don't want them to update the test as they update the definition?

Gustry: Well, they are not aware of these tests until you launch them on travis. (even me)

I think it's a little bit annoying when you change only one character in a string, it doesn't make sense to check the content of the string, especially these strings. You should be able to fix a typo without breaking tests. The test should only check the logic, that it returns a string with some content, not checking the actual content of the string.

To sum up, updating strings like this shouldn't break test.

My opinion about what is a good test. My 2 cents ;-)

ismailsunni: +1 for @Gustry 's comment

Issue #4613, opened 2017-11-14
Move PDF production to Print dialog only


timlinux:
### Problem

Remove auto generation of PDFs when IF is run
Update the batch runner to automatically call if.print()

This will make things much faster for users.

Gustry:
For multiexposure, the PDF is still generated before the print.

PR #4621, opened 2017-11-15
improve button status, remove duplicate group

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: improve button status, remove duplicate group when we have a custom ordering

@myarjunar I need your help after this merge. It still doesn't work. And I'm confused in what is needed for reporting.
Do you need all layers in the legend if the user sets a custom order?
Even if we are not generating PDF ? #4613 

Don't forget that Rizky will come with the multiexposure, so he will not use the multi_exposure_dialog so we shouldn't add too much logic in this file because he will use the web interface so better to use analysis_utilities.py to provide some helper

### Checklist:


Gustry:
Yes but the group is empty at the end of the process. I don't want to spend time if you roll back this behaviour in a few commits ;-)

Issue #4614, opened 2017-11-14
Make shakemap importer smoothing a developer only option


timlinux:
### Problem

Analysis should always be done on the raw shake. Making the smoothing part of the convertor might encourage users to run analysis on smoothed data. It will be useful to keep smoothing for developer mode, but we should hide it in production mode.

ismailsunni:
Done in #4624

PR #4624, opened 2017-11-16
Hide smoothing option in non developer mode.

ismailsunni:
### What does it fix?
* Ticket: #4614 
* Funded by: DFAT
* Description: 
   - Hide smoothing option in non developer mode

screen shot 2017-11-16 at 11 43 46

### Checklist:

### DIFF
--- a/safe/gui/tools/shake_grid/shakemap_converter_dialog.py
+++ b/safe/gui/tools/shake_grid/shakemap_converter_dialog.py
@@ -27,6 +27,7 @@ from safe.utilities.keyword_io import KeywordIO
 from safe.utilities.resources import html_footer, html_header, get_ui_class
 from safe.utilities.styling import mmi_ramp_roman
 from safe.utilities.i18n import tr
+from safe.utilities.settings import setting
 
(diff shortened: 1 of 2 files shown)


ismailsunni:
I will merge this, small change only.

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
```diff
@@ -94,6 +95,9 @@ def __init__(self, parent=None, iface=None, dock_widget=None):
         self.main_stacked_widget.setCurrentIndex(1)
         self.update_warning()
 
+        if not setting('developer_mode'):
+            self.smoothing_group_box.hide()
```
timlinux: Maybe its even better to remove it and then call [adjustsize](http://doc.qt.io/qt-4.8/qwidget.html#adjustSize)

ismailsunni: Will it make me need to check if the radio buttons exist or not in the `accept` method?

Issue #4649, opened 2017-11-20
Multi exposure: Print behaviour


Charlotte-Morgan:

Gustry:
The print behavior with a multiexposure analysis is not yet implemented. It will come very soon. We keep the ticket for after the merge.

Charlotte-Morgan:
@myarjunar - I see some discussion in the chat about default behaviour for multiexposure map when the user does not define the order of layers. may I suggest the following logic: vector over raster; points over lines over polys.
I think that will translate to buildings, roads and places over population and landcover.  I would put places on top.

PR #4653, opened 2017-11-21
Multi exposure with no keywords message

myarjunar:
### What does it fix?
* Ticket: #4649
* Funded by: DFAT
* Description: fix dock when selecting multi-exposure IF output layers and disable print button when multi-exposure IF layer is being selected

### Checklist:

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -43,7 +43,7 @@ from safe.definitions.constants import (
     PREPARE_FAILED_BAD_LAYER,
     PREPARE_SUCCESS,
     entire_area_item_aggregation,
-)
+    MULTI_EXPOSURE_ANALYSIS_FLAG)
 from safe.definitions.layer_purposes import (
     layer_purpose_hazard,
     layer_purpose_exposure,
@@ -87,6 +87,8 @@ from safe.gui.widgets.message import (
     ready_message,
     enable_messaging)
 from safe.impact_function.impact_function import ImpactFunction
+from safe.impact_function.multi_exposure_wrapper import (
+    MultiExposureImpactFunction)
 from safe.messaging import styles
 from safe.report.impact_report import ImpactReport
 from safe.report.report_metadata import ReportMetadata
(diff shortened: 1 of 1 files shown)


safe/gui/widgets/dock.py
```diff
@@ -756,7 +758,16 @@ def show_print_dialog(self):
                 keywords = self.keyword_io.read_keywords(active_layer)
 
                 provenances = keywords.get('provenance_data', {})
-                if provenances:
+                extra_keywords = keywords.get('extra_keywords', {})
+                is_multi_exposure = (
+                    extra_keywords.get(
+                        'analysis_type') == MULTI_EXPOSURE_ANALYSIS_FLAG)
```
Gustry: Is-there not an extra keywords key that you can use, from definitions?

safe/gui/widgets/dock.py
```diff
@@ -908,10 +919,21 @@ def layer_changed(self, layer):
             ]
 
             provenances = keywords.get('provenance_data', {})
+            extra_keywords = keywords.get('extra_keywords', {})
+            is_multi_exposure = (
+                extra_keywords.get(
+                    'analysis_type') == MULTI_EXPOSURE_ANALYSIS_FLAG)
```
Gustry: Same here

safe/gui/widgets/dock.py
```diff
@@ -930,6 +952,9 @@ def layer_changed(self, layer):
                         self.show_impact(table_report_path)
                         break
 
+            if self.impact_function.__class__ == MultiExposureImpactFunction:
```
Gustry: isinstance(self.impact_function, MultiExposureImpactFunction) is better I think than using using private members __XX__

Gustry:
Thanks, only small comments

safe/gui/widgets/dock.py
myarjunar: I don't find any, and I think my purpose on creating this `extra_keywords` at first is for this

safe/gui/widgets/dock.py
myarjunar: thanks

safe/gui/widgets/dock.py
Gustry: Yes there is: https://github.com/inasafe/inasafe/blob/develop/safe/definitions/extra_keywords.py#L14

myarjunar: ah you meant to use key from defintions, ok sure, missunderstanding :sweat_smile:

Gustry: Yes, using the key ;-)

PR #4662, opened 2017-11-22
map report for multi-exposure IF

myarjunar:
### What does it fix?
* Ticket: #4649 
* Funded by: DFAT
* Description: 

selection_156

### Checklist:

### DIFF
--- a/safe/gui/tools/print_report_dialog.py
+++ b/safe/gui/tools/print_report_dialog.py
@@ -6,7 +6,7 @@ import os
 
 from os.path import join, exists
 from PyQt4 import QtGui, QtCore, QtXml
-from qgis.core import QgsApplication, QgsExpressionContextUtils
+from qgis.core import QgsApplication
 
 from safe import messaging as m
 from safe.common.signals import send_error_message, send_static_message
@@ -17,7 +17,8 @@ from safe.definitions.reports.components import (
     map_report,
     infographic_report,
     standard_impact_report_metadata_pdf,
-    all_default_report_components)
+    all_default_report_components,
+    standard_multi_exposure_impact_report_metadata_pdf)
 from safe.definitions.utilities import (
     override_component_template, definition, update_template_component)
 from safe.gui.tools.help.impact_report_help import impact_report_help
@@ -25,6 +26,7 @@ from safe.messaging import styles
 from safe.report.impact_report import ImpactReport
 from safe.report.report_metadata import ReportMetadata
 from safe.utilities.i18n import tr
+from safe.utilities.keyword_io import KeywordIO
 from safe.utilities.resources import (
     get_ui_class, resources_path, html_header, html_footer)
 from safe.utilities.settings import setting
@@ -233,15 +235,23 @@ class PrintReportDialog(QtGui.QDialog, FORM_CLASS):
         report_path = os.path.dirname(impact_layer.source())
 
         # Get the hazard and exposure definition used in current IF
-        hazard = definition(
-            QgsExpressionContextUtils.projectScope().variable(
-                'hazard_keywords__hazard'))
-        exposure = definition(
-            QgsExpressionContextUtils.projectScope().variable(
-                'exposure_keywords__exposure'))
+        hazard_keywords = KeywordIO.read_keywords(
+            impact_layer, 'hazard_keywords')
+        exposure_keywords = KeywordIO.read_keywords(
+            impact_layer, 'exposure_keywords')
+        hazard = hazard_keywords.get('hazard', None)
+        exposure = exposure_keywords.get('exposure', None)
+        if hazard and exposure:
+            hazard = definition(hazard)
+            exposure = definition(exposure)
+        else:
+            hazard = None
+            exposure = None
 
         standard_impact_report_metadata = ReportMetadata(
             metadata_dict=standard_impact_report_metadata_pdf)
+        standard_multi_exposure_impact_report_metadata = ReportMetadata(
+            metadata_dict=standard_multi_exposure_impact_report_metadata_pdf)
         standard_map_report_metadata = ReportMetadata(
             metadata_dict=update_template_component(
                 component=map_report,
@@ -256,6 +266,7 @@ class PrintReportDialog(QtGui.QDialog, FORM_CLASS):
 
         standard_report_metadata = [
             standard_impact_report_metadata,
+            standard_multi_exposure_impact_report_metadata,
             standard_map_report_metadata,
             standard_infographic_report_metadata,
             custom_map_report_metadata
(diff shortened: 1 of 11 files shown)


safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -282,8 +296,8 @@ def aggregation_summary(self):
         return self._aggregation_summary
 
     @property
-    def analysis_summary(self):
-        """Property for the aggregation summary.
+    def analysis_impacted(self):
```
Gustry: We try to move to the new names: `analysis_summary` https://github.com/inasafe/inasafe/blob/develop/safe/definitions/layer_purposes.py#L134
Why did you change this one?

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -395,6 +409,24 @@ def current_impact_function(self):
         """
         return self._current_impact_function
 
+    @property
+    def output_layers_ordered(self):
+        """Return the custom order inout from user.
```
Gustry: s/inout/input/g

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -471,6 +503,9 @@ def _generate_provenance(self):
             provenance_data_store_uri,
             self.datastore.uri_path)
 
+        # Map title
+        set_provenance(self._provenance, provenance_map_title, self.name)
```
Gustry: Side question, is it not already `provenance_impact_function_name`?

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -876,6 +912,13 @@ def run(self):
 
             analysis_layers.append(impact_function.analysis_impacted)
             aggregation_layers.append(impact_function.aggregation_summary)
+            list_of_analysis_path.append(
+                impact_function.analysis_impacted.source())
+
+        set_provenance(
+            self._provenance,
+            provenance_layer_exposure_summary,
```
Gustry: Maybe better to create its own provenance because this one is a list. Look for `multi` in the provenance definition file.

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -876,6 +912,13 @@ def run(self):
 
             analysis_layers.append(impact_function.analysis_impacted)
             aggregation_layers.append(impact_function.aggregation_summary)
+            list_of_analysis_path.append(
+                impact_function.analysis_impacted.source())
```
Gustry: Using `source`, we don't know which provider it is (postgres, ogr...), can you switch to https://github.com/inasafe/inasafe/blob/develop/safe/gis/tools.py#L132 and the next function to read it?

safe/impact_function/multi_exposure_wrapper.py
myarjunar: I just try to make it matches the `ImpactFunction` , IMHO we should have the same property name for the same purpose, either to change what's in `ImpactFunction` or change what's in `MultiExposureImpactFunction` . no pressure, very flexible :smile:

safe/impact_function/multi_exposure_wrapper.py
myarjunar: yes, but I need map title provenance, and I can't think something better than the name to be the map title

safe/impact_function/multi_exposure_wrapper.py
Gustry: Agree. `analysis_impacted` was the old name, before the release of 4.0. Better to update `ImpactFunction` I think

safe/impact_function/multi_exposure_wrapper.py
Gustry: Ok, no problem, let's do that until they want a better naming

safe/impact_function/multi_exposure_wrapper.py
myarjunar: can I do this in another PR?

safe/gui/tools/print_report_dialog.py
```diff
@@ -233,15 +234,21 @@ def print_as_pdf(self):
         report_path = os.path.dirname(impact_layer.source())
 
         # Get the hazard and exposure definition used in current IF
-        hazard = definition(
-            QgsExpressionContextUtils.projectScope().variable(
-                'hazard_keywords__hazard'))
-        exposure = definition(
-            QgsExpressionContextUtils.projectScope().variable(
-                'exposure_keywords__exposure'))
+        hazard = QgsExpressionContextUtils.projectScope().variable(
```
Gustry: Sorry, I just realized.  We want to clean these variables in the project scope.
The end user can't use them out of the box so IMHO it's not good to expose them to public. Can you fetch these keywords with the impact layer?

safe/impact_function/multi_exposure_wrapper.py
Gustry: So why not revert back this one?

Gustry: Because you need to use the same property in `ImpactReport` ?

myarjunar: I use this somewhere in the code actually, ok let me revert this back and make some conditional code first

safe/gui/tools/print_report_dialog.py
myarjunar: yes of course

safe/impact_function/multi_exposure_wrapper.py
myarjunar: > Because you need to use the same property in ImpactReport ? Go for it, it's fine

yes,  you know me so well :smiley:  ok then, it's easier to refactor if we leave it like this

safe/report/extractors/composer.py
```diff
@@ -240,29 +240,40 @@ def qgis_composer_extractor(impact_report, component_metadata):
...
+        layers = [] + impact_report.extra_layers
+
+    add_supplementary_layers = (
+        not impact_report.multi_exposure_impact_function or not (
+            impact_report.multi_exposure_impact_function.output_layers_ordered)
+    )
+    if add_supplementary_layers:
+        # Check show only impact.
+        show_only_impact = setting(
+            'set_show_only_impact_on_report', False, bool)
```
ismailsunni: We have set the default value in the `default_settings.py`. So, you can remove the default value (False) here.

safe/report/extractors/composer.py
```diff
@@ -240,29 +240,40 @@ def qgis_composer_extractor(impact_report, component_metadata):
...
+            aggregation_layer_id = provenance['aggregation_layer_id']
+            if aggregation_layer_id:
+                aggregation_layer = layer_registry.mapLayers().get(
+                    aggregation_layer_id, None)
+                layers.append(aggregation_layer)
+
+            layers.append(hazard_layer)
+
+        # check hide exposure settings
+        hide_exposure_flag = setting('setHideExposureFlag', False, bool)
```
ismailsunni: Same comment like the previous one.

myarjunar:
thanks for the review @ismailsunni @Gustry

PR #4668, opened 2017-11-22
fix error when hide exposure flag is false

myarjunar:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: fix error when hide exposure flag is false

### Checklist:

### DIFF
--- a/safe/report/extractors/composer.py
+++ b/safe/report/extractors/composer.py
@@ -14,6 +14,9 @@ from qgis.core import QgsMapLayerRegistry
 
 from safe.common.version import get_version
 from safe.definitions.fields import analysis_name_field
+from safe.definitions.provenance import (
+    provenance_exposure_layer_id,
+    provenance_multi_exposure_layers_id)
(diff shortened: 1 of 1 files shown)


Gustry:
Hum, it might be a good catch
I totally forgot this option, and I wonder which behaviour we should have when we set the custom order. I think we must discard it

Is-it fixing the issue we mentioned with Tim?

myarjunar:
shout out to your french exception message :grinning: I don't know about Tim, he may have a different error.

> I think we must discard it

I just apply the same behaviour as single IF, lets wait until they say otherwise :smiley:

Gustry:
Can you use definitions?

safe/report/extractors/composer.py
```diff
@@ -14,6 +14,8 @@
 
 from safe.common.version import get_version
 from safe.definitions.fields import analysis_name_field
+from safe.definitions.provenance import provenance_exposure_layer_id, \
```
ismailsunni: backslash

Issue #4669, opened 2017-11-22
Using integer values for age/gender counts results in zero values for demographic breakdown


wcarthur:
### Problem
If using integer fields for the age/gender counts in exposure data, the resulting demographic breakdown gives zero values. The ratios are calculated, but because the values are integer, the resulting ratios all get determined to be zero. 

The log messages (below) show the ratios are being calculated, as the count fields are available.

```
2017-11-23T07:07:24	0	exposure: Compute ratios from counts
2017-11-23T07:07:24	0	Count field exposure_id_field not detected in the exposure. We will not compute a ratio from this field.
2017-11-23T07:07:24	0	Count field exposure_name_field not detected in the exposure. We will not compute a ratio from this field.
2017-11-23T07:07:24	0	Count field infant_count_field detected in the exposure, we are going to create a equivalent field infant_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field child_count_field detected in the exposure, we are going to create a equivalent field child_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field youth_count_field detected in the exposure, we are going to create a equivalent field youth_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field adult_count_field detected in the exposure, we are going to create a equivalent field adult_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field elderly_count_field detected in the exposure, we are going to create a equivalent field elderly_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field male_count_field detected in the exposure, we are going to create a equivalent field male_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field female_count_field detected in the exposure, we are going to create a equivalent field female_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field under_5_count_field detected in the exposure, we are going to create a equivalent field under_5_ratio_field in the exposure layer.
2017-11-23T07:07:24	0	Count field over_60_count_field detected in the exposure, we are going to create a equivalent field over_60_ratio_field in the exposure layer.
```

### Expected Behaviour
Age/gender ratios should be float values between 0 and 1. 

### Environment
* InaSAFE : 4.2.4
* QGIS : 2.18.6
* OS : Windows 7 (64-bit)
* Data used (filenames and paths so as to reproduce, a project is better):
  * Exposure : 
  * Hazard : 
  * Aggregation :
  * Extent : 
  
[...]

PR #4670, opened 2017-11-22
Fix 4669, ratio is 0 if using integer count

ismailsunni:
### What does it fix?
* Ticket: #4669 
* Funded by: DFAT
* Description: 
  - Convert to float when calculating ratio

### Checklist:


safe/gis/vector/from_counts_to_ratios.py
```diff
@@ -97,7 +97,7 @@ def from_counts_to_ratios(layer, callback=None):
         for count_field, index in mapping.iteritems():
             count = feature[count_field]
             try:
-                new_value = count / total_count
+                new_value = count / float(total_count)
```
Gustry: Would be nice if you can add reference to the ticket

PR #4677, opened 2017-11-23
Fix Welcome Message on Windows

ismailsunni:
### What does it fix?
* Ticket: No ticket
* Funded by: DFAT
* Description: 
  - The welcome message is keep showing although it's already unchecked.

### Checklist:

### DIFF
--- a/safe/plugin.py
+++ b/safe/plugin.py
@@ -801,7 +801,7 @@ class Plugin(object):
         # Set previous_version to the current inasafe_version
         set_setting('previous_version', inasafe_version)
 
-        if setting('always_show_welcome_message'):
+        if setting('always_show_welcome_message', expected_type=bool):
             # Show if it the setting said so
             show_message = True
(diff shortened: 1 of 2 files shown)


ivanbusthomi:
LGTM mas @ismailsunni

safe/plugin.py
```diff
@@ -801,7 +801,7 @@ def show_welcome_message(self):
         # Set previous_version to the current inasafe_version
         set_setting('previous_version', inasafe_version)
 
-        if setting('always_show_welcome_message'):
+        if setting('always_show_welcome_message', expected_type=bool):
```
Gustry: in the setting helper, you can add automatically the expected_type based on the type in default_settings.py and not having to specify the type here

Might be good to do the same with default value

ismailsunni: ah, brilliant.

I will add a ticket for it. Not sure if I had time today. I still don't understand why this code is working.

PR #4690, opened 2017-11-27
multiexposure improvments

Gustry:
### What does it fix?
* Ticket: #4688
* Funded by: DFAT
* Description: 
  * Add title to layers if a custom order is used
  * Always report exceptions in the message viewer if an exception is raised from the reporting
  * Manage settings about hide exposure layer, zoom to impact layer
  * Disable `AttributeError` from KeyWords IO because it's hiding us some real error.

### Checklist:

### DIFF
--- a/safe/gui/tools/multi_exposure_dialog.py
+++ b/safe/gui/tools/multi_exposure_dialog.py
@@ -193,7 +193,6 @@ class MultiExposureDialog(QDialog, FORM_CLASS):
                 error = ''
                 layer.exportNamedStyle(style_document, error)
 
-                # For a raster, the provider type is 'gdal'
                 layers.append((
                     FROM_CANVAS['key'],
                     layer.name(),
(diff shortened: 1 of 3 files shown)


safe/gui/tools/multi_exposure_dialog.py
```diff
@@ -588,12 +590,18 @@ def accept(self):
                         0, analysis.name)
                     detailed_group.setVisible(Qt.Checked)
                     add_impact_layers_to_canvas(analysis, group=detailed_group)
+
+                if self.iface:
+                    self.iface.setActiveLayer(
+                        self._multi_exposure_if.analysis_impacted)
```
myarjunar: I think we should put this outside the `if generate_report` condition, just like the behaviour in single IF

Gustry: I guess only me is running this option. The checkbox is displayed only with developer mode is ON and you restarted QGIS before so I think it's quite specific

Gustry: I just checked the code, this line is already outside the `if generate_report` ...

Issue #4630, opened 2017-11-17
More profile polishing


Charlotte-Morgan:
Hi @ismailsunni - this new profile wizard and the extra options tab looks great.  I would like to offer some suggestions on the text and a little bit about functionality. I apologise if any of my suggestions have already been considered - I'm sure some have.

**In the first run wizard: Welcome Screen**
1.  I propose the text be tweaked to say:

_Welcome to InaSAFE. Before you begin using InaSAFE, we would like to encourage you to review some options that will used by InaSAFE. On the "Organisation Profile" tab you can set options related to your organisation such as: the name and logo.
The "parameters" tab is used to define some parameters that will be used by InaSAFE in the  analysis of exposed population.  You have the option to change the parameters for whether exposed people are considered to be "affected" by each hazard type and class and the displacement rate that will be used for those affected people. 
Please refer to the InaSAFE manual for concept definitions and more information on the source of hazard classifications and default settings. We really encourage you to consider these parameters carefully and to choose appropriate values for your local situation based on past events and expert knowledge. 
You can return to these options any time by using the Plugins -> InaSAFE -> Options menu._

It would be great to add a link to the manual if that is possible.

2. It is possible to add a save option on this welcome screen or maybe a statement at the end of the welcome text about being able to save the options if you return to them at Plugins -> InaSAFE -> Options menu

3. I think the check box text should say "Always show welcome message when opening InaSAFE" or "Always show welcome message when opening QGIS with InaSAFE"

--------------------------------
**In the first run wizard: Parameters tab**
1. I propose we change the text to say " Please set parameters for population exposed to each hazard class below." 
and then add a link to the definitions for affected and displaced people from the manual.

ie make the following text appear on the parameters tab, just like we do in the reports.

Affected People | People who are affected by a hazardous event. People can be affected directly or indirectly. Affected people may experience short-term or long-term consequences to their lives, livelihoods or health and in the economic, physical, social, cultural and environmental assets. In InaSAFE, people who are killed during the event are also considered affected.UNISDR (2015)Proposed Updated Terminology on Disaster Risk Reduction: A Technical Review
-- | --

Displaced People | Displaced people are people who, for different reasons and circumstances because of risk or disaster, have to leave their place of residence. In InaSAFE, demographic and minimum needs reports are based on displaced / evacuated people.UNISDR (2015)Proposed Updated Terminology on Disaster Risk Reduction: A Technical Review
-- | --

[...]

PR #4691, opened 2017-11-28
Filter out non-population classification.

ismailsunni:
### What does it fix?
* Ticket: #4630 
* Funded by: DFAT
* Description: 
   - Filter out non population classification

screen shot 2017-11-28 at 07 47 48

### Checklist:

### DIFF
--- a/safe/definitions/utilities.py
+++ b/safe/definitions/utilities.py
@@ -679,7 +679,7 @@ def get_provenance(provenance_collection, provenance_dict):
     return provenance_collection.get(provenance_dict['provenance_key'])
 
 
-# TODO(IS): Add parameter to get from setting, or just get it from setting
+# TODO(IS): Add parameter to filter the exposure
 def generate_default_profile():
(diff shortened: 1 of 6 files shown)


Charlotte-Morgan:
Hi @ismailsunni - that looks good.
Do you plan to rename the tab from parameters to population parameters
The column widths might need tweaking again now that the classification column has shorter strings. As a user I can adjust the affected column width (even make it so wide the others disappear) and column order but not the width of the other columns.
image

BTW - I have only tested the interface not if making changes to things is reflected in the results.

ismailsunni:
Hi @Charlotte-Morgan 
I have rename the tab, and make the column width like this:
screen shot 2017-11-30 at 21 37 37

Another option that I can offer is same column width for all column, like this:
screen shot 2017-11-30 at 21 40 31

I will merge this PR for tomorrow walkthrough, and we can discuss more about the column width

Gustry:
The first one is better

Issue #3434, opened 2016-11-14
use renameAttributes if QGIS > 2.16


Gustry:
### Problem
For now, we are copying a field and removing the old one so as to rename a field.

In QGIS 2.16, we can now rename an attribute : 
https://qgis.org/api/classQgsVectorDataProvider.html#ae1661cdfd4cc785e5b1bf5afa23b7f3d

We will earn some processing time.

ismailsunni:
I found this also when we were starting working on new architecture, but only to realise for QGIS 2.16 :)

timlinux:
@Gustry lets move this into the 4.1 milestone unless you have some concrete plans to implement it now (presumably with an ifdef based on QGIS version so that the old code path still works?

PR #4694, opened 2017-11-28
use renameFields if available

Gustry:
### What does it fix?
* Ticket: #3434 
* Funded by: DFAT
* Description: use renameFields if available

Todo, check the `prepare_vector_layer` step. I'm not sure when we need to copy or where we can rename

### Checklist:


Gustry:
@ismailsunni Can you check the `def sum_fields` ? I think it came with the FMT tool

ismailsunni:
yes, it does. LGTM. I also wanted to use renameFields when I code it, but it's not available in 2.14.
Thanks

PR #4699, opened 2017-11-29
unicode is not str, use basestring instead

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: unicode is not str, use basestring instead

PR to master

That's something we should take care of @ismailsunni @myarjunar 

### Checklist:


ismailsunni:
Ah I see what you mean. Yeah, the expected value is important. Perhaps we should move to use definition for each setting and put the type and expected value.
LGTM.

Issue #4304, opened 2017-07-07
Analysis fail to complete


idyuham:
### Problem

I was trying to run analysis of floods on roads, then building then on pop, all of the failed. I run them separately.  UGANDA dateset provided for training in Kampala.
```
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1135, in run self._run()
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1296, in _run self.gis_overlay_analysis()
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1417, in gis_overlay_analysis self.aggregate_hazard_preparation()
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1806, in aggregate_hazard_preparation self._aggregate_hazard_impacted = union(self.hazard, self.aggregation)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\gis\vector\union.py", line 101, in union index_b = create_spatial_index(union_a)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\utilities\profiling.py", line 122, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/STEVE SAPIN/.qgis2/python/plugins\inasafe\safe\gis\vector\tools.py", line 251, in create_spatial_index spatial_index = QgsSpatialIndex(layer.getFeatures())
```

### Environment
* InaSAFE : 
* QGIS : 
[...]

PR #4749, opened 2017-12-07
display an error if we can not create the spatial index

Gustry:
### What does it fix?
* Ticket: #4304 or discussion on gitter https://gitter.im/inasafe/inasafe?at=5a2903d487680e6230e0359a
* Funded by: DFAT
* Description: display an error if we can not create the spatial index

I would be happy to not display a message like this and to provide a suggestion. But I'm clueless for now @timlinux 

### Checklist:

### DIFF
--- a/safe/gis/vector/clean_geometry.py
+++ b/safe/gis/vector/clean_geometry.py
@@ -5,6 +5,7 @@
 from safe.definitions.processing_steps import clean_geometry_steps
 from safe.gis.sanity_check import check_layer
 from safe.utilities.profiling import profile
+from safe.common.custom_logging import LOGGER
 
 __copyright__ = "Copyright 2016, The InaSAFE Project"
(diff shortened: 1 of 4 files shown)


safe/gis/vector/clean_geometry.py
```diff
@@ -5,6 +5,7 @@
 from safe.definitions.processing_steps import clean_geometry_steps
 from safe.gis.sanity_check import check_layer
 from safe.utilities.profiling import profile
+from safe.common.custom_logging import LOGGER
```
ismailsunni: Ah nice, I should just import from somewhere, not creating new one everytime :D

Gustry: one line instead of one.
Actually, you can do it from any file with `LOGGER` in it already. But maybe better to use this one in `custom_logging`

Issue #4625, opened 2017-11-16
Metadata Converter Polishing


ismailsunni:
### Problem
Basic functionality is working and merged to develop. But there are still some polishing to do:

- [x] Add unit test for the dialog (@ismailsunni)
- [x] Icon for the tool @MariaSolovyeva 
- [x] Fill the help text properly (currently lorem ipsum) @timlinux 

I will make it release blocker since it's not good to have wrong icon or missing help text.

Reference: #4604

Charlotte-Morgan:
see #4732 for UI issue and #4733 for screenie

ismailsunni:
Last item is done, closing this.

PR #4758, opened 2017-12-08
Add unit test for metadata converter dialog.

ismailsunni:
### What does it fix?
* Ticket: #4625 
* Funded by: DFAT
* Description: 
   - Add unit test for metadata converter dialog.

### Checklist:

### DIFF
--- a/safe/gui/tools/test/test_metadata_converter.py
+++ b/safe/gui/tools/test/test_metadata_converter.py
@@ -1,0 +1,66 @@
+# coding=utf-8
+
+"""Tests for metadata converter dialog."""
+
+import unittest
+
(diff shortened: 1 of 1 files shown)


ismailsunni:
> Usually, we make a PR with code and without test.
> But this PR is unusual and suspicious, there is only test in it!

aha, yes, because when I work on the metadata converter, we need it immediately for @timlinux 's training, so I didin't create the unit test (only put it as to do list in an issue). Now, I have time, so I created it. It's only the dialog's behaviour (tested manually already)

What's the thing that I need to change?

Gustry:
it was a joke ;-)
I said the PR was suspicious because no production code added.

ismailsunni:
ah, too sleepy :D thanks

Issue #4776, opened 2017-12-12
Polish Print Dialog


ismailsunni:
### Problem
- [x] Disable option that not possible, for example check box for `Print infographic report as PDF` if the 
analysis doesn't have it (only population IIRC)
- [x] Open the current template file's directory when choosing the custom template file.
- [ ] It's quite strange to have `Open as PDF` in the left (on MacOS), just make sure it's consistent with the other tool.
- [x] In the 3rd option of `Custom and map report`, give more explanantion how to add the template override (e.g. `You can add the [override_filename] in the [directory] to use this option`)

myarjunar:
> In the 3rd option of Custom and map report, give more explanantion how to add the template override (e.g. You can add the [override_filename] in the [directory] to use this option)

this is already there in the help of the dialog

myarjunar:
> It's quite strange to have Open as PDF in the left (on MacOS), just make sure it's consistent with the other tool.

is it still bothering you ? it is in normal order on Ubuntu @ismailsunni

ismailsunni:
I think it's fine (still in the left, but it's not a big problem). I will close this.

PR #4777, opened 2017-12-12
polish on print dialog

myarjunar:
### What does it fix?
* Ticket: #4776 
* Funded by: DFAT
* Description: 

### Checklist:

### DIFF
--- a/safe/gui/tools/print_report_dialog.py
+++ b/safe/gui/tools/print_report_dialog.py
@@ -12,6 +12,7 @@ from qgis.core import QgsApplication
 from safe import messaging as m
 from safe.common.signals import send_error_message, send_static_message
 from safe.definitions.constants import ANALYSIS_FAILED_BAD_CODE
+from safe.definitions.exposure import exposure_population
 from safe.definitions.provenance import (
     provenance_hazard_keywords, provenance_exposure_keywords)
 from safe.definitions.reports import (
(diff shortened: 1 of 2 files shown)


safe/gui/tools/print_report_dialog.py
```diff
@@ -556,11 +563,16 @@ def template_chooser_clicked(self):
 
         .. versionadded: 4.3.0
         """
+        path = setting('lastCustomTemplate', '', expected_type=str)
```
ismailsunni: Better to take from the current text box.

ismailsunni: Expect a unicode, just in case there is unicode character

safe/gui/tools/print_report_dialog.py
```diff
@@ -556,11 +563,18 @@ def template_chooser_clicked(self):
 
         .. versionadded: 4.3.0
         """
+        path = self.template_path.text()
+        if not path:
+            path = setting('lastCustomTemplate', '', expected_type=unicode)
```
Gustry: I would use `basestring`: https://github.com/inasafe/inasafe/commit/39eecfe1926e5d6014cdbf286dd45c96f89706c2
If you have `str`, it will fail with an exception because `str` ≠ `unicode`

I would very happy to move to Python 3. We have some accents in path, it's always scary for me.

ismailsunni: Ah sorry, I think @Gustry  is right.

PR #4779, opened 2017-12-13
avoid duplicate label in print dialog

Gustry:
### What does it fix?
* Ticket: #
* Funded by: DFAT
* Description: Remove the `print report as PDF`. I'm not a UX designer, but it's less words in the UI

Before:
screen shot 2017-12-13 at 10 01 55

After:
screen shot 2017-12-13 at 10 03 38

### Checklist:

### DIFF
--- a/safe/gui/ui/print_report_dialog.ui
+++ b/safe/gui/ui/print_report_dialog.ui
@@ -25,8 +25,8 @@
         <number>0</number>
        </property>
        <item row="0" column="0">
-        <widget class="QWebView" name="help_web_view" native="true">
-         <property name="url" stdset="0">
+        <widget class="QWebView" name="help_web_view">
+         <property name="url">
           <url>
            <string>about:blank</string>
           </url>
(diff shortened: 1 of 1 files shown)


ismailsunni:
LGTM. But perhaps @timlinux can suggest more :D 

I was thinking to make it a question, like:
> Which report that you want to print as PDF?

Gustry:
I'm happy to change of course.
In `Custom and Map report`, we should also avoid the word `print` as it will open the composer.

My personal experience:
* If the label is "print the map report as PDF", it's OK, not the best, I would prefer `Open as PDF`.
* If the label is "Print the map report", then people think that they need to switch on their printer and and they won't click if they don't have a physical printer. So I would like to avoid using the word "print" only, and use "open" or another word. @ismailsunni @timlinux ?

ismailsunni:
Yeah, I agree. But then we should also change the `Print` button in dock, shouldn't we?

Gustry:
Let's merge it for now, we can still review if needed later

Issue #4803, opened 2017-12-18
Print dialog again


Charlotte-Morgan:
In our 4.3 walk through we had some shared concerns that the print dialog may be a bit wordy and confusing for users. we also discussed moving some text to help

Here are some suggestions to make it a little simpler. .. skip to the mock up below....
![image](https://user-images.githubusercontent.com/5743135/34090496-8d5ff58a-e40b-11e7-8416-ada73311bbd3.png)
main point is to give the user a choice of templates and only state once that the default will not be used if they choose one of the other options.

here is my mock up - with apologies to expert mocker uppers
![image](https://user-images.githubusercontent.com/5743135/34090570-27c29bd2-e40c-11e7-87fb-8a47d207370a.png)

The dots on the top should all be possible - ie can select all
the custom map should only be possible if the print map report option is not selected 
only one option should be possible in the custom map reports section

myarjunar:
thanks for the feedback @Charlotte-Morgan , it looks much simpler than before. But what about putting the defaut map reports in the map report group? because the user should be able to open the default map template in qgis composer. so it would be something like this:
![impact report - preview _166](https://user-images.githubusercontent.com/11134669/34092722-746b8cba-e3f7-11e7-99ac-f9c6fbc6e237.png)

Or, we could just go with your mockup and simply put the `open in composer` button at the button group in the bottom.

Gustry:
@Charlotte-Morgan You should update from develop, we made some updates last Thursday about wording in this dialog

Issue #4795, opened 2017-12-15
Polish Impact Report Dialog


ismailsunni:
- [x] Fix the layout size (set ideal size)
- [x] Enable infographic for population in the multiexposure if there is population

PR #4826, opened 2017-12-21
Another print dialog polish

myarjunar:
### What does it fix?
* Ticket: #4803 #4795 
* Funded by: DFAT
* Description: 
impact report_168

impact report_172

### Checklist:

### DIFF
--- a/safe/gui/tools/help/impact_report_help.py
+++ b/safe/gui/tools/help/impact_report_help.py
@@ -65,7 +65,7 @@ def content():
 
     bullets = m.BulletedList()
     bullets.add(m.Text(
-        m.ImportantText(tr('Tabular reports')),
+        m.ImportantText(tr('InaSAFE reports')),
         tr(
             ' - There are four checkboxes available which are representing '
             'the type of report component that will be generated.'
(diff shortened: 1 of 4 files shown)


timlinux:
Can you share how it looks currently?

Gustry:
Can you change to `Print the map as PDF`?
According to my comments here https://github.com/inasafe/inasafe/pull/4779#issuecomment-351354602 when we put use `Print` without the `as PDF`

myarjunar:
@Gustry do you mean change the `Open as PDF` button to `Print the map as PDF`?

Gustry:
`Print the map using`.

I would also remove `InaSAFE report`, maybe only `Reports`, because we are already in InaSAFE, so it looks redundant.

myarjunar:
@Gustry ah ok, I'm just following what's in #4803 and there is still no other feedback, personally I prefer what's @Charlotte-Morgan proposing in #4803

Gustry:
Ah yes ok.
Will you just update the `as PDF` ?

timlinux:
One other thing we discussed @myarjunar was that the old text should be moved into the help dialog strings. Can you do this please? Add a paragraph for each checkbox / option in the dialog with the current (as per your screenshot above) text and then a : then the old detailed explanation.

Also please check that that help text goes into the definitions_help (it needs a couple of lines to add it into the list of help texts for dialogs).

Thanks.

Issue #4806, opened 2017-12-18
Can't open an analysis made on another computer


Gustry:
### Problem

if I open a project with an analysis made on another computer:
```
Problem
The following problem(s) were encountered whilst running the analysis.

Exception
Suggestion
You can try the following to resolve the issue:

Check that you have the latest version of InaSAFE installed - you may have encountered a bug that is fixed in a subsequent release.
Check the InaSAFE documentation to see if you are trying to do something unsupported.
Report the problem using the issue tracker at https://github.com/inasafe/inasafe/issues. Reporting an issue requires that you first create a free account at http://github.com. When you report the issue, please copy and paste the complete contents of this panel into the issue to ensure the best possible chance of getting your issue resolved.
Try contacting one of the InaSAFE development team by sending an email to info@inasafe.org. Please ensure that you copy and paste the complete contents of this panel into the email.
Details
These additional details were reported when the problem occurred.

Layer "C:/AIFDR II - SDA/10_InaSAFE Analysis/20171215 Java EQ/Exposure/Java_Population_WGS84.tif" is not valid
Diagnostics

In file "/Users/etienne/.qgis2/python/plugins/inasafe/safe/gui/widgets/dock.py", line 972, in layer_changed keywords))
In file "/Users/etienne/.qgis2/python/plugins/inasafe/safe/impact_function/impact_function.py", line 2703, in load_from_output_metadata impact_function.exposure = load_layer_from_registry(exposure_path)
In file "/Users/etienne/.qgis2/python/plugins/inasafe/safe/gis/tools.py", line 87, in load_layer_from_registry return load_layer(layer_path)[0]
In file "/Users/etienne/.qgis2/python/plugins/inasafe/safe/gis/tools.py", line 217, in load_layer raise Exception(message)
```

I'm not using the computer `Layer "C:/AIFDR II - SDA/10_InaSAFE Analysis/20171215 Java EQ/Exposure/Java_Population_WGS84.tif" is not valid`

### Expected Behaviour

InaSAFE should open the HTML reports and be friendly without an exception.

@myarjunar This errors occurs because we are relying on the layer.source()
But the project in QGIS is fine, it could open all layers from the project:

![screen shot 2017-12-18 at 09 37 06](https://user-images.githubusercontent.com/1609292/34096827-10116f02-e3d7-11e7-8028-8425f4af557f.png)

Any idea we can handle that correctly?
[...]

PR #4827, opened 2017-12-21
add handler for invalid layer path

myarjunar:
### What does it fix?
* Ticket: #4806 
* Funded by: DFAT
* Description: 

### Checklist:

### DIFF
--- a/safe/gui/widgets/dock.py
+++ b/safe/gui/widgets/dock.py
@@ -25,7 +25,8 @@ from safe.common.exceptions import (
     NoKeywordsFoundError,
     InvalidParameterError,
     HashNotFoundError,
-    MetadataReadError)
+    MetadataReadError,
+    InvalidLayerError)
 from safe.common.signals import (send_static_message, send_error_message)
 from safe.common.version import get_version
 from safe.defaults import supporters_logo_path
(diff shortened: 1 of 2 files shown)


timlinux:
LGTM thanks

safe/gui/widgets/dock.py
```diff
@@ -931,14 +931,17 @@ def layer_changed(self, layer):
...
+            try:
+                if is_multi_exposure:
+                    self.impact_function = (
+                        MultiExposureImpactFunction.load_from_output_metadata(
+                            keywords))
+                elif provenances:
+                    self.impact_function = (
+                        ImpactFunction.load_from_output_metadata(
+                            keywords))
+            except Exception:
```
Gustry: This kind of `except Exception` is hiding us all possible bugs raised by these functions. I'm personally against this because the user and mainly us won't have any information about what is wrong in `load_from_metadata`. Is-it not possible to have more granular catch statement?

ismailsunni: +1

myarjunar: fixed, I raised `InvalidLayerError` exception instead of `Extension` so we know which extension is  caught

Issue #4615, opened 2017-11-14
Smoothed MMI and contours should be produced as pre-processor step


timlinux:
### Problem

We should automatically generate the smoothed MMI grid and contours during IF run and then make this available as 'future products' so that they can be used in the multi-exposure dialog.

ismailsunni:
The smoothing is happening in the parsegrid function, in the ShakeGrid class. So, we need the grid.xml file to get the contour in the current structure. There are two options:

1. Put the grid.xml URI in the `extra_keywords ` of the shakemap.tif. It should be easy to do. But, I am wondering will it work properly in GeoNode / GeoSAFE / Realtime.
    a. GeoNode doesn't accept xml only file to be uploaded (CMIIW). So, we can't upload the grid.xml
    b. We only able to upload shakemap.tif to the GeoNode, that means we need to convert grid.xml first then upload shakemap.tif to the GeoNode. If we convert it in desktop, obviously the geonode couldn't find the grid.xml, then GeoSAFE can't have the contour.
    c. If it's in realtime, may be we can make the URI of grid.xml accessible from GeoSAFE also. Then it will be fine.

2. Change the structure so that we can do this conversion : grid.xml --> shakemap.tif --> grid.xml (or csv). This has some advantages:
     a. No need external file, since we can do everything from the shakemap
     b. Easier to use in realtime
    But, we should make sure we can do the shakemap.tif --> grid.xml (or csv) part. I checked the code, it use a gdal command `gdal_grid`:
```
/Library/Frameworks/GDAL.framework/Versions/Current/Programs/gdal_grid -a nearest -zfield "mmi" -txe 122.14 128.14 -tye 2.465 8.435 -outsize 241 240 -of GTiff -ot Float16 -a_srs EPSG:4326 -l mmi "/Users/ismailsunni/Downloads/20170429042741-grid-vakso.vrt" "/Users/ismailsunni/Downloads/20170429042741-grid-vakso-nearest.tif"
```
I am not sure this command is reversible (to make csv from tif).

What do you think @timlinux @Gustry @lucernae ?

timlinux:
@ismailsunni Maybe I am just missing something but:
if we have a gdal dataset (tif) and you want to process it with numpy
it is trivial to turn that gdal layer into a numpy array
we dont need to do such complex workflows
e.g. https://gis.stackexchange.com/questions/32995/how-to-fully-load-a-raster-into-a-numpy-array

[...]

Issue #4519, opened 2017-10-12
Can we drop layer_geometry in keywords?


Gustry:
### Problem

In the XML, we have a section like this:
```
          <layer_geometry>
            <gco:CharacterString>polygon</gco:CharacterString>
          </layer_geometry>
```

AFAIK, this is not used anymore in InaSAFE for reading. Only writing operations mostly.

I'm not old enough in InaSAFE, but I would say this was introduced when InaSAFE was using Numpy.

Is there any reason to hardcode the geometry type in metadata?
Or is-it ok if dynamically read the geometry type using the QGIS API with this function https://github.com/inasafe/inasafe/blob/develop/safe/gis/tools.py#L23 ?

Can we some inputs from @ismailsunni and @timlinux if we can drop this item from the metadata?

ismailsunni:
I am +0 for this. I forget why we put it as keywords, perhaps for making it explicit in the metadata. I hope you can recall it @timlinux

Gustry:
I'm trying to figure out how we can fit our metadata into the QGIS schema. And if it's legacy and not used anymore, I'm happy to do some cleaning and keeping our metadata model as small as possible.

Gustry:
Maybe be for volcano point, when InaSAFE was doing the buffering inside the impact function and we wanted to know the original geometry when doing the report?

PR #4832, opened 2017-12-23
Contour Creation for Earthquake

ismailsunni:
### What does it fix?
* Ticket: #4615 
* Funded by: DFAT
* Description: 
   - Add shakemap converter by converting it to ascii first
   - Remove reclassifiy in the shakemap converter dialog to make InaSAFE able creating contour from the shakemap better
   - More separation for smoothing using numpy
   - Tweak shakemap converter dialog
   - Set proper classification for shakemap tif from the shakemap converter
      screen shot 2017-12-23 at 17 00 18
   - Unit test
   - Put the contour creation in the pre processor
   - The new workflow
      shake-smoothing
   - Sample running:
       eq_contour_pre_process

### Checklist:


timlinux:
Looking good @ismailsunni - can you make a few more tweaks to the UI then LGTM to merge:

Input File -> Input file

Algorithm -> Interpolation Algorithm

Use ASCII -> None

ismailsunni:
Merging this....

safe/gis/raster/contour.py
```diff
@@ -0,0 +1,527 @@
...
+    field_definition = ogr.FieldDefn('ROMAN', ogr.OFTString)
+    layer.CreateField(field_definition)
+    # So that we can set the label horizontal alignment
+    field_definition = ogr.FieldDefn('ALIGN', ogr.OFTString)
+    layer.CreateField(field_definition)
+    # So that we can set the label vertical alignment
+    field_definition = ogr.FieldDefn('VALIGN', ogr.OFTString)
+    layer.CreateField(field_definition)
+    # So that we can set feature length to filter out small features
+    field_definition = ogr.FieldDefn('LEN', ogr.OFTReal)
```
Gustry: Since 4.0, all our fields are defined from definitions. Why not the same behavior for X, Y, RGB, Roman, Align, etc ?
You added new definitions about these fields but you are not using them, so we might change these definitions later, it won't make any effects in this code.

safe/gui/tools/shake_grid/test/test_shake_grid.py
```diff
@@ -62,62 +81,75 @@ def check_feature_count(path, count):
...
+        normal_shakemap_tif = NORMAL_SHAKE_GRID.mmi_to_raster()
+        normal_contour_shp = NORMAL_SHAKE_GRID.mmi_to_contours()
+
+        smoothed_shakemap_tif = SMOOTHED_SHAKE_GRID.mmi_to_raster()
+        smoothed_contour_shp = SMOOTHED_SHAKE_GRID.mmi_to_contours()
+
+        print normal_shakemap_tif
+        print normal_contour_shp
+        print smoothed_shakemap_tif
+        print smoothed_contour_shp
```
Gustry: We can move to python 3: `print()` it works on python two too ;-)

safe/processors/__init__.py
```diff
@@ -60,5 +60,6 @@
 ]
 
 pre_processors = [
-    # pre_processors_nearby_places, Disabled for now until with have contours
+    pre_processor_earthquake_contour  # Disable now, until we have metadata
```
Gustry: The comment is a left over?

safe/processors/pre_processors.py
```diff
@@ -79,3 +82,59 @@ def fake_nearby_preprocessor(impact_function):
...
+def earthquake_contour_preprocessor(impact_function):
+    """Preprocessor to create contour from an earthquake
+
+    :param impact_function: Impact function to run.
+    :type impact_function: ImpactFunction
+
+    :return: The contour layer.
+    :rtype: QgsMapLayer
+    """
+    from safe.gis.raster.contour import create_smooth_contour
```
Gustry: Any reason for this import here?

safe/processors/pre_processors.py
```diff
@@ -79,3 +82,59 @@ def fake_nearby_preprocessor(impact_function):
...
+    :param impact_function: Impact function to check.
+    :type impact_function: ImpactFunction
+
+    :return: If the preprocessor can run.
+    :rtype: bool
+    """
+    hazard_key = impact_function.hazard.keywords.get('hazard')
+    is_earthquake = hazard_key == hazard_earthquake['key']
+
+    geometry_key = impact_function.hazard.keywords.get('layer_geometry')
```
Gustry: this keyword is not very well maintained. Why not using the QGIS API to check this property? `is_raster_layer` exists in `safe/utilities/gis.py` See ticket https://github.com/inasafe/inasafe/issues/4519

safe/gis/raster/contour.py
ismailsunni: Yes, I want to do that also. I already tried it actually. But this one is using `ogr`, I can't use the `create_field_from_definition` method. Perhaps I will create new one for `ogr`. But yes, I agree with field creation from the definitions.

Gustry: Yes I saw, we can add one for OGR, it would be the best option. But at least we should read the dictionary for the name

Gustry: BTW, maybe in QGIS 3, we will not use OGR to create the dataset, so maybe just read the dictionary. We will be able to read the numpy array from the QGIS api.

safe/processors/pre_processors.py
ismailsunni: Done.

safe/processors/pre_processors.py
ismailsunni: Done, move up.

safe/processors/__init__.py
ismailsunni: Hehe, yes. Removed.

safe/gui/tools/shake_grid/test/test_shake_grid.py
ismailsunni: Removed.

safe/gis/raster/contour.py
ismailsunni: Ouch, I just created the function.

PR #4837, opened 2018-01-03
Fix 4788 polish population parameter

ismailsunni:
### What does it fix?
* Ticket: No ticket
* Funded by: DFAT
* Description: 
   - Fix restore button (previously it will restore not from the default value)
   - Use definition for the general content, and update the value based on the user's setting. Previously, it will directly use from the user one. It will make a problem if don't have a definition. Another feature by doing this, we can have the classes sorted like in definition.

### Checklist:

### DIFF
--- a/safe/gui/tools/options_dialog.py
+++ b/safe/gui/tools/options_dialog.py
@@ -3,6 +3,7 @@
 
 import logging
 import os
+from functools import partial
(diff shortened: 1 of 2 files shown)


ismailsunni:
Merging this, important bug fix.

Issue #4857, opened 2018-01-11
Needed Information for Realtime


ismailsunni:
### Problem
To replicate realtime analysis, we need to provide some information to get richer report. I already made some of them, but I think we are still missing some of them. Here the list that I have made:

```python
earthquake_extra_keywords = [
    extra_keyword_earthquake_latitude,
    extra_keyword_earthquake_longitude,
    extra_keyword_earthquake_magnitude,
    extra_keyword_earthquake_depth,
    extra_keyword_earthquake_description,
    extra_keyword_earthquake_location,
    extra_keyword_earthquake_event_time,
    extra_keyword_time_zone,
    extra_keyword_earthquake_x_maximum,
    extra_keyword_earthquake_x_minimum,
    extra_keyword_earthquake_y_maximum,
    extra_keyword_earthquake_y_minimum,
]

ash_extra_keywords = [
    extra_keyword_volcano_name,
    extra_keyword_region,
    extra_keyword_eruption_height,
    extra_keyword_volcano_eruption_event_time,
    extra_keyword_volcano_alert_level,
    extra_keyword_time_zone,
    extra_keyword_volcano_latitude,
    extra_keyword_volcano_longitude,
]
```

There are some more:
- [x] Vent height for volcano https://github.com/inasafe/inasafe-realtime/issues/151#issuecomment-356172268 (get it from data, we should make it as an extra keywords)
- [ ] Shakemap ID for earthquake. Where can I get it @lucernae ?

Please add @ivanbusthomi @samnawi @felix-yew @lucernae

lucernae:
@ismailsunni The parameter needed for hazard events:

Earthquake:
- All extra keywords you put in the issue + following extras.
- event_id/shake ID: from the grid.xml , which is event_time. But we have two kinds of shakemap. please refer to https://github.com/inasafe/inasafe-realtime/issues/60. I advise that this comes prefilled from grid.xml, but can be modified by user (if user convert grid.xml from InaSAFE desktop). Because we have 2 kinds of grid.xml and it might be unreliable to get the shake ID from the second one, because they mean different thing
- extra keywords to know the **source** of the EQ (so we can identify the type of grid.xml). In our terminologies, it can be **initial** or **post-processed/data-informed**. We have no reliable way to determine this from grid.xml. So, I advise we include it so user fill it from combo-box or text field when user convert the grid.xml

Ash:
- All extra keywords you put in the issue + following extras.
[...]

PR #4867, opened 2018-01-15
Fix 4857 missing realtime information

ismailsunni:
### What does it fix?

* Ticket: #4857 
* Funded by: DFAT
* Description: 
  * Make user able to set custom event ID and source type in 

shakemap_auto_id

### Checklist:

### DIFF
--- a/safe/definitions/extra_keywords.py
+++ b/safe/definitions/extra_keywords.py
@@ -38,6 +38,15 @@ extra_keyword_region = {
 }
 
 # Earthquake Extra Keywords
+extra_keyword_earthquake_event_id = {
+    'key': 'earthquake_event_id',
+    'name': tr('Earthquake event ID'),
+    'description': tr(
+        'The ID of the earthquake event or shakemap. It is constructed from '
+        'the timestamp of the event in YYYYMMDDHHmmSS format.'),
+    'type': unicode
+}
+
 extra_keyword_earthquake_latitude = {
     'key': 'earthquake_latitude',
     'name': tr('Latitude'),
@@ -117,6 +126,32 @@ extra_keyword_earthquake_y_maximum = {
     'description': tr(
         'The maximum value of y coordinate of the shakemaps. It indicates the '
         'extent of the event.')
+}
+
+extra_keyword_earthquake_source = {
+    'key': 'earthquake_source',
+    'name': tr('Source'),
+    'description': tr(
+        'Source of the earthquake, it can be initial or post-processed '
+        '(data-informed).'),
+    'type': unicode,
+    'options': [
+        {
+            'key': 'initial',
+            'name': tr('Initial'),
+            'description': tr(
+                'This shakemap is created in 5 minutes after the earthquake. '
+                'No extra processing happens.')
+        },
+        {
+            'key': 'post-processed',
+            'name': tr('Post processed'),
+            'description': tr(
+                'This shakemap is created after 5 minutes after the '
+                'earthquake. There are some processing happen in this '
+                'shakemap.')
+        }
+    ]
 }
 
 
@@ -258,6 +293,7 @@ extra_keyword_volcano_height = {
 }
 
 earthquake_extra_keywords = [
+    extra_keyword_earthquake_event_id,
     extra_keyword_earthquake_latitude,
     extra_keyword_earthquake_longitude,
     extra_keyword_earthquake_magnitude,
@@ -270,9 +306,11 @@ earthquake_extra_keywords = [
     extra_keyword_earthquake_x_minimum,
     extra_keyword_earthquake_y_maximum,
     extra_keyword_earthquake_y_minimum,
+    extra_keyword_earthquake_source,
 ]
 
 ash_extra_keywords = [
+    extra_keyword_volcano_event_id,
     extra_keyword_volcano_name,
     extra_keyword_region,
     extra_keyword_eruption_height,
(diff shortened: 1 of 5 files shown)


safe/definitions/extra_keywords.py
```diff
@@ -113,8 +128,46 @@
...
 
 # Volcano Extra Keywords
+extra_keyword_volcano_event_id = {
+    'key': 'volcano_event_id',
+    'name': tr('Volcano event ID'),
+    'description': tr(
+        'The ID of the volcano eruption. It is constructed from '
+        'YYYYMMDDHHmm[zoneoffset]_[volcano_name]. YYYYMMDDHHmm is the format '
+        'of the eruption event time. [zone offset] is the of set of its time '
+        'zoen. [volcano_name] is the name of the volcano. For example: '
```
Gustry: zone

Gustry: zone

safe/definitions/extra_keywords.py
```diff
@@ -113,8 +128,46 @@
...
+    'description': tr(
+        'Source of the earthquake, it can be initial or post-processed '
+        '(data-informed).'),
+    'type': unicode,
+    'options': [
+        {
+            'key': 'initial',
+            'name': tr('Initial'),
+            'description': tr(
+                'This shakemap is created in 5 minutes after the earthquake. '
```
Gustry: I would also remove reference to realtime in InaSAFE desktop. Can we change this 5 minutes? Is not specific to realtime? Maybe `This shakemap is created just after an EQ`. ? I'm not quite happy with what I'm proposing too.

safe/definitions/extra_keywords.py
```diff
@@ -113,8 +128,46 @@
         'extent of the event.')
 }
 
+extra_keyword_earthquake_source = {
+    'key': 'earthquake_source',
+    'name': tr('Source'),
+    'description': tr(
+        'Source of the earthquake, it can be initial or post-processed '
+        '(data-informed).'),
```
Gustry: I think, outside of realtime, InaSAFE users will struggle a little bit to understand this one. Advice?

safe/definitions/extra_keywords.py
```diff
@@ -113,8 +128,46 @@
...
+
 
 # Volcano Extra Keywords
+extra_keyword_volcano_event_id = {
+    'key': 'volcano_event_id',
+    'name': tr('Volcano event ID'),
+    'description': tr(
+        'The ID of the volcano eruption. It is constructed from '
+        'YYYYMMDDHHmm[zoneoffset]_[volcano_name]. YYYYMMDDHHmm is the format '
+        'of the eruption event time. [zone offset] is the of set of its time '
```
Gustry: `off`

safe/definitions/extra_keywords.py
Gustry: Add unit?

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
```diff
@@ -228,14 +240,28 @@ def accept(self):
...
 
+        extra_keywords = {}
+        if self.check_box_custom_shakemap_id.isChecked():
+            event_id = self.line_edit_shakemap_id.text()
+            extra_keywords[extra_keyword_earthquake_event_id['key']] = event_id
+
+        current_index = self.combo_box_source_type.currentIndex()
+        source_type = self.combo_box_source_type.itemData(current_index)
+        LOGGER.debug(current_index)
+        LOGGER.debug(source_type)
```
Gustry: Just a question, if we let these debug statements, can we have more descriptive informations? When someone send us some debug message, it make it difficult to debug when only something like `yes`, `3`, `false`, `tab changed` :D

safe/gui/tools/wizard/step_kw57_extra_keywords.py
```diff
@@ -127,8 +135,10 @@ def set_widgets(self):
             # Timezone
             timezone_checkbox = QCheckBox()
             timezone_combo_box = QComboBox()
-            for timezone in pytz.common_timezones:
-                timezone_combo_box.addItem(timezone, timezone)
+            for timezone in extra_keyword_time_zone['options']:
+                timezone_combo_box.addItem(
+                    timezone['key'], timezone['name']
+                )
             index = timezone_combo_box.findText('Asia/Jakarta')
```
Gustry: Maybe you can move it to definitions? I like the way you did `options`. `default_option` ? Feel free to skip this request

safe/definitions/extra_keywords.py
ismailsunni: Yeah, I understand what you mean. I also will struggle with it. But, I don't what's the alternative for this case. What do you think @lucernae ?

safe/definitions/extra_keywords.py
ismailsunni: Unit for what?

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
ismailsunni: oops, it means to be deleted. Personal use only :D

safe/definitions/extra_keywords.py
Gustry: the offset?

safe/definitions/extra_keywords.py
ismailsunni: I changed it to:
> This shakemap is created in just after the earthquake '
                'without any extra processing happens.

since the main point is extra processing's existence.

safe/definitions/extra_keywords.py
ismailsunni: Fixed.

safe/gui/tools/shake_grid/shakemap_converter_dialog.py
Gustry: We don't have a clear policy about these debug messages. If it's debug level, I don't mind if we have a kind of sentence to introduce the variable

PR #4884, opened 2018-01-25
Flood extra keywords

ismailsunni:
### What does it fix?

* Ticket: #4857 
* Funded by: DFAT
* Description: 
  * Add flood event ID and flood event date time

### Checklist:

### DIFF
--- a/safe/definitions/extra_keywords.py
+++ b/safe/definitions/extra_keywords.py
@@ -40,6 +40,18 @@ extra_keyword_region = {
     'type': unicode
 }
 
+# Flood Extra Keywords
+extra_keyword_flood_event_id = {
+    'key': 'flood_event_id',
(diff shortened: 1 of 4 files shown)


safe/definitions/extra_keywords.py
```diff
@@ -375,6 +387,11 @@
     extra_keyword_volcano_longitude,
 ]
 
+flood_extra_keywords = [
+    extra_keyword_flood_event_id,
+    extra_keyword_volcano_eruption_event_time,
```
ivanbusthomi: Hi @ismailsunni - why does the volcano eruption event time is added as the flood extra keywords?

ismailsunni: Ah yes, sorry, forget to add then the autocomplete finish it :D. Fixed in the incoming commit.

Issue #4888, opened 2018-01-26
Not Able to Open Moved Analysis Layers


ismailsunni:
### Problem
After I move my analysis directory to another place (same computer), I try to load all the layers in QGIS. But, when I load it, I got this error message:
```
Layer "/var/folders/h7/k16m33695_g2ptlzy0z8vxb00000gp/T/inasafe/2017-10-31/ismailsunni/impacts/VolcanicAshRasterOnRoadsLine_26January2018_07h55-28.808346/hazard_aggregation_summary.geojson" is not valid
```

It seems InaSAFE read the impact layer metadata, and try to open the layer. Unfortunatelly, since I have moved the whole directory to the new location, the path to that layer is no longer valid.

### Expected Behaviour
InaSAFE should be able to open a moved analysis.

### Environment
* InaSAFE : 4.3.1
  
Checklist: 

* [ ] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? 

### Proposed Solution
- Perhaps try to load it as file in the same directory?

ismailsunni:
Unfortunatelly, the IF.load from metadata method only use the content of the metadata (not able to get the path to the metadata).
Perhaps we need to add more parameter to it to know where is the metadata located.

Gustry:
More or less linked to https://github.com/inasafe/inasafe/issues/4806

timlinux:
So we will try to read from the analysis XML the impact layers using the absolute path specified. If the layer open fails, we try to use a relative path to find the layer in the analysis directory.

PR #4891, opened 2018-01-30
fix hazard, exposure, and aggregation layer id on IF serialization

myarjunar:
### What does it fix?

* Ticket: #
* Funded by: DFAT
* Description: 
  * Fix a problem on inasafe-headless when generating report (only impact layer on map canvas)

### Checklist:

### DIFF
--- a/safe/report/extractors/composer.py
+++ b/safe/report/extractors/composer.py
@@ -261,11 +261,25 @@ def qgis_composer_extractor(impact_report, component_metadata):
         if not show_only_impact:
             hazard_layer = layer_registry.mapLayers().get(
                 provenance['hazard_layer_id'], None)
+            if not hazard_layer:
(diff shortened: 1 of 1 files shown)


Gustry:
The same as @ismailsunni. Either add a ticket number or please add some description. We don't really know what you are fixing. But yes the code is LGTM.
Is-it to find layers based on their IDs if the path doesn't exist but the layer is in the legend (the layer has moved like in @ismailsunni #4888 ?)

ismailsunni:
> Is-it to find layers based on their IDs if the path doesn't exist but the layer is in the legend (the layer has moved like in @ismailsunni #4888 ?)

No, that ticket is more on the loading IF from a metadata file. We need to pass the location of the metadata, so that we can find the layers in the same directory as in the metadata file.

Issue #4724, opened 2017-12-04
We need to replicate the realtime reports in  desktop


timlinux:

timlinux:
* Ash fall
* Flood in Jakarta
* EQ

ivanbusthomi:
This is the first test realtime report of volcanic ash using InaSAFE v.4.3
![ash_v1](https://user-images.githubusercontent.com/7351977/35207284-aaf786e2-ff74-11e7-89b6-97cf9c091fdd.png)

there are some elements that still need to be updated, but the information shown there are already based on the analysis.

ivanbusthomi:
This is the first test realtime report of earthquake using InaSAFE v.4.3

![earthquake_replicate](https://user-images.githubusercontent.com/7351977/35272139-d5ae9a72-0066-11e8-8a0e-2f5fbee1a909.png)

ivanbusthomi:
This is the first test realtime report of flood using InaSAFE v.4.3
there is some information at the top of the report that still missing, maybe we can add that information from the website side?
cc @ismailsunni @myarjunar @lucernae 

![flood_v1](https://user-images.githubusercontent.com/7351977/35318727-3a46a1de-010f-11e8-8125-5ac1c24c7843.png)

ivanbusthomi:
Hi @Gustry and @myarjunar - this is what I found a little bit strange.
so basically I used two different expression, utilizing information from two different analysis summary layer:
* fatalities information from population analysis summary layer - using `inasafe_impact_analysis_field_value('fatalities') `, 
* nearby places information from place analysis summary layer using `name_of_the_nearest_place()`

when I tried selecting either layer in the layers panel, only one of the information is shown at a time.

![composer](https://user-images.githubusercontent.com/7351977/35322074-89c89fd0-011b-11e8-89fd-31a756d10f97.gif)

ivanbusthomi:
This is the project files:
📦 [Realtime ash](https://github.com/inasafe/inasafe/files/1659969/realtime_ash.zip)
📦 [Realtime earthquake](https://github.com/inasafe/inasafe/files/1659970/realtime_earthquake.zip)
📦 [Realtime flood](https://github.com/inasafe/inasafe/files/1659971/realtime_flood.zip)

Each of them contains the qpt templates. Each project file also contains a template in the composer manager.

ismailsunni:
Hi @ivanbusthomi 
[...]

PR #4907, opened 2018-02-15
add expression for multi exposure analysis specific

myarjunar:
### What does it fix?

* Ticket: #4724 
* Funded by: 
* Description: 
  * fix expression so @ivanbusthomi can work with the case as mentioned in #4724 

### Checklist:

### DIFF
--- a/safe/gis/generic_expressions.py
+++ b/safe/gis/generic_expressions.py
@@ -85,7 +85,8 @@ help_message = generate_expression_help(description, examples)
 @qgsfunction(
     args='auto', group='InaSAFE', usesGeometry=False, referencedColumns=[],
     help_text=help_message.to_html(), helpText=help_message.to_html())
-def inasafe_sub_analysis_summary_field_value(exposure_key, field, feature, parent):
+def inasafe_sub_analysis_summary_field_value(
+        exposure_key, field, feature, parent):
     """Retrieve a value from field in the specified exposure analysis layer.
 
     """
--- a/safe/gis/generic_expressions.py
+++ b/safe/gis/generic_expressions.py
@@ -73,10 +73,10 @@ def inasafe_analysis_summary_field_value(field, feature, parent):
     return feature[index]
 
 
-description = tr('Retrieve a value from a field in the sub impact analysis '
+description = tr('Retrieve a value from a field in the sub analysis summary '
                  'layer from a multi exposure analysis layer.')
 examples = {
-    'inasafe_multi_exposure_field_value('
+    'inasafe_sub_analysis_summary_field_value('
     '\'population\', \'total_not_exposed\')': 3
 }
 help_message = generate_expression_help(description, examples)
@@ -85,7 +85,7 @@ help_message = generate_expression_help(description, examples)
 @qgsfunction(
     args='auto', group='InaSAFE', usesGeometry=False, referencedColumns=[],
     help_text=help_message.to_html(), helpText=help_message.to_html())
-def inasafe_multi_exposure_field_value(exposure_key, field, feature, parent):
+def inasafe_sub_analysis_summary_field_value(exposure_key, field, feature, parent):
     """Retrieve a value from field in the specified exposure analysis layer.
 
     """
--- a/safe/gis/generic_expressions.py
+++ b/safe/gis/generic_expressions.py
@@ -11,7 +11,9 @@ from qgis.core import (
 )
 
 from safe.definitions.extra_keywords import all_extra_keywords_description
-from safe.definitions.provenance import provenance_layer_analysis_impacted_id
+from safe.definitions.provenance import (
+    provenance_layer_analysis_impacted_id,
+    provenance_multi_exposure_analysis_summary_layers_id)
 from safe.definitions.utilities import definition
 from safe.gis.tools import load_layer
 from safe.utilities.i18n import tr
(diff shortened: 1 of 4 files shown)


safe/report/expressions/map_report.py
```diff
@@ -208,7 +239,7 @@ def bearing_to_nearest_place(feature, parent):
 @qgsfunction(
     args='auto', group=label_group, usesGeometry=False, referencedColumns=[],
     help_text=help_message.to_html(), helpText=help_message.to_html())
-def name_of_the_nearest_place(feature, parent):
```
Gustry: This comment is for all expressions in this file, is-it not always `place` for the "exposure_key" ? Or can't we make it not compulsory? (ie keep the previous behavior make the new exposure_key field optional?)

myarjunar: Ivan said we might be able to use it for volcanic ash as well. CMIIW @ivanbusthomi

Gustry: ok.
But can't we use this expression outside of the multiexposure context? (keep the previous behavior?)

myarjunar: yes, we can 😃  I will make some changes first

Gustry: ok cool. let me know if it's possible

Gustry:
@myarjunar Maybe you need to explain/show the new modified expressions to @ivanbusthomi. He is already using these expressions: `name_of_the_nearest_place()` but we are breaking our public API in QGIS expressions (need to warn our users)

Gustry:
QGIS expressions are made public so we shouldn't update them as it breaks templates that users already made (not only DMI), see comment here: https://github.com/inasafe/inasafe/pull/4908

I now that it's new and maybe not a lot of users are using these expressions but we should keep this in mind when we design expressions. When they are public, they are there to stay ;-)

ivanbusthomi:
So I tried this PR on earthquake multiexposure analysis. I received the following error
```
Problem
AttributeError

Details
'NoneType' object has no attribute 'source'

Diagnostics
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\gui\tools\multi_exposure_dialog.py", line 555, in accept code, message = self._multi_exposure_if.run()
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\impact_function\multi_exposure_wrapper.py", line 978, in run impact_function.exposure_summary)
In file "C:/Users/busthi/.qgis2/python/plugins\inasafe\safe\gis\tools.py", line 143, in full_layer_uri return layer.source() + '|qgis_provider=' + layer.providerType()
```

myarjunar:
@Gustry I have restore the old behaviour so user wouldnt be confused by our change. it's only a matter of selecting the layer. anyway, could you help why am I keep getting this failing travis. (it's green on my desktop)

safe/gis/generic_expressions.py
```diff
@@ -71,6 +73,46 @@ def inasafe_analysis_summary_field_value(field, feature, parent):
...
+    'inasafe_multi_exposure_field_value('
+    '\'population\', \'total_not_exposed\')': 3
+}
+help_message = generate_expression_help(description, examples)
+
+
+@qgsfunction(
+    args='auto', group='InaSAFE', usesGeometry=False, referencedColumns=[],
+    help_text=help_message.to_html(), helpText=help_message.to_html())
+def inasafe_multi_exposure_field_value(exposure_key, field, feature, parent):
```
Gustry: It's not clear that the expression if fetching the field in the analysis summary layer. Only by reading the code I understand it's fetching in the analysis summary layer (and not in the aggregation layer for instance) We need to update the expression name IMHO?
`inasafe_sub_analysis_summary_field_value`?

I know it's difficult to find a good name :D

safe/impact_function/multi_exposure_wrapper.py
```diff
@@ -966,13 +972,39 @@ def run(self):
...
+                impact_function.analysis_impacted)
+            dict_of_analysis_summary_id[exposure_key] = (
+                impact_function.analysis_impacted.id())
+
+            # eq raster on pop raster might not having this
+            if impact_function.exposure_summary:
+                dict_of_exposure_summary_path[exposure_key] = full_layer_uri(
+                    impact_function.exposure_summary)
+                dict_of_exposure_summary_id[exposure_key] = (
+                    impact_function.exposure_summary.id())
```
Gustry: I don't know if you care, but also the `exposure_breakdown` might not exist.

safe/report/expressions/map_report.py
```diff
@@ -101,20 +110,40 @@ def distance_to_nearest_place(feature, parent):
     project_context_scope = QgsExpressionContextUtils.projectScope()
     registry = QgsMapLayerRegistry.instance()
 
+    active_layer = iface.activeLayer()
```
Gustry: So we don't use project variable?

We start having different behaviors like having an active layer in the legend and/or having project variables.

Can we select any layer in the multiexposure group, it will work? (whatever the outlayer selected in the legend, it will find the correct layer right?)

safe/report/expressions/map_report.py
```diff
@@ -208,7 +278,7 @@ def bearing_to_nearest_place(feature, parent):
 @qgsfunction(
     args='auto', group=label_group, usesGeometry=False, referencedColumns=[],
     help_text=help_message.to_html(), helpText=help_message.to_html())
-def name_of_the_nearest_place(feature, parent):
+def name_of_the_nearest_place(exposure_key, feature, parent):
```
Gustry: So we still have the exposure_key? Even if you added the logic with the selectedlayer?

safe/report/expressions/map_report.py
```diff
@@ -169,7 +219,7 @@ def direction_to_nearest_place(feature, parent):
 @qgsfunction(
     args='auto', group=label_group, usesGeometry=False, referencedColumns=[],
     help_text=help_message.to_html(), helpText=help_message.to_html())
-def bearing_to_nearest_place(feature, parent):
+def bearing_to_nearest_place(exposure_key, feature, parent):
```
Gustry: exposure key and the selected layer?

safe/impact_function/multi_exposure_wrapper.py
myarjunar: what do you mean by `exposure_breakdown` sorry ?

safe/gis/generic_expressions.py
myarjunar: aggree :+1:

safe/report/expressions/map_report.py
myarjunar: my fault

safe/impact_function/multi_exposure_wrapper.py
Gustry: exposure summary table sorry: https://github.com/inasafe/inasafe/blob/develop/safe/definitions/layer_purposes.py#L161

The exposure breakdown by exposure type. If you have a census vector exposure or pop raster exposure, you won't have this layer. I think it's fine in your case

ivanbusthomi:
Hi @myarjunar - I have tested it and it looks good to me

PR #4923, opened 2018-02-28
Headless report

myarjunar:
### What does it fix?

* Ticket: #
* Funded by: DFAT
* Description: 
  * fix unevaluated expressions in inasafe realtime
  * fix duplicated impact layer on the map report for multi-exposure

### Checklist:

### DIFF
--- a/safe/gis/generic_expressions.py
+++ b/safe/gis/generic_expressions.py
@@ -4,6 +4,7 @@
 
 import datetime
 
+from dateutil.parser import parse
 from qgis.core import (
     qgsfunction,
     QgsMapLayerRegistry,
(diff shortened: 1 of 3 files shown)


safe/gis/generic_expressions.py
```diff
@@ -287,14 +295,21 @@ def beautify_time(inasafe_time, feature, parent):
     """
     _ = feature, parent  # NOQA
     try:
-        datetime_object = datetime.datetime.strptime(
-            inasafe_time, '%Y-%m-%dT%H:%M:%S.%f')
-    except ValueError:
+        gmt_offset_str = inasafe_time[-6:-4]
```
ismailsunni: It seems duplicate code with the one in beautify_date. Can you make it as a function?

safe/gis/generic_expressions.py
```diff
@@ -252,14 +253,21 @@ def beautify_date(inasafe_time, feature, parent):
...
```
Gustry: This looks dangerous. Is there not a function in datetime module to get the offset?

ismailsunni: Agree with @Gustry , perhaps you can check this one https://stackoverflow.com/questions/12281975/convert-timestamps-with-offset-to-datetime-obj-using-strptime

myarjunar: I did a lot of googling for this and not sure there is such a function in datetime library.

safe/gis/generic_expressions.py
Gustry: +1

safe/gis/generic_expressions.py
Gustry: Ok, strange from a famous python module.
So make it safer as possible and future proof ;-)

Can you move the logic to safe.utilities folder in a single function (linked to the comment from Ismail) and add some quick tests?

myarjunar: thanks @ismailsunni @Gustry , ok so with `parse` method we could generally convert it into the datetime object. but my intention was for knowing how much the time difference is so we could convert it to localtime. we could figure it put later

PR #4943, opened 2018-03-14
update layer id whenever IF is being serialized

### LINKED ISSUE #4942: Problem with provenance layer id on IF serialization
### Problem
IF serialization did not update the layer id in the provenance so the information is not reliable. We need to update the layer id whenever the IF is being serialized.

### Expected Behaviour
layer id representing the actual layer available 

### Environment
* InaSAFE : 
* QGIS : 
* OS : 

* Data used :
  * Exposure : 
  * Hazard : 
  * Aggregation :
  * Extent : 
  
Checklist: 

* [ ] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ?

myarjunar:
### What does it fix?

* Ticket: fix #4942 
* Funded by: DFAT
* Description: 
  * update layer id whenever IF is being serialized

### Checklist:

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -356,12 +356,17 @@ class ImpactFunction(object):
             'hazard',
             'exposure',
             'aggregation',
-            'impact',
-            'exposure_summary',
-            'aggregate_hazard_impacted',
-            'aggregation_summary',
-            'analysis_impacted',
-            'exposure_summary_table',
+
+            # Output layers on new IF object will have a different provenance
+            # data with the one from original IF.
+
+            # 'impact',
+            # 'exposure_summary',
+            # 'aggregate_hazard_impacted',
+            # 'aggregation_summary',
+            # 'analysis_impacted',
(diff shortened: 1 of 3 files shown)


myarjunar:
@Gustry  because it's being used by a lot of expressions and we want to keep the old behavior. So the id information needs to be reliable.

Gustry:
But we should have use path or something else for these expressions. Something reliable with time, not having a temporary ID.

I think you need to backport this to the master branch now.

Or maybe it's time for me to be retired from InaSAFE ;D

myarjunar:
I came with this solution because you said you want to keep the old behavior first (layer id in the provenance), but then check the layer path if the id doesn't exist :sweat_smile:  (I will add the code for checking the layer path)

> Or maybe it's time for me to be retired from InaSAFE ;D

nooooo :sob: :sob:

Gustry:
I meant during our first implementation of using this ID ;-)

PR #4952, opened 2018-03-16
fix save default style in datastore

Gustry:
### What does it fix?

* Ticket: :point_up: [March 16, 2018 12:22 AM](https://gitter.im/inasafe/inasafe?at=5aab00383fe1be3704c9d7c9) by @lucernae 
* Funded by: DFAT
* Description: 
  * save default style in datastore

### Checklist:

### DIFF
--- a/safe/impact_function/impact_function.py
+++ b/safe/impact_function/impact_function.py
@@ -1884,7 +1884,8 @@ class ImpactFunction(object):
         for pre_processor in self._preprocessors:
             layer = pre_processor['process']['function'](self)
             purpose = pre_processor['output'].get('value')['key']
-            result, name = self.datastore.add_layer(layer, purpose)
+            save_style = pre_processor['output'].get('save_style', False)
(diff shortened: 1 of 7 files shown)


safe/impact_function/impact_function.py
```diff
@@ -1884,7 +1884,7 @@ def pre_process(self):
         for pre_processor in self._preprocessors:
             layer = pre_processor['process']['function'](self)
             purpose = pre_processor['output'].get('value')['key']
-            result, name = self.datastore.add_layer(layer, purpose)
+            result, name = self.datastore.add_layer(layer, purpose, True)
```
lucernae: I'm not sure if I understand it correctly.
So actually the style was copied in temp folder but it doesn't get copied into the datastore?

Gustry: yes indeed. And it wouldn't be a good idea to save the style by default because we would have a QML file even if there is no style defined (not different than the default style provided by QGIS).
The datastore for now was not taking care about a possible previous style.

PR #4954, opened 2018-03-18
read specified extent from template

myarjunar:
### What does it fix?

* Ticket: #
* Funded by: DFAT
* Description: 
  * in flood realtime, we had a template with map extent specified in it. we need to read that specified extent and using it instead of using the default extent selection.  

### Checklist:


safe/report/processors/default.py
```diff
@@ -527,10 +549,10 @@ def qgis_composer_renderer(impact_report, component):
             half_length = longest_width / 2
             margin = half_length / 5
             center = extent.center()
-            min_x = center.x() - half_length - margin
-            max_x = center.x() + half_length + margin
-            min_y = center.y() - half_length - margin
-            max_y = center.y() + half_length + margin
+            min_x = (min_x or center.x() - half_length) - margin
```
lucernae: I'm not sure if this is intended. But if you do it like this then margin will always be applied either min_x is coming from the template or default value.
User who specify the extent manually will not see the same thing as what he set (because margin is applied)

myarjunar: I thought the margin should always be applied for layouting purpose even though the user set the extent manually in the template. if it's not the case, yeah I can update it.

myarjunar: updated, agree that we need to apply whatever the user put in the template without changing anything

PR #4977, opened 2018-04-17
add locale as a parameter on getting the qgis app

myarjunar:
### What does it fix?

adding a locale parameter on method for getiing the qgis app. this is an attempt on making inasafe headless able to run analysis on different locale
* Ticket: 
* Funded by: DFAT
* Description: 
  * 

### Checklist:


safe/test/utilities.py
```diff
@@ -69,7 +73,13 @@ def get_qgis_app():
     """
     global QGIS_APP, PARENT, IFACE, CANVAS  # pylint: disable=W0603
 
-    if iface:
+    from PyQt4.QtCore import QSettings
+    settings = QSettings()
+
+    current_locale = settings.value('locale/userLocale', 'en_US')
```
ismailsunni: Consider using our `set_general_setting` method. Be careful to change this in the unit test. It may break another test since you change the language to Indonesia in `test_qgis_app_locale`.

I suggest adding `QSettings` as a parameter so that you can specify which QSetting object that you want to change.

safe/test/utilities.py
```diff
@@ -100,6 +110,25 @@ def get_qgis_app():
         # noinspection PyCallByClass,PyArgumentList
         QCoreApplication.setApplicationName('QGIS2InaSAFETesting')
 
+        # Save some settings
+        settings = QSettings()
+        settings.setValue('locale/overrideFlag', True)
```
ismailsunni: Please use `set_setting` from InaSAFE code.

safe/test/utilities.py
myarjunar: thanks, updated

safe/test/utilities.py
```diff
@@ -100,6 +117,29 @@ def get_qgis_app():
...
+        if qsetting:
+            settings = QSettings(qsetting)
+        else:
+            settings = QSettings()
+
+        settings.setValue('locale/overrideFlag', True)
+        settings.setValue('locale/userLocale', requested_locale)
+        # We disabled message bars for now for extent selector as
+        # we don't have a main window to show them in TS - version 3.2
+        settings.setValue('inasafe/show_extent_confirmations', False)
```
ismailsunni: Use our `set_setting` method

safe/test/utilities.py
```diff
@@ -100,6 +117,29 @@ def get_qgis_app():
         # noinspection PyCallByClass,PyArgumentList
         QCoreApplication.setApplicationName('QGIS2InaSAFETesting')
 
+        # Save some settings
+        if qsetting:
+            settings = QSettings(qsetting)
+        else:
+            settings = QSettings()
+
+        settings.setValue('locale/overrideFlag', True)
```
ismailsunni: Use our `set_general_setting` method

myarjunar: yes I changed it again for testing reason, I'm aware of it

safe/common/parameters/test/example.py
```diff
@@ -5,7 +5,7 @@
 
 from safe.test.utilities import get_qgis_app
 
-QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app()
+QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app(qsetting='InaSAFETest')
```
Gustry: Could you make this 'InaSAFETest' the default string in the function? instead of updating all usages.

safe/common/parameters/test/example_group_select.py
```diff
@@ -34,7 +32,8 @@
 
 def main():
     """Main function to run the example."""
-    app, iface = qgis_app()
+    from safe.test.utilities import get_qgis_app
+    QGIS_APP, CANVAS, IFACE, PARENT = get_qgis_app(qsetting='InaSAFETest')
```
Gustry: Why do you rollback this change? We are trying to update our test suite to use the internal mock iface provided by QGIS core instead of own one (not very well maintained in InaSAFE)
Especially with the move to QGIS 3 soon, this code will be updated again.

safe/gui/tools/test/test_multi_exposure.py
```diff
@@ -20,7 +19,9 @@
 class MultiExposureDialogTest(unittest.TestCase):
 
     def setUp(self):
-        _, self.iface = qgis_app()
+        from safe.test.utilities import get_qgis_app
+        QGIS_APP, CANVAS, self.iface, PARENT = get_qgis_app(
```
Gustry: Same question here. (and all similar updates)

safe/test/test_translations.py
```diff
@@ -136,6 +136,43 @@ def test_qgis_translations(self):
         message = 'expected %s but got %s' % (expected_message, real_message)
         self.assertEqual(expected_message, real_message, message)
 
+    @unittest.skipIf(
```
Gustry: In this file, can you also clean these 'Xtest'. We put an X to avoid travis to run them. We can't add the decorator for some reason but it seems your new test is covering these tests already, so maybe you can remove them? It would be nice. Thanks

safe/common/parameters/test/example_group_select.py
myarjunar: because @ismailsunni said that it would be nice if we are using qgis_app in for testing purpose then we set the qsetting to `InaSAFETest` context so it doesn't change the main context. So I assume that I need to update every qgis_app usage with the new one that I just updated. 

Are you suggesting me to leave the usage of the `qgis_app()` as it is?

safe/common/parameters/test/example.py
myarjunar: Hi ! yes of course !

safe/common/parameters/test/example_group_select.py
Gustry: `get_qgis_app` is the legacy way of running tests. We want to move to the builtin QGIS core class instead of InaSAFE class. I tried a prototype with the `qgis_app`. So IMHO, we are moving backward with this change

safe/test/test_translations.py
```diff
@@ -136,6 +76,50 @@ def test_qgis_translations(self):
...
+        from safe.definitions.constants import no_field
+        self.assertNotEqual(no_field, u'No Field')
+
+        expected_locale = 'id'
+        self.assertEqual(locale(INASAFE_TEST), expected_locale)
+
+        # check for bahasa translation
+        expected_message = (
+            'Tidak ada informasi gaya yang ditemukan pada lapisan %s')
+        real_message = QCoreApplication.translate(
```
ismailsunni: Why not test the `tr` function?

safe/common/parameters/test/example_group_select.py
myarjunar: ah yes I remember, we've tried `qgis_app()` before as we updated the code for headless, but it seems like the `qgis_app()` causing a segmentation fault as we developed the Inasafe headless porting to version 4 (CMIIW @ismailsunni @lucernae ). So I'm just being pragmatic to update the code that works for headless now.

safe/test/test_translations.py
myarjunar: no reason, I'm just applying the old unittest which is already been there before. I agree it's more usefull to test the `tr` function it self

safe/common/parameters/test/example_group_select.py
myarjunar: I will merge this PR first since we urgently need this for headless internationalization. If you still feel uncomfortable with this, I have created a ticket relating to this https://github.com/inasafe/inasafe/issues/4982 . (I'm agree with your point of using qgis built in testing utilities)

PR #4979, opened 2018-04-21
Add a document template for QGIS3 migration API break

elpaso:
### What does it fix?

Add a document template for QGIS3 migration API break


Gustry:
This PR is also targeting the master branch and this file is also included in #5000. I think this PR can be closed.

PR #4996, opened 2018-06-04
update logic so it doesnt read keywords twice

myarjunar:
### What does it fix?

* Ticket: #
* Funded by: Kartoza
* Description: 
  * `monkey_patch_keywords()` duplicate the logic that we've done on the previous code (reading the keywords)

### Checklist:

### DIFF
--- a/safe/gis/tools.py
+++ b/safe/gis/tools.py
@@ -13,7 +13,7 @@ from qgis.core import (
     QgsDataSourceURI,
 )
 
-from safe.common.exceptions import NoKeywordsFoundError, InvalidLayerError
+from safe.common.exceptions import InvalidLayerError
 from safe.definitions.constants import (
     VECTOR_DRIVERS,
     RASTER_DRIVERS,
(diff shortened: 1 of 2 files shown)


safe/gis/tools.py
```diff
@@ -249,10 +250,12 @@ def load_layer(full_layer_uri_string, name=None, provider=None):
...
             layer_purpose = keywords['layer_purpose']
-    except NoKeywordsFoundError:
-        pass
+    except (NoKeywordsFoundError, MetadataReadError):
+        keywords = {}
 
-    monkey_patch_keywords(layer)
+    if not keywords.get('inasafe_fields'):
+        keywords['inasafe_fields'] = {}
+    layer.keywords = keywords
```
Gustry: I can't remember exactly the reason. But because this monkey patching is a kind of hack in the QgsMapLayer class (not part of the official QgsMapLayer API), I think I wanted to centralize this hack in a single function. But do as you want

myarjunar: sure we can centralize it using this monkey patching method, and remove the upper code instead because it's duplicated. let me update it. thanks for the review!

PR #5007, opened 2018-06-26
Ensure that clean_layer updates layer with repaired geometries

nyalldawson:
The current clean_layer implementation checks and repairs the feature's geometries, but never actually writes any repaired geometries back to the layer. Accordingly the method currently:

1. Checks for validity, and attempts to fix invalid geometries
2. If the geometry repair fails, the feature IS correctly removed from the layer
BUT
3. If the geometry repair is successful, the repaired geometry is only set for the local copy of the feature, which is then discarded as it immediately goes out of scope.

With this change we correctly update the layer's stored geometry by calling  QgsVectorLayer.changeGeometry with the repaired geometry.

### DIFF
--- a/safe/gis/vector/clean_geometry.py
+++ b/safe/gis/vector/clean_geometry.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 """Try to make a layer valid."""
+from qgis.core import QgsFeatureRequest
 
 from safe.common.custom_logging import LOGGER
 from safe.definitions.processing_steps import clean_geometry_steps
@@ -31,11 +32,12 @@ def clean_layer(layer):
     count = 0
 
     # iterate through all features
-    for feature in layer.getFeatures():
+    request = QgsFeatureRequest().setSubsetOfAttributes([])
+    for feature in layer.getFeatures(request):
         geom = feature.geometry()
         geometry_cleaned = geometry_checker(geom)
-        if geometry_cleaned:
-            feature.setGeometry(geometry_cleaned)
+        if geometry_cleaned and not geom.equals(geometry_cleaned):
+            layer.changeGeometry(feature.id(), geometry_cleaned, True)
         else:
             count += 1
             layer.deleteFeature(feature.id())
@@ -50,7 +52,6 @@ def clean_layer(layer):
 
     # save changes
     layer.commitChanges()
-
     layer.keywords['title'] = output_layer_name
 
     check_layer(layer)


nyalldawson:
Note that this causes lots of test failures, presumably because the current test results are based off calculations using the invalid geometries.

nyalldawson:
> Why not using makeValid BTW now in this file? https://qgis.org/api/classQgsGeometry.html#a9e91518c5ea16bc24ff60e6ba8d7ae66
We would discard less geometries.

nyalldawson:
> Why not using makeValid BTW now in this file? https://qgis.org/api/classQgsGeometry.html#a9e91518c5ea16bc24ff60e6ba8d7ae66
We would discard less geometries.

That was already done in the original PR (along with changing all other `buffer(0)` calls to `makeValid`. It's part of what the `geometry_checker` function used here does.

nyalldawson:
@Gustry @ismailsunni I'm not sure how to proceed with this PR. As noted, it causes MANY failed tests. But it's definitely a bug in the current code, so the existing test results must be incorrect.

I don't feel comfortable blindly changing the expected values here though without full knowledge of exactly what each test is supposed to protect against.

Gustry:
Arf sorry. Mea culpa. It was a very big PR I didn't review too much

nyalldawson:
> Arf sorry. Mea culpa. It was a very big PR I didn't review too much

No problem! I'd suggest renaming `geometry_checker` to `repair_invalid_geometry` would be a good move too - it would be more descriptive and make immediate understanding of this function clearer.

Gustry:
I think you can. InaSAFE is used as a python API in some projects or in other QGIS plugins, but this function is kind private IMHO. It's mainly `impact_function.py` which is public. So feel free to do it.

Gustry:
Quick look in travis:
* Tests in `report.test.test_impact_report.TestImpactReport` and `report.test.test_impact_report_earthquake` : these tests are safe to fix blindly I think. We are not testing the GIS part in these tests. Only that we can generate reports (HTML tables) given some QGIS layers. It's a few tests failing here. @lucernae @myarjunar  Do you confirm?
* impact_function.test.test_multi_exposure_impact_function This one, is not checking the GIS part too I think so safe to fix. But the error message is showing us another bug I think (invalid layer).
* `impact_function.test.test_impact_function` these ones need to be checked ! Some error messages are warning me ... (@ismailsunni) 

My 2 cents late in the evening ...

ismailsunni:
Yes, I agree with @Gustry 's last comment above. I think we need to check / fix /investigate the `impact_function.test.test_impact_function` first. 

The report related test can be fixed blindly if it's only different in the number. Just bear in mind, if the changes are too much (e.g. previously it has some number, no it's gone), we need to check it carefully. Since probably it's failed to run aggregation or post processor.

If possible, we should make a smaller unit test for the `clean_layer` method, so that we can see the difference in the lower level.

Actually, @Gustry has more experience in this area :), but of course, I can still help.

timlinux:
@ismailsunni this is what I was referring to about backporting to 4.x too please - we need to get this fixes out to GeoSAFE / InaSAFE realtime via a 4.x update

ismailsunni:
@timlinux yes, I will merge it to `inasafe_4` also. Actually, I found some bugs from v4 from my testing with InaSAFE 5

timlinux:
@ismailsunni can you work to get the tests passing for this so we can merge it?

ismailsunni:
Hi @nyalldawson @Gustry @timlinux 
Based on this PR, it will actually remove the feature although the geometry is already valid. The `geometry_checker` return a not None value if the geometry is already valid (https://github.com/inasafe/inasafe/pull/5007/files#diff-13d4402cb1b69c4b5d5f01c53499b8e0R74). 

Since @nyalldawson added the `and not geom.equals(geometry_cleaned)`, the case where geometry is already valid will be removed in the `else` case (https://github.com/inasafe/inasafe/pull/5025/files#diff-13d4402cb1b69c4b5d5f01c53499b8e0R43)

So, I made another PR to fix it (https://github.com/inasafe/inasafe/pull/5025) to only change the geometry if the geometry is cleaned and if the cleaned geom is changed. If the geometry is cleaned, but the geometry is not changed, it will do nothing.

Let me know your thought :)

I will merge my PR tomorrow if all is good.

ismailsunni:
Updated in #5025

PR #5014, opened 2018-06-28
Replace custom clip alg with built-in processing algorithm

nyalldawson:
Because:
- it's native c++ code (faster)
- lots of unit tests, user tests (so should be more stable)
- burden of maintainence is on QGIS project


nyalldawson:
Thanks @ismailsunni ! Next up - reclassify and smart clip

PR #5025, opened 2018-08-07
Fix geom repair

ismailsunni:
### What does it fix?
Trying to update https://github.com/inasafe/inasafe/pull/5007

* Ticket: #
* Funded by: Kartoza
* Description: 
  * 

### Checklist:

### DIFF
--- a/safe/gis/vector/clean_geometry.py
+++ b/safe/gis/vector/clean_geometry.py
@@ -36,8 +36,15 @@ def clean_layer(layer):
     for feature in layer.getFeatures(request):
         geom = feature.geometry()
         geometry_cleaned = geometry_checker(geom)
-        if geometry_cleaned and not geom.equals(geometry_cleaned):
-            layer.changeGeometry(feature.id(), geometry_cleaned, True)
+        # Geometry now is clean (can be the same geom, or different geom)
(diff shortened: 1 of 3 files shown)


nyalldawson:
Looks good to me!

I'd suggest for API clarity that geometry_checker should return both a repaired geometry and a bool to represent whether the geometry was already valid. This would avoid the extra geometry "equals" check too, which is sometimes expensive to calculate.

safe/gis/vector/clean_geometry.py
```diff
@@ -31,12 +32,19 @@ def clean_layer(layer):
...
+    request = QgsFeatureRequest().setSubsetOfAttributes([])
+    for feature in layer.getFeatures(request):
         geom = feature.geometry()
-        geometry_cleaned = geometry_checker(geom)
-        if geometry_cleaned:
-            feature.setGeometry(geometry_cleaned)
+        was_valid, geometry_cleaned = geometry_checker(geom)
+
+        if was_valid:
+            # Do nothing it it was valid
```
Gustry: it it?

ismailsunni: argh, thanks

PR #5052, opened 2018-11-28
Use UTF-8 encoding for minimum needs definition/processor key

lucernae:
### Refactor Template String to use unicode

Minimum Needs are locale based file and can contain unicode/utf-8 encoding.
These values were interpolated by InaSAFE definitions' template string for key definition and postprocessors.
In order for unicode based minimum needs file to works, these template string needs to be in unicode.

### Checklist:

- [X] Commit messages are descriptive and explain the rationale for changes
- [X] Request someone to review or test your PR


lucernae:
I was thinking maybe we are allowed to put Mozambique's minimum needs profile to include it as test data for unicode min needs.

lucernae:
Gah, I forgot that you had to restart InaSAFE Desktop to apply Minimum Needs into definitions and post processors.

lucernae:
Thanks. I will merge this for Headless.

PR #5059, opened 2019-02-19
Fix raster classification issue

### LINKED ISSUE #5056: polygons in 'low' flood level are classified as 'medium'
### Problem

When running an analysis of raster flood hazard vs polygon exposure, buildings in the 'low' hazard category are instead reported as part of the 'medium' category.

See in that screenshot that some buildings indeed fall into the 'low' category, but are not classified as such (number of structures in 'Low' is zero):

screenshot from 2018-12-11 17-14-03

### Expected Behaviour

I would expect the buildings fully included in the relevant flooding category to be classified accordingly.

### Environment
* InaSAFE : 5.0.0 831d60abba
* QGIS : 3.4.2
* OS : Ubuntu 18.10

* Data used :
  * Exposure : OSM polygons
  * Hazard : Continuous raster water depth
  * Aggregation : None
  * Extent : Entire Area

lrntct:
### What does it fix?
* Ticket: fix #5056
* Funded by: NERC NE/R014361/1
* Description: 
  * The bounding box clipping of the raster changed its type to integer, messing up the classification
  * Clipping now keeps the raster data type

### Checklist:
- [x] Add to the changelog in metadata.txt if it's a new feature

### DIFF
--- a/safe/gis/raster/clip_bounding_box.py
+++ b/safe/gis/raster/clip_bounding_box.py
@@ -83,7 +83,7 @@ def clip_by_extent(layer, extent):
         parameters['INPUT'] = layer.source()
         parameters['NO_DATA'] = ''
         parameters['PROJWIN'] = ','.join(bbox)
-        parameters['DATA_TYPE'] = 5
+        parameters['DATA_TYPE'] = 0
         parameters['COMPRESS'] = 4
         parameters['JPEGCOMPRESSION'] = 75
         parameters['ZLEVEL'] = 6


Gustry:
Thanks for your PR.

This change breaks some unittest. Can you check if the expected result is correct?

I see these tests are in the reporting module, so I think it's fine to update. (it's not the main goal of these tests to check GIS calculation). But need to check manually.

lrntct:
That is strange, `run-docker-test.sh` is not reporting any of those on my machine, only `test_flake8` (which I do not think is relevant here).
The two other failing tests (`test_analysis_detail` and `test_aggregation_area_result`) are built with vector hazards, so I do not believe they should be impacted by the change in the raster clipping / classification, but I am not familiar enough with inasafe to completely rule out that.
Any clue?

timlinux:
Hi @lrntct

Thank you for your PR.

It looks like the PR changes some test results (besides the flake-8 issue) - see https://travis-ci.org/inasafe/inasafe/jobs/495571666

lrntct:
I understand that it is my responsibility to make the code passes the tests. But some insight on what the tests actually do would help.

I see the values that make the tests fail, but I am not sure what they actually mean.
If the total number of affected building changes, and the previous number was estimated with the buggy code, then I would suppose the new number is right.

Knowing what the failures concretely mean would help to fix either the code or the test.

I could change the tests to make them pass if a senior dev tell me it is OK to do so.

Gustry:
As I said, failing tests were coming from the reporting module (testing the logic about building reports). It was not the tests about GIS calculation. So I think, IRRC, it's fine to update (after a quick check as @timlinux said).

lrntct:
The tests fail in the same random manner even without the change. I am quite confused.

lucernae:
Hi everyone. We fixed the numbers errors in report here:
https://github.com/inasafe/inasafe/pull/5071

We might be able to continue with the PR.
I will try to trigger Travis to redo the unittest, maybe we will have a better idea.

lucernae:
Thanks for your PR @lrntct .

We can close this PR, I included your commit in PR #5076 .
It doesn't break existing unit test, except the minimum needs count as expected.
Minimum needs count report an increased numbers because more class count exists in 'affected' class, also expected.

Issue #5063, opened 2019-09-23
inasafe 5 won't run in qgis3.x


husnimubarok:
### Problem
inasafe run resulting error "TypeError"
```
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1449, in run self._run()
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\utilities\profiling.py", line 126, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1626, in _run self.pre_process()
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\utilities\profiling.py", line 126, in with_profiling ret = fn(*args, **kwargs)
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\impact_function\impact_function.py", line 1889, in pre_process layer = pre_processor['process']['function'](self)
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\processors\pre_processors.py", line 115, in earthquake_contour_preprocessor contour_path = create_smooth_contour(impact_function.hazard)
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\gis\raster\contour.py", line 272, in create_smooth_contour active_band=active_band
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\gis\raster\contour.py", line 426, in shakemap_contour set_contour_properties(output_file_path)
In file "C:/Users/husnix/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\inasafe\safe\gis\raster\contour.py", line 458, in set_contour_properties line = feature.geometry().asPolyline()
```
### Expected Behaviour
With the same dataset, inasafe 4.3.4 and qgis 2.18 inasafe analysis working without problem
### Environment
* InaSAFE : 5
* QGIS : 3.4 and 3.8
* OS : Win 10 and Ubuntu 18.04

[...]

PR #5071, opened 2019-10-31
Bugfix update

lucernae:
Added various commits related with bugfix and unittest improvements.

- Critical update for #5063 
- Enable previously disabled unittest


safe/report/test/test_impact_report.py
```diff
@@ -1157,7 +1156,7 @@ def test_minimum_needs_outputs(self):
                     'needs': [
                         {
                             'header': 'Toilets',
-                            'value': '0'
+                            'value': '10'
```
ismailsunni: do you know why it gives different values in 3.4 and 2.x?

lucernae: Different rounding mechanism from inside QGIS (or maybe Qt?).
To illustrate,
Our displaced count field is an integer. for some dataset that we test, in a certain aggregation boundary, the total count of affected is close to one, like 1.01.
With displacement ratio of 0.9, displacement count in double is 0.909.
But our field have a type of Qt Int.
In QGIS 2.x it gets rounded to 0.
Minimum needs are based on displaced count, meaning toilet min needs is 0.

In QGIS 3.x it gets rounded to the nearest integer, which is 1.
Minimum needs are based on displaced count, meaning toilet min needs is now 1.
In the report, as convention we round values below 10 to 10.

lucernae: But this is only apparent because the number is small though... For large counts, I don't think it will matter.

ismailsunni: Ah, I see. Perhaps the reason is the python 2 to 3 itself. See https://docs.python.org/3/library/functions.html#round

Anyway, good eye

safe/messaging/item/text.py
```diff
@@ -164,7 +164,7 @@ def to_html(self):
         attributes = self.html_attributes()
         # Deal with long file names that prevent wrapping
         wrappable_text = self.to_text().replace(os.sep, '<wbr>' + os.sep)
-        if icon is not '' and attributes is not '':
+        if icon and attributesRe:
```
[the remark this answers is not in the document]
lucernae: Yes typo :D

safe/report/test/test_impact_report.py
lucernae: Yeah, it could be. I was horrified when seeing that this unittest fails because the value changed!
It should not happen because the process were deterministic, and people already uses InaSAFE 5.0 in QGIS. It has to be fixed ASAP.

lucernae:
Send moar patches to see what else is breaking.

lucernae:
@Gustry @ismailsunni I need your opinion on this.
The unittests stalled in QGIS Release 3.4.x, for test_contour package. Probably because I used geopackage to store the contour.
Does QGIS 3.4 not support geopackage?
Or maybe something to do with OGR GDAL in QGIS 3.4.x caused it to stall?
The QGIS latest build seems to be green.

Gustry:
> Does QGIS 3.4 not support geopackage?

No, AFAIK, nothing that you change the behavior of this test.

>  Or maybe something to do with OGR GDAL in QGIS 3.4.x caused it to stall?

According to travis, it's the same version.

SOrry, I have no idea. Except that you can run the test in QGIS 3.4 manually, by using the blue triangle, specifiying the package. Or you can also run the test in docker on your machine. `make docker_test` should work.

lucernae:
> SOrry, I have no idea. Except that you can run the test in QGIS 3.4 manually, by using the blue triangle, specifiying the package. Or you can also run the test in docker on your machine. make docker_test should work.

Blue triangle works. So I don't know what to do besides skipping the unittest.

We will gradually fix tests in QGIS 3.4 by skipping all affected tests.
At least QGIS latest one doesn't break.

PR #5073, opened 2019-11-11
Skip stalling unit tests in QGIS 3.4

lucernae:
Some of the unit test is stalling in QGIS 3.4
We don't know exactly why, yet.
The stalling code is inside qgis binary.
Running the unit test from inside QGIS Desktop works.
Running the unit test from QGIS 3.10 docker container works.

### DIFF
--- a/safe/impact_function/test/test_impact_function.py
+++ b/safe/impact_function/test/test_impact_function.py
@@ -2,7 +2,6 @@
 
 """Test for Impact Function."""
 
-
 import getpass
 import json
 import logging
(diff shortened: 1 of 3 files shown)


safe/impact_function/test/test_impact_function.py
```diff
@@ -415,6 +414,10 @@ def test_minimum_extent(self):
             message
         )
 
+    @unittest.skipIf(
+        os.environ.get('QGIS_VERSION_TAG') == 'release-3_4',
```
Gustry: I saw your trick to add the docker image as environment variable. You could use the QGIS API for that ;-) Like https://github.com/inasafe/inasafe/blob/develop/safe/utilities/gis.py#L366

lucernae: Thanks for the hint!
That is a better way.
I asked @ann26 to continue with the PR, but she submit it to my repo instead.
I will merge it after this.
https://github.com/lucernae/inasafe/pull/4

ann26: done, I've merged in here @Gustry @lucernae

PR #5076, opened 2019-11-21
Follow up on PR #5059

lucernae:
We now fixed the minimum needs calculation and bug on the unit test.
Probably this PR #5059 will succeed.


lucernae:
Will fix #5059 and fix #5056

lucernae:
Minimum needs in the report has increased because the raster classification now changed and report the correct class count.

Issue #5092, opened 2021-06-23
Canonical aggregation area sorting failed because of unknown hazard/aggregation class


lucernae:
### Problem

When we do aggregated analysis, we create an aggregated hazard summary layer.
We sort the layer according to the tuple (hazard class, aggregation id) in order to get
consistent table view in the result table.

In some cases (like uneven geometry in aggregation layer), we got a feature that 
may contain hazard class, but null aggregation id, in which InaSAFE failed to sort it.

### Expected Behaviour

In this cases, the sorter must recognize that null aggregation id means you just 
need to sort by hazard class (ignoring the aggregation tuple element).

### Environment
* InaSAFE : 5.0.1
* QGIS : Any
* OS : Any

* Data used :
  * Exposure : any exposures
  * Hazard : any raster hazard
  * Aggregation : aggregation layer that is strictly smaller than the hazard feature itself (so that we got a hazard that don't belong to any aggregation area)
  * Extent : the hazard layer (which is bigger than the aggregation layer)
  
Checklist: 

* [x] Did you try to run it in debug mode ?
* [ ] Is the data available ?
  * If yes, where (BTSync, data.inasafe.org, zip) ? 

### Proposed Solution

Fix the sorting algorithm.

Issue #5089, opened 2021-04-11
Couldn't load InaSAFE 5.0.1 in Qgis 3.16.5


Sheepnok:
This problem was found after I updated the QGIS version from 3.16.0 to 3.16.5. This window was displayed when I started the program.

**Problem**

Couldn't load plugin 'inasafe' due to an error when calling its classFactory() method 

ModuleNotFoundError: No module named 'ogr' 
Traceback (most recent call last):
  File "/Applications/QGIS-LTR.app/Contents/MacOS/../Resources/python/qgis/utils.py", line 334, in _startPlugin
    plugins[packageName] = package.classFactory(iface)
  File "/Users/nokyoong/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/inasafe/__init__.py", line 89, in classFactory
    from .safe.plugin import Plugin
  File "/Applications/QGIS-LTR.app/Contents/MacOS/../Resources/python/qgis/utils.py", line 792, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/Users/nokyoong/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/inasafe/safe/plugin.py", line 33, in 
    from safe.utilities.expressions import qgis_expressions
  File "/Applications/QGIS-LTR.app/Contents/MacOS/../Resources/python/qgis/utils.py", line 792, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/Users/nokyoong/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/inasafe/safe/utilities/expressions.py", line 8, in 
    from safe.report.expressions import infographic, map_report, html_report
  File "/Applications/QGIS-LTR.app/Contents/MacOS/../Resources/python/qgis/utils.py", line 792, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "/Users/nokyoong/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/inasafe/safe/report/expressions/html_report.py", line 24, in 
    from safe.definitions.reports.components import (
  File "/Applications/QGIS-LTR.app/Contents/MacOS/../Resources/python/qgis/utils.py", line 792, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
[...]

PR #5094, opened 2021-06-23
Aggregate sort

lucernae:
Fix https://github.com/inasafe/inasafe/issues/5092 by using QGIS own sorting API.
Fix https://github.com/inasafe/inasafe/issues/5089 by using standard osgeo import.


timlinux:
can we merge this @lucernae ?

PR #5105, opened 2021-10-13
fix slot errors

timlinux:


lucernae:
@timlinux was this supposed to be the proper fix?

